PR #3, opened 2020-11-16
move log setup out of init

gitlawr:
Problem:
Set up console log in init may affect k3os' logging behavior

Solution:
Move it out and do init specifically.


guangbochen (approved):
LGTM

PR #4, opened 2020-11-16
Update text

gitlawr:
1. Update install log. It is printed in wrong order
image

2. Remove node role color rendering, which is hard to read.
image


guangbochen (approved):
LGTM

PR #6, opened 2020-11-17
Use generated harvester hostname

gitlawr:
change generated hostname from k3os to harvester

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/nsf/termbox-go v0.0.0-20200418040025-38ba6e5628f1 // indirect
 	github.com/pkg/errors v0.9.1
 	github.com/rancher/k3os v0.19.2-dev.4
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/nsf/termbox-go v0.0.0-20200418040025-38ba6e5628f1 // indirect
 	github.com/pkg/errors v0.9.1
 	github.com/rancher/k3os v0.19.2-dev.4
+	github.com/sirupsen/logrus v1.6.0
```
guangbochen, line 12: why is the `logrus` needed here?

gitlawr: This is missing previously, k3os has logurs vendor so it passed the build without this line. It is added after running go mod tidy, helping to build console package independently

guangbochen (approved):

PR #9, opened 2020-11-27
Embed k3os

gitlawr:
Embed k3os to the repo instead of tracking the patch files. Update on k3os side is easier in this way


guangbochen (approved):
LGTM

PR #12, opened 2020-12-03
Import sshkeys in runtime phase

gitlawr:
https://github.com/rancher/harvester/issues/284

Problem:
Cert validation fails when ssh key URL is added because the server time on the init phase is out of date.

Solution:
Do ssh key import in the runtime phase.


guangbochen:
@gitlawr can u please help to rebase it, thanks

gitlawr:
Updated

guangbochen (approved):
LGTM

PR #13, opened 2020-12-03
Add install option for selecting management network NIC

gitlawr:
https://github.com/rancher/harvester/issues/203

Add the option in installation GUI console.
Update manifests and chart answers to adapt to the current chart.
Update to use embedded flannel.


guangbochen (approved):
LGTM

PR #14, opened 2020-12-03
update status in console dashboard

gitlawr:
https://github.com/rancher/harvester/issues/254

Show unknown status when the node is not present in the cluster


guangbochen (approved):
LGTM

PR #19, opened 2020-12-11
Fix inconsistent versions

gitlawr:
https://github.com/rancher/harvester/issues/315

1. Pass harvester version to the k3os build flags.
2. Prepare offline images based on dependencies versions on the harvester side.


guangbochen (approved):
LGTM

PR #23, opened 2020-12-15
Revert "K3OS: change sh symlink"

gitlawr:
This reverts commit 114373c8604504d216bca4035f8714ad38ae8e1d.

The issue is fixed on the Longhorn side so this workaround is not needed with longhorn-v1.1.0-rc2+


guangbochen (approved):
LGTM

PR #25, opened 2020-12-15
Fix Cloud init

gitlawr:
https://github.com/rancher/harvester/issues/350

Get remote cloud-init and merge with the install configurations.


guangbochen (approved):
LGTM

Issue #8, opened 2020-11-23
can't join the k3s cluster after promote an agent node to server node


futuretea:
I'm trying to promote an agent node by these commands:
```bash
sudo sed -i "s/command_args=\"agent/command_args=\"server/g" /etc/init.d/k3s-service
sudo /etc/init.d/k3s-service restart
```
but after do that, this node's status can't become Ready, check /var/log/k3s-service.log, errors reported as follows:
```bash
time="2020-11-22T03:10:25.695164274Z" level=info msg="Starting k3s v1.18.9+k3s1 (630bebf9)"
time="2020-11-22T03:10:25.793658747Z" level=fatal msg="starting kubernetes: preparing server: https://10.5.3.134:6443/v1-k3s/server-bootstrap: 403 Forbidden"
time="2020-11-22T03:10:31.689770648Z" level=info msg="Starting k3s v1.18.9+k3s1 (630bebf9)"
time="2020-11-22T03:10:31.778964598Z" level=fatal msg="starting kubernetes: preparing server: https://10.5.3.134:6443/v1-k3s/server-bootstrap: 403 Forbidden"
time="2020-11-22T03:10:37.699735384Z" level=info msg="Starting k3s v1.18.9+k3s1 (630bebf9)"
time="2020-11-22T03:10:37.786083390Z" level=fatal msg="starting kubernetes: preparing server: https://10.5.3.134:6443/v1-k3s/server-bootstrap: 403 Forbidden"
time="2020-11-22T03:10:43.705767537Z" level=info msg="Starting k3s v1.18.9+k3s1 (630bebf9)"
time="2020-11-22T03:10:43.799993533Z" level=fatal msg="starting kubernetes: preparing server: https://10.5.3.134:6443/v1-k3s/server-bootstrap: 403 Forbidden"
time="2020-11-22T03:10:49.715082958Z" level=info msg="Starting k3s v1.18.9+k3s1 (630bebf9)"

```
add "-- cluster-init" to the start option for the first server in /etc/init.d/k3s-service
```bash
k3s.io/node-args: '["server","--cluster-init","--disable","local-storage","--flannel-backend","none","--node-label","k3os.io/mode","local","--node-label","k3os.io/version","v0.11.1-2-g0c73235-dirty"]'
```
restart k3s both server and agent again, the agent node become server and ready
```bash
k3os-6055 [~]$ k3s kubectl get node
NAME         STATUS   ROLES    AGE   VERSION
[...]

PR #29, opened 2021-01-19
Support HA

futuretea:
**related issues**
https://github.com/rancher/harvester/issues/259
#8 

**related PR**
https://github.com/rancher/harvester/pull/270

1. dump k3os version to 1.19.5.rc.1( k3s version v1.19.5+k3s2)
2.  add k3s-restarter
3.  add --cluster-init to first server's k3s_args
4. add node label "harvester.cattle.io/version"
5. sync harvester management url


pkg/console/util.go
```diff
@@ -97,6 +98,10 @@ func customizeConfig() {
 		cfg.Config.Runcmd = append(cfg.Config.Runcmd, fmt.Sprintf(`keys=$(curl -sfL --connect-timeout 30 %q) && echo "$keys">>%s`, cfg.Config.SSHKeyURL, authorizedFile))
 	}
 
+	cfg.Config.K3OS.Labels = map[string]string{
+		"harvester.cattle.io/version": version.Version,
```
gitlawr, line 102: Why do we need the version label on nodes? We will need to consider the upgrade for this

pkg/console/util.go
```diff
@@ -120,12 +125,12 @@ func customizeConfig() {
 			Content:            getHarvesterManifestContent(harvesterChartValues),
 		},
 	}
+	cfg.Config.K3OS.Labels["svccontroller.k3s.cattle.io/enablelb"] = "true"
```
gitlawr, line 128: The klipper LB only provides access to one of the node IP so we might need to change it for HA. In that context, this label is not needed then.

scripts/image-list.txt
```diff
@@ -11,6 +11,7 @@ docker.io/kubevirt/virt-controller:v0.35.0
 docker.io/kubevirt/virt-handler:v0.35.0
 docker.io/kubevirt/virt-launcher:v0.35.0
 docker.io/kubevirt/virt-operator:v0.35.0
+docker.io/library/busybox:1.32.0
```
gitlawr, line 14: Let's remove the image-list.txt. It is expected to be generated during the build. Probably committed by mistake.

pkg/console/dashboard_panels.go
```diff
@@ -188,22 +181,58 @@ func initState() error {
...
+	cmd.Env = os.Environ()
+	output, err := cmd.CombinedOutput()
 	if err != nil {
-		return err
+		return "Unavailable"
 	}
-	current.harvesterURL = fmt.Sprintf("https://%s:8443", ip.String())
-	return nil
+
+	harvesterAddress := string(output)
```
gitlawr, line 226: I am concerned that if the command has unexpected output we might show something odd in the console. Is it possible?

gitlawr (approved):
LGTM

PR #30, opened 2021-01-22
Use K3OS CloudConfig to install SSH public keys

bk201:
Use K3OS CloudConfig to install SSH public keys

- Fetch and check user-provided SSH config and use K3OS CloudConfig to inject public keys.
- Remove `InstallConfig.SSHKeyURL`.

Co-authored-by: JenTing Hsiao <jenting.hsiao@suse.com>

**Related** to https://github.com/rancher/harvester/issues/217
To make PXE Boot easier, I would like to rely on the existing K3OS CloudConfig more, this PR removes the `SSHKeyURL` struct field.

## Screenshots

After the user press enter, the installer will check if the remote file contains correct public keys.

### 404
Screen Shot 2021-01-21 at 4 29 11 PM

### bad format
Screen Shot 2021-01-21 at 4 29 56 PM


gitlawr:
The HTTP request is done synchronously and the console UI will be irresponsive during this time(imagine a slow network and 15s timeout). So we might need to make an improvement for user experience. e.g., add a loading icon to show the ongoing work. What do you think?

bk201:
> The HTTP request is done synchronously and the console UI will be irresponsive during this time(imagine a slow network and 15s timeout). So we might need to make an improvement for user experience. e.g., add a loading icon to show the ongoing work. What do you think?

Sure, let me see what I can do to enhance it.

bk201:
I add a spinner panel if the operation takes too long. e.g.,
fetch

gitlawr:
The spinner looks nice! I think we don't need to hide the input panel during the process

bk201:
> The spinner looks nice! I think we don't need to hide the input panel during the process

fixed!
fetch2

pkg/console/install_panels.go
```diff
@@ -610,3 +633,61 @@ func addInstallPanel(c *Console) error {
...
+		c.Gui.Cursor = false
+		ticker := time.NewTicker(100 * time.Millisecond)
+		go func(g *gocui.Gui) {
+			printToPanel(g, asyncTaskPanel, fmt.Sprintf("%s|", prefix), true, infoColor)
+			for {
+				for _, s := range `\-/|` {
+					<-ticker.C
+					printToPanel(g, asyncTaskPanel, fmt.Sprintf("\r%s%c", prefix, s), false, infoColor)
+				}
+			}
```
gitlawr, line 671: The loop never breaks

bk201: Thanks. Fixed.
I also fixed some issues:
- The user still can type during the spinning
- Error message is not displayed if the async task is finished too fast.

I moved the logic back to the "Enter" key event handler, I feel it's more intuitive there.

gitlawr (approved):
LGTM

PR #32, opened 2021-01-27
update yq syntax from v3 to v4

gitlawr:
Fix CI failure: https://drone-publish.rancher.io/rancher/harvester-installer/30/1/2
The yq version installed by https://github.com/rancher/harvester-installer/blob/master/Dockerfile.dapper#L7 has changed


guangbochen (approved):
LGTM

PR #33, opened 2021-01-28
Support automatic installation

bk201:
Partially fixed: https://github.com/rancher/harvester/issues/217

## How it works

Support automatic installation

- Add Harvester config
  - The installer now accepts Harvester config and converts the config to any internal form if needed. (e.g., k3os CloudConfig).
- Add kernel parameters to support automatic installation
  - `harvester.install.automatic=true`. If this parameter is specified, then the installer is in automatic mode. No questions will be asked.
  - `harvester.install.config_url=http://to.my/config`. The installer will fetch the config at this URL and perform the automatic installation.

### Create mode

*iPXE script*
```
kernel harvester/vmlinuz k3os.mode=install console=ttyS0 console=tty1 harvester.install.automatic=true harvester.install.config_url=http://10.100.0.10/config-create.yaml
initrd harvester/initrd
boot
```

Where:
 - `k3os.install.config_url=http://10.100.0.10/config-create.yaml`: The Harvester config to be used.
   - e.g.,
```yaml
token: token
os:
  hostname: node1
  ssh_authorized_keys:
  - ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQDbeUa9A7Kee+hcCleIXYxuaPksn2m4PZTd4T7wPcse8KbsQfttGRax6vxQXoPO6ehddqOb2nV7tkW2mEhR50OE7W7ngDHbzK2OneAyONYF44bmMsapNAGvnsBKe9rNrev1iVBwOjtmyVLhnLrJIX+2+3T3yauxdu+pmBsnD5OIKUrBrN1sdwW0rA2rHDiSnzXHNQM3m02aY6mlagdQ/Ovh96h05QFCHYxBc6oE/mIeFRaNifa4GU/oELn3a6HfbETeBQz+XOEN+IrLpnZO9riGyzsZroB/Y3Ju+cJxH06U0B7xwJCRmWZjuvfFQUP7RIJD1gRGZzmf3h8+F+oidkO2i5rbT57NaYSqkdVvR6RidVLWEzURZIGbtHjSPCi4kqD05ua8r/7CC0PvxQb1O5ILEdyJr2ZmzhF6VjjgmyrmSmt/yRq8MQtGQxyKXZhJqlPYho4d5SrHi5iGT2PvgDQaWch0I3ndEicaaPDZJHWBxVsCVAe44Wtj9g3LzXkyu3k= root@admin
  password: rancher
install:
  mode: create
  mgmt_interface: eth0
  device: /dev/sda
  iso_url: http://10.100.0.10/harvester-amd64.iso
```
Note the `install.mode` is `create`. 

### Join mode

*iPXE script*
```
kernel harvester/vmlinuz k3os.mode=install console=ttyS0 console=tty1 harvester.install.automatic=true harvester.install.config_url=http://10.100.0.10/config-join.yaml
initrd harvester/initrd
boot
```

 - `k3os.install.config_url=http://10.100.0.10/config-join.yaml`: The Harvester config to be used.
   - e.g.,
```yaml
server_url: https://10.100.0.130:6443
token: token
os:
  hostname: node1
  ssh_authorized_keys:
  - ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQDbeUa9A7Kee+hcCleIXYxuaPksn2m4PZTd4T7wPcse8KbsQfttGRax6vxQXoPO6ehddqOb2nV7tkW2mEhR50OE7W7ngDHbzK2OneAyONYF44bmMsapNAGvnsBKe9rNrev1iVBwOjtmyVLhnLrJIX+2+3T3yauxdu+pmBsnD5OIKUrBrN1sdwW0rA2rHDiSnzXHNQM3m02aY6mlagdQ/Ovh96h05QFCHYxBc6oE/mIeFRaNifa4GU/oELn3a6HfbETeBQz+XOEN+IrLpnZO9riGyzsZroB/Y3Ju+cJxH06U0B7xwJCRmWZjuvfFQUP7RIJD1gRGZzmf3h8+F+oidkO2i5rbT57NaYSqkdVvR6RidVLWEzURZIGbtHjSPCi4kqD05ua8r/7CC0PvxQb1O5ILEdyJr2ZmzhF6VjjgmyrmSmt/yRq8MQtGQxyKXZhJqlPYho4d5SrHi5iGT2PvgDQaWch0I3ndEicaaPDZJHWBxVsCVAe44Wtj9g3LzXkyu3k= root@admin
  password: rancher
install:
  mode: join 
  mgmt_interface: eth0
  device: /dev/sda
  iso_url: http://10.100.0.10/harvester-amd64.iso
```
Note the `install.mode` is `join` and `server_url` needs to be provided.

## Discussion

**Update: we discussed this in sync call and decided to go this way**

Another idea is to completely not using k3os parameters and configs. Instead, we can use harvester-only parameters and configs, then convert them to k3os configs internally. For example, we can have a `harvester.install.iso_url` parameter rather than using `k3os.install.iso_url`.

 The advantages of this approach are:
  - It's clear: the user doesn't need to know k3os at all.
  - If the underlying OS is switched to another distro in the future, the user doesn't need to change the boot parameters.

### DIFF
--- a/k3os/install.sh
+++ b/k3os/install.sh
@@ -13,6 +13,7 @@ get_url()
 {
     FROM=$1
     TO=$2
+    echo "Downloading ${FROM} to ${TO}..."
     case $FROM in
         ftp*|http*|tftp*)
             n=0
--- a/pkg/console/validator.go
+++ b/pkg/console/validator.go
@@ -1,0 +1,106 @@
+package console
+
+import (
+	"fmt"
+	"net"
+
+	"github.com/pkg/errors"
+	"github.com/rancher/harvester-installer/pkg/config"
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	ErrMsgModeCreateContainsServerURL   = fmt.Sprintf("ServerURL need to be empty in %s mode", modeCreate)
+	ErrMsgModeJoinServerURLNotSpecified = fmt.Sprintf("ServerURL can't empty in %s mode", modeJoin)
+	ErrMsgModeUnknown                   = "Unknown mode"
+	ErrMsgTokenNotSpecified             = "token not specified"
+
+	ErrMsgMgmtInterfaceNotSpecified = "no management interface specified"
+	ErrMsgMgmtInterfaceNotFoud      = "interface not found"
+	ErrMsgMgmtInterfaceIsLoop       = "interface is a loopbck interface"
+	ErrMsgDeviceNotSpecified        = "no device specified"
+	ErrMsgDeviceNotFound            = "device not found"
+	ErrMsgNoCredentials             = "no SSH authorized keys or passwords are set"
+)
+
+type ValidatorInterface interface {
+	checkMgmtInterface(name string) error
+	checkDevice(device string) error
+}
+
+type Validator struct {
+}
+
+func prettyError(errMsg string, value string) error {
+	return errors.Errorf("%s: %s", errMsg, value)
+}
+
+func (v Validator) checkMgmtInterface(name string) error {
+	if name == "" {
+		return errors.New(ErrMsgMgmtInterfaceNotSpecified)
+	}
+	ifaces, err := net.Interfaces()
+	if err != nil {
+		return err
+	}
+	for _, i := range ifaces {
+		if i.Name == name {
+			if i.Flags&net.FlagLoopback != 0 {
+				return prettyError(ErrMsgMgmtInterfaceIsLoop, name)
+			}
+			return nil
+		}
+	}
+	return prettyError(ErrMsgMgmtInterfaceNotFoud, name)
+}
+
+func (Validator) checkDevice(device string) error {
+	if device == "" {
+		return errors.New(ErrMsgDeviceNotSpecified)
+	}
+	options, err := getDiskOptions()
+	if err != nil {
+		return err
+	}
+	for _, option := range options {
+		if device == option.Value {
+			return nil
+		}
+	}
+	return prettyError(ErrMsgDeviceNotFound, device)
+}
+
+func validateConfig(v ValidatorInterface, cfg *config.HarvesterConfig) error {
+	logrus.Debug("Validating config: ", cfg)
+
+	// modes
+	switch mode := cfg.Install.Mode; mode {
+	case modeCreate:
+		if cfg.ServerURL != "" {
+			return errors.New(ErrMsgModeCreateContainsServerURL)
+		}
+	case modeJoin:
+		if cfg.ServerURL == "" {
+			return errors.New(ErrMsgModeJoinServerURLNotSpecified)
+		}
+	default:
+		return prettyError(ErrMsgModeUnknown, mode)
+	}
+
+	if cfg.Token == "" {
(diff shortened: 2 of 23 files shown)


gitlawr:
Using Harvester specific parameters to avoid k3os domain knowledge makes sense to me.
We also need an update in the Harvester docs: https://github.com/rancher/harvester/tree/master/docs

bk201:
> Using Harvester specific parameters to avoid k3os domain knowledge makes sense to me.
> We also need an update in the Harvester docs: https://github.com/rancher/harvester/tree/master/docs

This means we need to completely hide k3os config and make this dialog accept harvester config. Does this cause any side effects? Not sure if k3os configs are used in components outside the installer.
Screen Shot 2021-01-29 at 5 24 24 PM

bk201:
The Harvester config proposal: https://gist.github.com/bk201/49927e0f02d5cc196e3a0e9ca3c13a82
Some k3s/k3os fields are removed and some hard-coded fields need to be refactoreded later.

pkg/console/validator.go
```diff
@@ -0,0 +1,106 @@
...
+
+	ErrMsgMgmtInterfaceNotSpecified = "no management interface specified"
+	ErrMsgMgmtInterfaceNotFoud      = "interface not found"
+	ErrMsgMgmtInterfaceIsLoop       = "interface is a loopbck interface"
+	ErrMsgDeviceNotSpecified        = "no device specified"
+	ErrMsgDeviceNotFound            = "device not found"
+	ErrMsgNoCredentials             = "no SSH authorized keys or passwords are set"
+)
+
+type ValidatorInterface interface {
```
innobead, line 26: These two methods not really like exported general functions for interface. I would suggest to do the below way, then it should be better for adding other validators in the future. WDYT?

type Validator interface {
  Validate() error
}

type HostValidator struct {} which has internal functions like checkMgmtInterface and checkDevice being called in Valdiate.

innobead (approved):
LGTM, thanks for the effort. 

Just few comments to resolve, but not really impact the approval.

pkg/console/validator.go
bk201: Fixed!

k3os/install.sh
```diff
@@ -13,6 +13,7 @@ get_url()
 {
     FROM=$1
     TO=$2
+    echo "Fetching ${FROM} to ${TO}..."
```
gitlawr, line 16: If a change in the k3os directory is needed, put it in a separate commit

bk201: Update in https://github.com/rancher/harvester-installer/pull/33/commits/126f269b0e1a3fa67efcc4363c344d710fe68dc5

gitlawr:
Check if we want to address [this comment](https://github.com/rancher/harvester/pull/478/files#r579614103). Other than that looks good to me

bk201:
> Check if we want to address [this comment](https://github.com/rancher/harvester/pull/478/files#r579614103). Other than that looks good to me

The files are included. Please see 386bdf3 and babeb94

gitlawr (approved):
LGTM

PR #34, opened 2021-01-30
Use yq to promote node

futuretea:
1. add `yq`
2. (harvester) after promote node to master, use `yq` to add `--disable local-storage` into k3sArgs
3. use `rc-service ccapply restart ` to generate /etc/init.d/k3s-service and restart k3s

**Related Issue:**
https://github.com/rancher/harvester/issues/455
https://github.com/rancher/harvester/issues/259


gitlawr (approved):
LGTM

PR #35, opened 2021-02-02
Fix current status

futuretea:
1.  rancher/harvester#462 
    After restart node, pods of `helm-install-harvester` will be deleted, the current status will alway  `Setting up Harvester`,
    Use `get job` instead of `get po` and only the first node check if `Setting up  Harvester`.

2.  rancher/harvester#458
   If  harvester pod has multiple replicas, the current status will show `Runing Runing`,
   `sort -u`

3. if harvester pod's status phase is empty, show nothing instead of `Unknown`

### DIFF
--- a/pkg/console/dashboard_panels.go
+++ b/pkg/console/dashboard_panels.go
@@ -7,15 +7,17 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


gitlawr (approved):
LGTM

PR #38, opened 2021-02-05
Support http proxy and exclude offline images

futuretea:
Support configure http_proxy and exclude offline images by env to speed up local  build.

Docker Desktop eg:
```bash
export http_proxy=http://host.docker.internal:1087
export https_proxy=http://host.docker.internal:1087
export HARVESTER_INSTALLER_OFFLINE_BUILD=1
make
```

### DIFF
--- a/k3os/install.sh
+++ b/k3os/install.sh
@@ -158,50 +158,50 @@ do_copy()
 
     offline_image_path="var/lib/rancher/k3s/agent/images/harvester-images.tar"
     if [ -f "${root_path}/${offline_image_path}.zst" ]; then
-        echo "Decompressing container images"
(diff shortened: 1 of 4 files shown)


k3os/install.sh
```diff
@@ -172,28 +171,31 @@ do_copy()
...
-    do
-      sleep 1
-    done
-    # import images
-    ctr -n k8s.io images import /var/lib/rancher/k3s/agent/images/harvester*
-    rm /var/lib/rancher/k3s/agent/images/harvester*
-    # stop containerd
-    pkill containerd
-    exit
+    if [ -f "${root_path}/${offline_image_path}" ]; then
```
gitlawr, line 174: The whole image handling(including the mount and unmount) can be wrapped in the if statement in L160

gitlawr (approved):
LGTM

PR #40, opened 2021-02-07
Change Service type to NodePort

futuretea:
Change Service type to NodePort and use 30443

**Related Issue:**
rancher/harvester#459

**Related PR:**
rancher/harvester#468


gitlawr (approved):
LGTM

PR #41, opened 2021-02-18
Validate server URL when joining existing cluster

c3y1huang:
Checking for server URL availability can ensure invalid IP is discovered at an earlier stage instead of allowing for installation.

Screenshot-2021-02-18-14:17:47

Relate to https://github.com/rancher/harvester/issues/472

### DIFF
--- a/pkg/console/install_panels.go
+++ b/pkg/console/install_panels.go
@@ -227,15 +227,48 @@ func addServerURLPanel(c *Console) error {
 			if serverURL == "" {
 				return c.setContentByName(validatorPanel, "Management address is required")
 			}
-			serverURLV.Close()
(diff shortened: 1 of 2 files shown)


gitlawr:
A 200 from K3s health check endpoint(https://1.2.3.4:6443/ping) would be a better option so that we don't accept arbitrary responses.

gitlawr:
A 200 from K3s health check endpoint(https://1.2.3.4:6443/ping) would be a better option so that we don't accept arbitrary responses.

gitlawr:
Can we use a spinner as we did for ssh keys?

c3y1huang:
@gitlawr thanks! I've made some changes:
- check endpoint(https://1.2.3.4:6443/ping
- use spinner

gitlawr (approved):
LGTM

PR #42, opened 2021-02-22
Support configure static IP 

futuretea:
**DHCP**
image
image

image

**Static**
image
image

**Related Issue:**
rancher/harvester#367
rancher/harvester#470
rancher/harvester#471
rancher/harvester#490


pkg/console/install_panels.go
```diff
@@ -498,6 +546,350 @@ func addNetworkPanel(c *Console) error {
...
+				return err.Error(), nil
+			}
+		}
+		installData.DNSServers = dnsServers
+		cfg.Config.K3OS.DNSNameservers = dnsServerList
+
+		// run-cmd
+		configureNetworkManual := fmt.Sprintf("/sbin/configure-network-manual %s %s %s %s %s",
+			installData.Interface, installData.IP, installData.NetMask, installData.Gateway, strings.Join(dnsServerList, " "))
+		cfg.Config.Runcmd = []string{configureNetworkManual}
```
bk201, line 853: Does the static IP take effect only after installation?
If that's the case, the fetching URL behavior in SSH keys and Cloud Config dialog might not work.

futuretea: During installation，live os will use dhcp to fetch URL.

gitlawr: Some users request to install with no DHCP. Sounds like this does not satisfy their requirements

bk201 (approved):
LGTM. Thanks for adapting the config structure change.

pkg/console/util.go
```diff
@@ -90,14 +90,13 @@ func getRemoteSSHKeys(url string) ([]string, error) {
...
 }
 
-func getFormattedServerURL(addr string) string {
-	if !strings.HasPrefix(addr, "https://") {
-		addr = "https://" + addr
+func getFormattedServerURL(addr string) (string, error) {
+	ipErr := checkIP(addr)
+	domainErr := checkDomain(addr)
+	if ipErr != nil && domainErr != nil {
+		return "", fmt.Errorf("%s is not an valid ip/domain", addr)
```
gitlawr, line 97: s/an/a/

pkg/console/console.go
```diff
@@ -92,6 +93,23 @@ func (c *Console) setContentByName(name string, content string) error {
 	return err
 }
 
+func (c *Console) CloseElement(name string) error {
```
gitlawr, line 96: Seems like we seldom handle the error in the caller. Can we log the error and do no error return for the function?

pkg/console/install_panels.go
```diff
@@ -483,33 +510,313 @@ func addTokenPanel(c *Console) error {
...
+	dnsServersV, err := widgets.NewInput(c.Gui, dnsServersPanel, dnsServersLabel, false)
+	if err != nil {
+		return err
+	}
+
+	networkValidatorV := widgets.NewPanel(c.Gui, networkValidatorPanel)
+
+	gotoPrevPanel := func(c *Console, names ...string) func(g *gocui.Gui, v *gocui.View) error {
+		return func(g *gocui.Gui, v *gocui.View) error {
+			_ = c.CloseElement(networkValidatorPanel)
```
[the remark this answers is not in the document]
futuretea: Indicates that this error was deliberately ignored, not forgotten to deal with.

pkg/console/console.go
futuretea: yes

gitlawr:
LGTM, can you squash the commits?

futuretea:
> LGTM, can you squash the commits?

Updated

gitlawr (approved):

PR #43, opened 2021-02-23
Scale harvester repliacs to 3

futuretea:
Related Issue:
rancher/harvester#459


futuretea:
close in favor of rancher/harvester#486

PR #47, opened 2021-03-03
Fix dashboard

futuretea:
1. Show Running if one pod Running at least
2. Use node ip if get first ready master node's internal ip failed

**Relate issue**
rancher/harvester#459
rancher/harvester#353

**Relate PR**
rancher/harvester#497

### DIFF
--- a/pkg/console/dashboard_panels.go
+++ b/pkg/console/dashboard_panels.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


pkg/console/dashboard_panels.go
```diff
@@ -27,6 +27,12 @@ const (
 	colorYellow
 	colorBlue
 
+	statusRunning   = "Running"
+	statusUnKnown   = "UnKnown"
```
gitlawr, line 31: Unknown

gitlawr (approved):
LGTM

PR #48, opened 2021-03-03
K3OS: support various TTYs

bk201:
- Use getty to spawn the harvester-installer. By doing this, we can define a
  default window dimension. This prevents installer cui initialization
  failure on a serial console.
- Use the last `console=` kernel parameter value as the TTY which the installer
  and Dashboard should be displayed on.
- Do not run `k3os install` in ccapply.

**Relate issue**

### DIFF
--- a/k3os/overlay/libexec/k3os/boot
+++ b/k3os/overlay/libexec/k3os/boot
@@ -2,29 +2,19 @@
 
 setup_ttys()
 {
-    for i in 2 3 4 5 6; do
+    for i in 1 2 3 4 5 6; do
         if [ -e /dev/tty${i} ]; then
             echo 'tty'$i'::respawn:/sbin/getty 38400 tty'$i >> /etc/inittab
             echo tty$i >> /etc/securetty
         fi
     done
 
-    if [ "$K3OS_MODE" != "local" ]; then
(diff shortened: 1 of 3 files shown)


k3os/overlay/libexec/k3os/boot
```diff
@@ -34,14 +24,27 @@ EOF
             ;;
         console=*)
             CONSOLE_SPEC=${x#console=}
-            IFS=, read TTY BAUDRATE _ <<<"${CONSOLE_SPEC},9600"
+            IFS=, read TTY BAUDRATE _ <<<"${CONSOLE_SPEC},38400"
```
gitlawr, line 27: 9600 is the kernel default, any reason we want to change it?

bk201: The intention is to make tty1 use baudrate `38400` like [the default in the k3os](https://github.com/rancher/k3os/blob/4738bb8f7c4bfcb22656e536077d71bd30c1138f/overlay/libexec/k3os/boot#L5).
Will fix! Thanks.

bk201: The default is changed back to `9600`.

gitlawr (approved):
LGTM

PR #51, opened 2021-03-05
Remove unused image and add size check

futuretea:
After disable local-storage, servicelb, traefik, theirs image can be removed

iso file will just 2G

**Related issue**
rancher/harvester#508

### DIFF
--- a/scripts/build
+++ b/scripts/build
@@ -53,8 +53,12 @@ curl ${longhorn_image_url}>>${image_list_file}
 curl ${K3S_IMAGE_URL}>>${image_list_file}
 
 # format image list
-awk -F ':' '{if($2==""){print $1":latest"}else{print $0}}' ${image_list_file} >"${image_list_file}.tmp"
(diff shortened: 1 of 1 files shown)


gitlawr:
Can we add a check to ensure the built image does not exceed the size limit?

futuretea:
> Can we add a check to ensure the built image does not exceed the size limit?

Good idea, I will add it.

futuretea:
> Can we add a check to ensure the built image does not exceed the size limit?

Updated

scripts/build
```diff
@@ -85,8 +89,28 @@ go mod vendor
...
+find "${artifacts_dir}"  \( -name '*.iso' -or -name 'k3os-vmlinuz-*' -or -name 'k3os-initrd-*' \) -type f -print0 |
+  while IFS= read -r -d '' file; do
+      mv "${file}" "${file/k3os/harvester}"
+  done
+
+# check artifacts size
+max_size=2147483647
+find "${artifacts_dir}"  -type f -print0 |
+  while IFS= read -r -d '' file; do
+      size=$(du -k "${file}" | cut -f 1)
```
gitlawr, line 108: This gets the size in KB

gitlawr (approved):
LGTM

PR #52, opened 2021-03-05
Fix mismatching chart version

gitlawr:
package manifests and remove writeFiles configurations

Problem:
1. chart version mismatches the chart manifest spec
2. WriteFiles session is mounted read-only and inappropriate for upgrade.

Issue: https://github.com/rancher/harvester/issues/509


futuretea (approved):
LGTM

PR #53, opened 2021-03-07
Fix backward and validate display bug on install panels

futuretea:
Fix backward and validate display bug on install panels

1. If the user selects Join mode and enters Management Address and then returns and reselects Create mode, the serverURL in the configuration is not cleared.
2. If the user moves between the password and password confirmation box, the display of the password is not preserved.
3. The Proxy address configuration could not be cleared
4. Cannot clear display of sshkey address configuration
5. https://github.com/rancher/harvester/issues/490#issuecomment-791298151

**Related issue**
rancher/harvester#471
rancher/harvester#490

### DIFF
--- a/pkg/console/install_panels.go
+++ b/pkg/console/install_panels.go
@@ -20,11 +20,12 @@ import (
 )
 
 type UserInputData struct {
-	ServerURL  string
-	SSHKeyURL  string
-	Password   string
-	Address    string
-	DNSServers string
+	ServerURL       string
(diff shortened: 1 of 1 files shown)


gitlawr:
Can you describe the problem and the cause that this PR addresses?

futuretea:
> Can you describe the problem and the cause that this PR addresses?

Updated on the pr description

gitlawr (approved):
LGTM

PR #56, opened 2021-03-10
Fix configure network failed

futuretea:
**Related issue**
rancher/harvester#516


gitlawr:
Not sure if this covers all the cases. Can we make sure that even if the setupNetwork fails, we don't exit the console but pop the error to the users?

futuretea:
> Not sure if this covers all the cases. Can we make sure that even if the setupNetwork fails, we don't exit the console but pop the error to the users?

Updated

gitlawr (approved):
LGTM

PR #57, opened 2021-03-10
Allow specifying webhooks to get installer event notifications

bk201:
**Relate issue**
https://github.com/rancher/harvester/issues/469

With this change, the user can add webhooks in the config and the installer will send HTTP
requests to custom URLs on some predefined events.

The user can add webhooks under the `install` key of the [harvester configuration file](https://github.com/rancher/harvester/blob/fef31431c43b3abcde96c8b4209e6cd27a10048d/docs/harvester-configuration.md):

The following is a configuration example:

```yaml
install:
  webhooks:
    - event: SUCCEEDED
      method: GET
      url: http://10.100.0.100/cblr/svc/op/nopxe/system/{{.Hostname}}
    - event: STARTED
      method: GET
      url: https://10.100.0.100/started/{{.Hostname}}
      insecure: true
      basicAuth:
        user: admin
        password: p@assword
    - event: FAILED
      method: POST
      url: http://10.100.0.100/record
      headers:
        Content-Type:
           - 'application/json; charset=utf-8'
      payload: |
        {
          "host": "{{.Hostname}}",
          "device": "hd"
        }
```

Explanation of the fields:

* `event`: Event type to trigger HTTP action on the webhook
  * `STARTED`: The installer just begins to perform installation.
  * `SUCCEEDED`: The installation completes without errors.
  * `FAILED`: The installation is failed.

  Example:

  ```yaml
  event: SUCCEEDED
  ```
- `method`: [HTTP method](https://golang.or.g/pkg/net/http/#pkg-constants)

  Example:

  ```yaml
  method: GET
  ```

- `url`*: The URL to send the HTTP request.

  ```yaml
  url: http://mycompany.com
  ```
- `insecure`:  Do not verify server's certificate if `true`. Default is `false`.

  Example:

  ```yaml
  insecure: true
  ```

- `basicAuth`: Use 'Basic' HTTP Authentication.

  Example: 
  
  ```yaml
  basicAuth:
    user: admin
    password: p@ssword
  ```

- `headers`: Include custom headers in the HTTP requests.

  Example: 
  
  ```yaml
  headers:
    Content-Type:
      - 'application/json; charset=utf-8'
  ```

  Above configuration will includes `Content-Type: application/json; charset=utf-8` in the HTTP request.

  > **NOTE**: Headers like `Content-Length` are automatically included.

- `payload`*: The payload data to be sent with the request.

  Example: 
  
  ```yaml
  payload: |
    {
      "host": "{{.Hostname}}",
      "device": "hd"
    }
  ```

  > **NOTE**: The user might need to set the correct `Content-Type` header with the `headers` configuration to make the server accept the request.

**\*** The values for `url` and `payload` can be written in Go template, the following variables will be substituted with appropriate values of the node that is being installed:

- `Hostname`: the hostname of the node.
- `MACAddr`: the MAC address of the management interface
- `IPAddrV4`: the IPv4 address of the management interface
- `IPAddrV6`: the IPv6 address of the management interface

**Additional information**
- The installer can send multiple requests for a single event. But if one of them fails, the remaining requests will not be sent.
- The installer will retry 3 times with a 5 seconds-sleep if the request is timed out.

### DIFF
--- a/pkg/console/install_panels.go
+++ b/pkg/console/install_panels.go
@@ -1184,7 +1184,12 @@ func addInstallPanel(c *Console) error {
 				printToPanel(c.Gui, err.Error(), installPanel)
 				return
 			}
-			doInstall(c.Gui, toCloudConfig(c.config))
+
+			webhooks, err := PrepareWebhooks(c.config.Webhooks, getWebhookContext(c.config))
+			if err != nil {
+				printToPanel(c.Gui, fmt.Sprintf("invalid webhook: %s", err), installPanel)
+			}
+			doInstall(c.Gui, toCloudConfig(c.config), webhooks)
 		}()
 		return c.setContentByName(footerPanel, "")
 	}
--- a/pkg/console/util.go
+++ b/pkg/console/util.go
@@ -15,8 +15,8 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/jroimartin/gocui"
 	"github.com/pkg/errors"
+	"github.com/rancher/harvester-installer/pkg/util"
 	k3os "github.com/rancher/k3os/pkg/config"
-	"github.com/sirupsen/logrus"
 	"golang.org/x/crypto/ssh"
 	"k8s.io/apimachinery/pkg/util/rand"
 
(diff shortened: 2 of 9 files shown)


pkg/console/install_panels.go
```diff
@@ -1050,7 +1050,12 @@ func addInstallPanel(c *Console) error {
 				printToInstallPanel(c.Gui, err.Error())
 				return
 			}
-			doInstall(c.Gui, toCloudConfig(c.config))
+
+			webhooks, err := PrepareWebhooks(c.config.Webhooks, getWebhookContext(c.config))
+			if err != nil {
+				printToInstallPanel(c.Gui, fmt.Sprintf("invalid webhooks: %s", err))
+			}
```
gitlawr, line 1057: Seems like the PrepareWebhooks function never returns a non-nil error

pkg/console/util.go
```diff
@@ -247,7 +247,37 @@ func toCloudConfig(cfg *config.HarvesterConfig) *k3os.CloudConfig {
...
+	scanner = bufio.NewScanner(stderr)
+	scanner.Split(bufio.ScanLines)
+	for scanner.Scan() {
+		printToInstallPanel(g, scanner.Text())
+	}
+	return cmd.Wait()
+}
+
+func doInstall(g *gocui.Gui, cloudConfig *k3os.CloudConfig, webhooks RendererWebhooks) error {
+	webhooks.Handle(EventInstallStarted)
```
gitlawr, line 279: Handle() may return an error but is ignored by the caller.

guangbochen:
@bk201 can u please rebase the PR and take a look at the above comments, thanks.

pkg/console/util.go
bk201: Thanks, make the function return nothing since we don't want to stop the installer if there are any errors.

pkg/console/install_panels.go
bk201: Fixed. Log the error message and also print it to the install panel.

bk201:
> @bk201 can u please rebase the PR and take a look at the above comments, thanks.

Rebased!

gitlawr (approved):
LGTM

PR #61, opened 2021-03-12
Publish on cron event

gitlawr:
https://github.com/rancher/harvester/issues/503

Do publish on daily cron events to keep artifacts up to date when related components like network-controller are updated.


futuretea (approved):
LGTM

PR #62, opened 2021-03-15
Choose a default TTY if not specified

bk201:
The installer can be run on TTY other than `tty1` after https://github.com/rancher/harvester-installer/pull/48.
But if the user doesn't specify [the TTY key in the harvester configuration](https://github.com/rancher/harvester/pull/478/files#diff-fea66d96528ac5bb18202ca7c217092e39ce1270105bb9e0d3201c354ca20b40R281-R290), the TTY that the provisioned OS uses is still `tty1.`

This change ensures the provisioned OS uses the same TTY as the installer one if a `tty` key is not specified.

Related issue:


gitlawr (approved):
LGTM

PR #64, opened 2021-03-18
Add rancher deployment

guangbochen:
**Problem:**
Add Rancher deployment manifest to the Harvester and enable by default on HCI mode.

**Solution:**
1. add Rancher & cert-manager manifests
2. expose Rancher by NodePort with the same IP address as the Harvester
3. package offline images

**Related Issue:**
https://github.com/rancher/harvester/issues/513

### DIFF
--- a/scripts/build
+++ b/scripts/build
@@ -10,6 +10,10 @@ echo "Start building ISO"
 K3S_VERSION=v1.20.4+k3s1
 K3S_IMAGE_URL=https://raw.githubusercontent.com/rancher/k3s/${K3S_VERSION}/scripts/airgap/image-list.txt
 
+RANCHER_VERSION=2.5.7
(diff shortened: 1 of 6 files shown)


scripts/build
```diff
@@ -57,6 +77,35 @@ curl ${longhorn_image_url}>>${image_list_file}
...
+  do
+    if [[ $img == "$img_rq"* ]]; then
+      echo "$img" >> ${image_list_file}
+    fi
+  done < "$RANCHER_IMAGES"
+done < "$RANCHER_REMOTE_IMAGES"
+echo "Done adding rancher images"
+
+# add cert-manager image list
+CERT_MANAGER_IMAGES=scripts/images/cert-manager-images.txt
```
gitlawr, line 102: Can we generate the image list with $CERT_MANAGER_VERSION so we don't need to update both the script and the image files when we bump cert-manager version?

gitlawr (approved):
LGTM

PR #65, opened 2021-03-18
Update termbox-go dependency

bk201:
The version opens TTY device in non-blocking mode and it can prevent
stuck problem on serial consoles.

**Related issue**


gitlawr (approved):
LGTM

PR #72, opened 2021-03-31
K3OS: use install mode if k3s.mode is not specified

bk201:

### DIFF
--- a/k3os/overlay/libexec/k3os/mode
+++ b/k3os/overlay/libexec/k3os/mode
@@ -44,8 +44,11 @@ fi
 done
 
 if [ -z "$MODE" ]; then
-    pfatal "Failed to determine boot mode (did you forget to set k3os.mode?)"
-elif [ ! -e $SCRIPTS/mode-${MODE} ]; then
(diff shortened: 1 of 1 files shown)


k3os/overlay/libexec/k3os/mode
```diff
@@ -44,8 +44,11 @@ fi
 done
 
 if [ -z "$MODE" ]; then
-    pfatal "Failed to determine boot mode (did you forget to set k3os.mode?)"
-elif [ ! -e $SCRIPTS/mode-${MODE} ]; then
+  echo "Failed to determine boot mode. Fallback to install mode."
```
gitlawr, line 47: Can we set FALLBACK_MODE to "install"(before the mode-wait loop) if it is not specified?

bk201: Won't this make the loop only run the first iteration? It's possible for a disk that contains the `HARVESTER_STATE` label to appear later. We might want the loop to run 30 times to make sure there is no disk that contains the label.

bk201:
Closed. See https://github.com/rancher/harvester/pull/478#discussion_r607614885

PR #74, opened 2021-04-07
Fix proxy

futuretea:
Currently, proxy only works after installation, now it only works during the installation process.

**Related issue**
rancher/harvester#646


gitlawr (approved):
LGTM

PR #81, opened 2021-04-27
k3os: always copy k3os for upgrade

gitlawr:
https://github.com/harvester/harvester/issues/383

Problem:
After an upgrade completes, the terminal console still shows the previous version. k3os does not upgrade itself if the version does not change. 

Solution:
Since we compile the console into the k3os binary we need to upgrade it anyway. Also, rename the previous directory to a unique name so that it does not conflict after multiple upgrades.


futuretea (approved):
LGTM

guangbochen (approved):

PR #85, opened 2021-04-30
Fix invalid cert-manager chart

guangbochen:
Fix invalid cert-manager chart
https://drone-publish.rancher.io/harvester/harvester/275/3/2


futuretea (approved):
LGTM

PR #86, opened 2021-05-13
Support write_files in Harvester configuration

bk201:
Add `os.write_files`.

Doc is also updated in https://github.com/harvester/harvester/pull/862

Related Issues:
https://github.com/harvester/harvester/issues/690


futuretea (approved):
LGTM

PR #87, opened 2021-06-08
Retry if fetching remote config fails

bk201:
It's possible that the system's network stack is not ready before the installer is started. Add a retry loop to workaround this situation.

The retry interval is 10 seconds and the retries count is 30 times (5 minutes timeout).

**Related issue**:
https://github.com/harvester/harvester/issues/907

**Screenshot**
Screen Shot 2021-06-08 at 4 01 54 PM

### DIFF
--- a/pkg/console/util.go
+++ b/pkg/console/util.go
@@ -17,6 +17,7 @@ import (
 	"github.com/jroimartin/gocui"
 	"github.com/pkg/errors"
 	k3os "github.com/rancher/k3os/pkg/config"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/crypto/ssh"
 	"golang.org/x/net/http/httpproxy"
 	"k8s.io/apimachinery/pkg/util/rand"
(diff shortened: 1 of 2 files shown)


pkg/console/util.go
```diff
@@ -398,6 +410,42 @@ func getRemoteConfig(configURL string) (*config.HarvesterConfig, error) {
...
+			logrus.Error(e)
+			printToPanel(g, e.Error(), installPanel)
+			printToPanel(g, fmt.Sprintf("Retry after %d seconds (Remaining: %d)...", interval, remainingRetries), installPanel)
+		}
+		return e
+	})
+
+	if err != nil {
+		msg := fmt.Sprintf("Fail to fetch config: %v", err)
+		logrus.Error(msg)
```
gitlawr, line 434: Log the error in the caller. If you want to add more info just return a wrapped error: fmt.Errorf("Fail to fetch config: %w", err)

pkg/console/util.go
```diff
@@ -398,6 +410,42 @@ func getRemoteConfig(configURL string) (*config.HarvesterConfig, error) {
...
+		printToPanel(g, msg, installPanel)
+		return nil, err
+	}
+
+	harvestCfg, err := config.LoadHarvesterConfig(confData)
+	if err != nil {
+		msg := fmt.Sprintf("Fail to load config: %v", err)
+		logrus.Error(msg)
```
gitlawr, line 442: ditto

pkg/console/util.go
```diff
@@ -73,15 +74,15 @@ func validatePingServerURL(url string) error {
...
 		_, err := getURL(client, url)
 		return err
 	})
 }
 
-func retryOnError(retryNum, retryInterval int64, process func() error) error {
+func retryOnError(retryNum, retryInterval int64, process func(remainingRetries int64) error) error {
 	for {
-		if err := process(); err != nil {
+		if err := process(retryNum); err != nil {
```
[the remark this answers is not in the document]
bk201: To output information messages.
But I can do it in the caller.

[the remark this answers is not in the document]
bk201: Removed.

pkg/console/util.go
bk201: Move up to the caller.

gitlawr (approved):
LGTM

PR #89, opened 2021-06-21
Fix webhook pods fail to start in air gapped environment

bk201:
The pull policy need to be "IfNotPresent" to use the pre-loaded images.

Related issue: https://github.com/harvester/harvester/issues/962


gitlawr (approved):
LGTM

PR #91, opened 2021-07-05
Bump rancher to master-head

gitlawr:
https://github.com/harvester/harvester/issues/922

Depend on PR: https://github.com/harvester/harvester/pull/998

Pinned to a specific commit in case upstream master-head is broken or has incompatible changes.


guangbochen (approved):
LGTM

PR #88, opened 2021-06-16
Wrap the install panel to display entire output

weihanglo:
Related issue harvester/harvester#800.

image

PR #94, opened 2021-07-19
Autoscroll install panel

bk201:
We need to adjust Y-axis movement because lone-lines are wrapped in the install panel (new feature in https://github.com/harvester/harvester-installer/pull/88.).
The `Autoscroll` flag of a View seems to work well, I remove the manually movement and turn the flag on.

### Screenshots (the outputs are mocked messages)
**Before**
before
**After**

after

### Test plans
- Install harvester and make sure no messages are dropped in the install (last) panel.

Developer can apply this [patch](https://gist.github.com/bk201/40d8655f0a5f08bc95ae0f585ee23d43) to artificially generate some output without really running the installation.

### DIFF
--- a/pkg/console/install_panels.go
+++ b/pkg/console/install_panels.go
@@ -1251,6 +1251,7 @@ func addInstallPanel(c *Console) error {
 	installV.Title = " Installing Harvester "
 	installV.SetLocation(maxX/8, maxY/8, maxX/8*7, maxY/8*7)
 	installV.Wrap = true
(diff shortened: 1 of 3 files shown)


weihanglo (approved):
Tested and looks great.

https://user-images.githubusercontent.com/14314532/126153227-0c8a5369-971b-406b-99b9-5438c1554b97.mov

gitlawr (approved):
LGTM

PR #100, opened 2021-08-17
Fix feature flag

gitlawr:


guangbochen (approved):
LGTM, can u please merge it after https://github.com/harvester/harvester/pull/1098, so we can fix the network-controller bug at the same time, thanks.

PR #108, opened 2021-09-01
Manage Harvester chart with RKE2 HelmChart CR and bump Rancher/RKE2 versions

bk201:
- Manage Harvester chart with HelmChart CRD in RKE2
  - Harvester-repo is removed.
  - MangedChart is removed.
- Bump Rancher to `v2.6.0`
- Bump RKE2 to `v1.21.4+rke2r2`


manifests/harvester.yaml
```diff
@@ -0,0 +1,40 @@
...
+    harvester-network-controller:
+      enabled: true
+      image:
+        pullPolicy: "IfNotPresent"
+    rancherEmbedded: true
+    webhook:
+      image:
+        repository: rancher/harvester-webhook
+        tag: master-head
+        imagePullPolicy: "IfNotPresent"
```
guangbochen, line 40: can we use the set option like [previously](https://github.com/harvester/harvester-installer/blob/b0a0ebdaf2db53b3289d6c8624c7ff85ae180768/manifests/harvester.yaml#L15-L23) and do not include those default image values since they were defined by the chart we have packaged in.

bk201: Is there any difference between the `valuesContent` and `set` ?

guangbochen: Both will work for the same result, mostly we use `set` to overwrite simple default values and `valuesContent` for complex default values. Since both are working so the important stuff is we need to remove those image tags.

gitlawr (approved):
LGTM

manifests/harvester.yaml
bk201: I keep `imagePullPolicy: IfNotPresent` lines, they are mostly `Always` in the chart values file.

bk201:
I update the way how dashboard check status too.
https://github.com/harvester/harvester-installer/pull/108/files#diff-602bf17c3838a0b8cb3d90dd408056f7c2f210bdef589f9e7b1beaaee1b24dd8

It should check the helm chart install job rather than MangaedChart anymore.

guangbochen (approved):
LGTM

PR #109, opened 2021-09-02
Fix netconfig fail to update name servers in target OS

bk201:
The previous way to set name servers in the image preloading phase
removes the symbolic link:

```
 /etc/resolv.conf -> /var/run/netconfig/resolv.conf
```

This makes netconfig fail to update name servers from DHCP protocol.
Using netconfig to keep the link.

**Depends on**: https://github.com/harvester/os2/pull/8
**Related**: harvester/harvester#1137


gitlawr:
Does it work for manually configured NIC?

bk201:
> Does it work for manually configured NIC?

After some tests, no. Good catch. I'll improve it.

gitlawr (approved):
LGTM

guangbochen (approved):

PR #111, opened 2021-09-03
Remove Rancher expose svc and other configs

guangbochen:
Depending on https://github.com/harvester/harvester/pull/1152

- remove auth config since the UI is authenticated with the Rancher server
- remove rancher expose svc and make it deployed on the Harvester side when the server is ready
- disable cloud controller to save resources as it is not used by the Harvester

https://github.com/harvester/harvester/issues/1139


pkg/config/templates/rke2-90-harvester-server.yaml
```diff
@@ -3,3 +3,4 @@ disable: rke2-ingress-nginx
 cluster-cidr: 10.52.0.0/16
 service-cidr: 10.53.0.0/16
 cluster-dns: 10.53.0.10
+disable-cloud-controller: true
```
bk201, line 6: Do we also need this on promoted nodes?
https://github.com/harvester/harvester/blob/c4584348c05d4c0923fadf69f95abdb481dc0bae/pkg/controller/master/node/promote_controller.go#L48-L53

guangbochen: will it be re-enabled if not set on the node promotion? if yes then probably will need to ad it as well.

guangbochen: added via https://github.com/harvester/harvester/pull/1152/commits/fc575c596fca5154b87b918209eea8cc852864d6

bk201 (approved):

gitlawr (approved):
LGTM

PR #114, opened 2021-09-09
Fix VIP expose svc

yaocw2020:
- Fix VIP expose svc
- Support case-insensitive value for network modes

issue: 
https://github.com/harvester/harvester/issues/1186
https://github.com/harvester/harvester/issues/1187

Related PR:
https://github.com/harvester/harvester/pull/1188

### DIFF
--- a/pkg/config/templates/harvester-config.yaml
+++ b/pkg/config/templates/harvester-config.yaml
@@ -9,7 +9,8 @@ spec:
       config:
         vip_interface: "{{ .MgmtInterface }}"
     service:
-      harvester:
-        vipMode: "{{ .VipMode }}"
-        vip: "{{ .Vip }}"
-        vipHwAddr: "{{ .VipHwAddr }}"
+      vip:
+        enabled: true
+        mode: "{{ .VipMode }}"
+        ip: "{{ .Vip }}"
+        hwAddress: "{{ .VipHwAddr }}"
--- a/pkg/console/dashboard_panels.go
+++ b/pkg/console/dashboard_panels.go
@@ -213,9 +213,15 @@ func syncManagementURL(ctx context.Context, g *gocui.Gui) {
 
 func doSyncManagementURL(g *gocui.Gui) {
 	managementURL := "Unavailable"
-	if managementIP := getVIP(); managementIP != "" {
+	managementIP := getVIP()
+	if managementIP != "" {
 		managementURL = fmt.Sprintf("https://%s", managementIP)
+	} else {
+		if managementIP = getFirstReadyMasterIP(); managementIP != "" {
+			managementURL = fmt.Sprintf("https://%s:%s", managementIP, harvesterNodePort)
+		}
 	}
+
 	g.Update(func(g *gocui.Gui) error {
 		v, err := g.View("url")
 		if err != nil {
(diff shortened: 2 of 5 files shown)


pkg/config/templates/harvester-config.yaml
```diff
@@ -9,7 +9,7 @@ spec:
       config:
         vip_interface: "{{ .MgmtInterface }}"
     service:
-      harvester:
-        vipMode: "{{ .VipMode }}"
-        vip: "{{ .Vip }}"
-        vipHwAddr: "{{ .VipHwAddr }}"
+      vip:
```
guangbochen, line 12: `enabled: true` is required here

gitlawr (approved):
LGTM

pkg/config/templates/harvester-config.yaml
```diff
@@ -9,7 +9,8 @@ spec:
...
+        enabled: true
+        mode: "{{ .VipMode }}"
```
bk201, line 14: Do we need to convert `dhcp` to `DHCP` here?

yaocw2020: We use lower case in the harvester chart.
https://github.com/harvester/harvester/pull/1188

bk201: Got it, thanks!

bk201 (approved):

gitlawr (approved):

pkg/console/dashboard_panels.go
```diff
@@ -213,9 +213,15 @@ func syncManagementURL(ctx context.Context, g *gocui.Gui) {
...
 func doSyncManagementURL(g *gocui.Gui) {
 	managementURL := "Unavailable"
-	if managementIP := getVIP(); managementIP != "" {
+	managementIP := getVIP()
+	if managementIP != "" {
 		managementURL = fmt.Sprintf("https://%s", managementIP)
+	} else {
+		if managementIP = getFirstReadyMasterIP(); managementIP != "" {
+			managementURL = fmt.Sprintf("https://%s:%s", managementIP, harvesterNodePort)
+		}
```
gitlawr, line 222: On second thought, I think we should get the VIP from the vip configmap. The service is not ready until harvester is up and running and it could take some time. The console probably will show "Unavailable" => "https://some-node-ip:30443" => "https://vip". The switchover seems bad to users.

bk201:
@yaocw2020 For the network mode case-sensitive part, this PR fixes the validation part.
But the original value in the config is still used to generate templates.

Suppose the user uses the value `DHCP`, it will pass the validation. But the generated helm chart config will be:

```
    service:
      vip:
        enabled: true
        mode: "DHCP"
        ip: "10.10.0.100"
        hwAddress: "52:54:00:ec:0e:0b"
```

Will this work with the change in https://github.com/harvester/harvester/pull/1188 ?

guangbochen:
> @yaocw2020 For the network mode case-sensitive part, this PR fixes the validation part.
> But the original value in the config is still used to generate templates.
> 
> Suppose the user uses the value `DHCP`, it will pass the validation. But the generated helm chart config will be:
> 
> ```
>     service:
>       vip:
>         enabled: true
>         mode: "DHCP"
>         ip: "10.10.0.100"
>         hwAddress: "52:54:00:ec:0e:0b"
> ```
> 
> Will this work with the change in [harvester/harvester#1188](https://github.com/harvester/harvester/pull/1188) ?

https://github.com/harvester/harvester/pull/1188/files#diff-891b2975a5b9aa41d64348b35f9f1233f219e7145962ac572c0a51f61b58fe3aR107

bk201:
I see, thanks.
It seems `Static` or `STATIC` will not work:
https://github.com/harvester/harvester/pull/1188/files#diff-b8916dccf848e9983fc1caeded29a01d766b8e0191877b6b857808f7e43ab54fR7

PR #115, opened 2021-09-09
Fix iPXE installing when VIP is not configured

voima-eetu:
Fixes https://github.com/harvester/harvester/issues/1194

### DIFF
--- a/pkg/console/validator.go
+++ b/pkg/console/validator.go
@@ -200,7 +200,7 @@ func (v ConfigValidator) Validate(cfg *config.HarvesterConfig) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/console/validator.go
```diff
@@ -200,7 +200,7 @@ func (v ConfigValidator) Validate(cfg *config.HarvesterConfig) error {
 		return err
 	}
 
-	if cfg.Install.Mode == config.ModeCreate {
```
guangbochen, line 203: since VIP is required so u should add pre-check validation instead

bk201:
The main issue was closed: https://github.com/harvester/harvester/issues/1194
Please let us know if you see any issues, thanks for the contribution.

PR #118, opened 2021-09-10
Make network interface names predictable

bk201:
**Related issue**
https://github.com/harvester/harvester/issues/1197

**NOTE**: The UX in the network config dialog needs to be improved.
It takes time to configure IPs, using a spinner will be better.

### DIFF
--- a/package/harvester-os/files/etc/cos/bootargs.cfg
+++ b/package/harvester-os/files/etc/cos/bootargs.cfg
@@ -1,9 +1,9 @@
 set console_params="console=tty1"
 set kernel=/boot/vmlinuz
 if [ -n "$recoverylabel" ]; then
-    set kernelcmd="$console_params root=live:LABEL=$recoverylabel rd.live.dir=/ rd.live.squashimg=$img panic=5"
+    set kernelcmd="$console_params root=live:LABEL=$recoverylabel rd.live.dir=/ rd.live.squashimg=$img panic=5 net.ifnames=1"
 else
-    set kernelcmd="$console_params root=LABEL=$label cos-img/filename=$img panic=5"
+    set kernelcmd="$console_params root=LABEL=$label cos-img/filename=$img panic=5 net.ifnames=1"
 fi
 
 set initramfs=/boot/initrd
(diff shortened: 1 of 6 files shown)


gitlawr:
We used the mgmtInterface as the `kube-vip.config.vip_interface` in helmchartconfig. Need to evaluate the impact on vip side with this change @yaocw2020

guangbochen (approved):
LGTM

package/harvester-os/files/etc/cos/bootargs.cfg
```diff
@@ -1,9 +1,9 @@
 set console_params="console=tty1"
 set kernel=/boot/vmlinuz
 if [ -n "$recoverylabel" ]; then
-    set kernelcmd="$console_params root=live:LABEL=$recoverylabel rd.live.dir=/ rd.live.squashimg=$img panic=5"
+    set kernelcmd="$console_params root=live:LABEL=$recoverylabel rd.live.dir=/ rd.live.squashimg=$img panic=5 net.ifnames=1"
 else
-    set kernelcmd="$console_params root=LABEL=$label cos-img/filename=$img panic=5"
+    set kernelcmd="$console_params root=LABEL=$label cos-img/filename=$img panic=5 net.ifnames=1"
```
yaocw2020, line 6: Is It the key modification to set net.ifnames as 1 to make the network interfaces predictable? Referring to https://man7.org/linux/man-pages/man8/systemd-udevd.service.8.html, it seems that the default value of net.ifnames is 1.
```
       net.ifnames=
           Network interfaces are renamed to give them predictable names
           when possible. It is enabled by default; specifying 0
           disables it.
```

guangbochen:
additionally, the time to get the IP of each NIC is comparatively slow, can we either improve it or add spinner for now, thanks

package/harvester-os/files/etc/cos/bootargs.cfg
bk201: it disabled by default in leap 15.3 from my test.

yaocw2020:
> We used the mgmtInterface as the `kube-vip.config.vip_interface` in helmchartconfig. Need to evaluate the impact on vip side with this change @yaocw2020

It is supposed to have no impact on the VIP.

yaocw2020 (approved):

alexdepalex:
Adding net.ifnames should also be included in the ipxe-examples.

bk201:
> Adding net.ifnames should also be included in the ipxe-examples.

https://github.com/harvester/ipxe-examples/pull/14
A PR is created, thanks!

PR #119, opened 2021-09-12
feat(install_panels): add NTP servers panel

FrankYang0529:
related https://github.com/harvester/harvester/issues/935

### DIFF
--- a/pkg/console/util.go
+++ b/pkg/console/util.go
@@ -3,8 +3,10 @@ package console
 import (
 	"bufio"
 	"crypto/tls"
+	"encoding/binary"
 	"fmt"
 	"io/ioutil"
+	"log"
 	"net"
 	"net/http"
 	"net/url"
@@ -97,6 +99,29 @@ func validateNTPServers(ntpServerList []string) error {
 		// RFC: https://datatracker.ietf.org/doc/html/rfc4330
 		conn, err := net.Dial("udp", fmt.Sprintf("%s:%s", host, port))
 		if err != nil {
+			return err
+		}
+		if err := conn.SetDeadline(time.Now().Add(5 * time.Second)); err != nil {
+			return err
+		}
+
+		// RFC: https://datatracker.ietf.org/doc/html/rfc4330#section-4
+		// NTP Packet is 48 bytes and we set the first byte for request.
+		// 00 100 011 (or 0x2B)
+		// |  |   +-- client mode (3)
+		// |  + ----- version (4)
+		// + -------- leap year indicator, 0 no warning
+		req := make([]byte, 48)
+		req[0] = 0x2B
+
+		// send time request
+		if err := binary.Write(conn, binary.BigEndian, req); err != nil {
+			log.Fatalf("failed to send request: %v", err)
+		}
+
+		// block to receive server response
+		rsp := make([]byte, 48)
+		if err := binary.Read(conn, binary.BigEndian, &rsp); err != nil {
 			return err
 		}
 		conn.Close()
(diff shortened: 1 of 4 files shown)


bk201:
Please don't use [linking keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) in the PR description/commit/comments. We want some verification before closing the issue.

bk201:
I wonder if we should still allow users to finish the installation if the NTP server can't be reached.
If the cluster is in an air-gapped environment, the NTP server can't be reached.

FrankYang0529:
Probably, we can just valid whether the domain is FQDN. Also, we can allow users to input empty string.

bk201:
The NTP server can be an IP address (e.g. A local server).
How about displaying a warning dialog to the user if the entered server can't be reached, but still allow the user to continue?


> Probably, we can just validate whether the domain is a FQDN. Also, we can allow users to input empty string.

pkg/console/util.go
```diff
@@ -79,6 +80,35 @@ func validatePingServerURL(url string) error {
 	})
 }
 
+func validateNTPServers(ntpServerList []string) error {
+	if len(ntpServerList) == 0 {
+		return fmt.Errorf("NTP Server should not be empty.")
+	}
```
gitlawr, line 86: Should it be optional? We tend to show optional inputs after the required ones so that users can easily skip them

FrankYang0529: It's optional now and I put it after the password panel. Thank you.

FrankYang0529:
Hi @bk201 @gitlawr , I updated it. Please help me review again. Thank you 😄 .

bk201:
Looks good! But there is a minor issue:
- Enter an NTP server that can't be reached
- It prompts an error and I can press enter to conintue.
- Press ESC to go back
- If I press enter again, the NTP server is not validated again.

FrankYang0529:
I will reset `c.config.OS.NTPServers` when users press ESC. Thank you.

bk201:
> I will reset `c.config.OS.NTPServers` when users press ESC. Thank you.

I meant going back from Proxy page to NTP page. The validation is not performed.

gitlawr:
>If I press enter again, the NTP server is not validated again.

Seems like the problem still exists if I go back from the proxy page.

FrankYang0529:
@bk201 @gitlawr  Users already press Enter to ignore unreachable NTP servers. Do we need to validate it again after users go back from the proxy panel?

bk201:
The user can change his mind and use a different NTP server, it would be good if validation is done again.

FrankYang0529:
> The user can change his mind and use a different NTP server, it would be good if validation is done again.

Yes, I think we already do it now. Like following steps:

* Users input `error.ntp` and press Enter twice to ignore error.
* Users press Esc to go back NTP servers panel.
* Users change `error.ntp` to `another.error.ntp` and press Enter. We check the input again.
* Users press Enter again to go to the proxy panel and press Esc to go back to NTP servers panel.
* Users change `another.error.ntp` to `error.ntp`. We still check the input again here.

So we revalidate when users change the input, but we go through if users have already ignored the error.

bk201:
Screen Recording 2021-09-15 at 12 37 25 PM (1)
Looks like it doesn't validate the server provided in IP form.

FrankYang0529:
> Screen Recording 2021-09-15 at 12 37 25 PM (1)
> Looks like it doesn't validate the server provided in IP form.

Thanks for indicating the error. I send real NTP request now, not only test server connection. Thank you 😄

gitlawr (approved):
LGTM

PR #120, opened 2021-09-13
Optimize function getIPThroughDHCP

yaocw2020:
issue: https://github.com/harvester/harvester/issues/1209

Use the nclient4 package to get the IP through DHCP instead of executing the wicked command.


guangbochen (approved):
LGTM

gitlawr:
Can you explain why this solution resolves the issue?

yaocw2020:
> Can you explain why this solution resolves the issue?

There are two reasons:
- To avoid parsing wicked output.
- Generally, the DHCP server should have a record about the vip after installation before kube-vip applying, but `wickedd-dhcp4` does not. It may cause kube-vip fail to requests the vip.

pkg/console/vip.go
```diff
@@ -21,21 +25,23 @@ func createMacvlan(name string) (netlink.Link, error) {
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to fetch %s", name)
 	}
+	rand.Seed(time.Now().UnixNano())
+	macvlanName := tempMacvlanPrefix+strconv.Itoa(rand.Intn(100))
```
gitlawr, line 29: What happens if we have a conflict name?  A random integer in 100 doesn't look safe

yaocw2020: I think 100 is far enough. After requesting the VIP, the temporary macvlan device will be deleted immediately. Unless users add network interfaces on purpose, there would not have a conflict network interface. With a very small probability, there will be an error message on the console if a conflict occurs.

gitlawr (approved):
LGTM

bk201 (approved):
The change also fixes the problem that I can't get VIP from dnsmasq after rebooting.

PR #121, opened 2021-09-13
Create booting interface configuration files in PXE boot installation

bk201:
Create booting interface configuration files in PXE boot installation

The configuration files are needed to let wicked correctly apply DNS
settings.

For booting parameter `ip=dhcp`, the script creates DHCP configuration files for the booting interface(s).
If a machine has multiple network interfaces and more than one of them receive DHCP IPs, the booting interfaces are undetermined. Sometimes it's the first interface that receives a DHCP IP, sometimes all interfaces are listed.

For booting parameter `ip=ens5:dhcp ip=ens6:dhcp`, the script creates DHCP configuration files for interface `ens5` and `ens6`

**Why do we need this?**
A PXE-booted Live OS can be divided into 3 stages, each has its own network configuration.
- PXE firmware: The firmware sends DHCP requests to get an IP to download iPXE config, kernel, and initrd back. The kernel is booted.
- Initramfs: If the iPXE config downloaded from the previous stage contains kernel parameter `ip=dhcp`, a Dracut module will try to send DHCP requests again. If it gets IP, it will download squashfs and pivot to it. At this stage, gateway information and DNS server offered by DHCP server are correctly configured.
- Switch to real rootfs (the squashfs image): wicked is in charge of the network. But because the rootfs is changed, it no longer contains the correct DNS server information in the previous stage. We need to tell wicked to make DHCP requests again and it will correctly set the DNS servers. That's why we need to generate wicked configs.

**Related issue**
https://github.com/harvester/harvester/issues/1214

### DIFF
--- a/package/harvester-os/files/system/oem/90_network.yaml
+++ b/package/harvester-os/files/system/oem/90_network.yaml
@@ -1,7 +1,27 @@
-name: "Bring up interface"
+name: "Network configuration"
 stages:
+  rootfs:
+    - name: Create DHCP config for boot interfaces
+      if: cat /proc/cmdline | grep -q "harvester.install.automatic=true"
+      commands:
+      - |
+        if [ ! -e /tmp/net.ifaces ]; then
+          echo "net.ifaces is not found. Skip configuring boot interface."
+          exit 0
+        fi
+        
+        for iface in $(</tmp/net.ifaces); do
+          echo "Create DHCP config for $iface"
+        
+          cat > /sysroot/etc/sysconfig/network/ifcfg-$iface <<'EOF'
+        BOOTPROTO='dhcp'
+        STARTMODE='onboot'
+        EOF
+        
+        done
   initramfs:
-    - commands:
+    - name: Bring up interfaces
+      commands:
       - |
         for i in /sys/class/net/{eth*,en*,ib*}; do
           [ -e $i ] || continue


gitlawr (approved):
LGTM

yaocw2020:
We update wicked configuration here, so why do we have to update the configuration again?
https://github.com/harvester/harvester-installer/blob/e69db50cd254777515aafccf0ce158bead9849aa/pkg/config/cos.go#L96

bk201:
> We update wicked configuration here, so why do we have to update the configuration again?
> https://github.com/harvester/harvester-installer/blob/e69db50cd254777515aafccf0ce158bead9849aa/pkg/config/cos.go#L96

These configuration files take effect after reboot.
What the PR adds are configuration files in the Live OS, PXE boot skips to the last panel, and the files are not created.

package/harvester-os/files/system/oem/90_network.yaml
```diff
@@ -1,7 +1,27 @@
...
+        if [ ! -e /tmp/net.ifaces ]; then
+          echo "net.ifaces is not found. Skip configuring boot interface."
+          exit 0
+        fi
+        
+        for iface in $(</tmp/net.ifaces); do
+          echo "Create DHCP config for $iface"
+        
+          cat > /sysroot/etc/sysconfig/network/ifcfg-$iface <<'EOF'
+        BOOTPROTO='dhcp'
```
yaocw2020, line 17: In PXE boot installation, Do we always get the IP through DHCP and not support static mode?

bk201: Yeah, that's what we have in the document so far.
PXE boot is supposed to work within a DHCP environment.

yaocw2020 (approved):

yaocw2020 (approved):
LGTM

PR #123, opened 2021-09-15
Report status while configuring network

johnliu55tw:
**Related issue**
https://github.com/harvester/harvester/issues/1212

* Add functionality to the spinner for trapping the cursor focus so
  users can't adjust settings while we're configuring the system:
  `NewFocusSpinner`.

* Set up the focus spinner while we're requesting IP through DHCP and
  setting up networking.

* For other types of message, call `updateValidatorMessage` to prevent
  trapping focus.

* Fix spinner behavior: When it can't find the view, don't do anything.

* Update `validateDHCPAddresses`: Calls `getIPThroughDHCP` to validate
  DHCP. No need to call `setupNetwork` at this time because it will be
  called when we go to next page.

**Note**
I didn't change the way how IP is shown after it's received from DHCP.

### DIFF
--- a/pkg/console/install_panels.go
+++ b/pkg/console/install_panels.go
@@ -765,18 +765,6 @@ func addNetworkPanel(c *Console) error {
 	c.AddElement(askInterfacePanel, askInterfaceV)
 
 	// askNetworkMethodV
-	validateDHCPAddresses := func() (string, error) {
-		if mgmtNetwork.Method == config.NetworkMethodStatic {
-			return "", nil
-		}
-		addr, err := getIPThroughDHCP(mgmtNetwork.Interface)
-		if err != nil {
-			msg := fmt.Sprintf("Requesting IP through DHCP failed: %s", err)
-			return msg, nil
-		}
-		logrus.Infof("DHCP test passed. Got IP: %s", addr)
-		return "", nil
-	}
 	askNetworkMethodVConfirm := func(g *gocui.Gui, _ *gocui.View) error {
 		selected, err := askNetworkMethodV.GetData()
 		if err != nil {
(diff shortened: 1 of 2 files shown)


gitlawr (approved):
LGTM

bk201:
I tested by stopping the DHCP server. The installer still allows me to go to the next step.
Screen Recording 2021-09-15 at 12 24 28 PM

johnliu55tw:
Good catch, I'll look into what has gone wrong

johnliu55tw:
@bk201 Fixed the bug. However, I noticed that the installer will only report error **once** if DHCP configuration failed. If users press Enter again, the installer will proceed:
Peek 2021-09-15 18-58

 The main reason seems to be that `wicked ifreload <interface>` won't report error again if the configuration doesn't change (won't request for IP through DHCP again):
image

I think our intent is to request for IP every time users press Enter, regardless of how many fails have occurred. I might have to find a way to tell `wicked` to do just that.

bk201:
> The main reason seems to be that `wicked ifreload <interface>` won't report error again if the configuration doesn't change (won't request for IP through DHCP again):

I did some tests but can't find a reliable way via `wicked` command.

@yaocw2020 did find a nice way to test dhcp requests:
https://github.com/harvester/harvester-installer/blob/4a9762490e3e7fce2fe08962b44d46b36463bc60/pkg/console/vip.go#L75

But he is also replacing it with a dhcp client library: https://github.com/harvester/harvester-installer/pull/120

yaocw2020:
> > The main reason seems to be that `wicked ifreload <interface>` won't report error again if the configuration doesn't change (won't request for IP through DHCP again):
> 
> I did some tests but can't find a reliable way via `wicked` command.
> 
> @yaocw2020 did find a nice way to test dhcp requests:
> https://github.com/harvester/harvester-installer/blob/4a9762490e3e7fce2fe08962b44d46b36463bc60/pkg/console/vip.go#L75
> 
> But he is also replacing it with a dhcp client library: #120

According to my testing,  I think `/usr/lib/wicked/bin/wickedd-dhcp4 --test <ifname>` command is just for testing purpose. The requested IP can not be recorded in the DHCP server.

johnliu55tw:
> According to my testing, I think `/usr/lib/wicked/bin/wickedd-dhcp4 --test <ifname>` command is just for testing purpose. The requested IP can not be recorded in the DHCP server. @johnliu55tw @bk201

I think this is fine in our case, cause we won't store the IP either. Even if the IP is recorded on the DHCP server, there's no guarantee that the IP will remain the same after a reboot cause DHCP server might have been reconfigured or something.

johnliu55tw:
@bk201 @yaocw2020 So I'm **only** calling `getIPThroughDHCP` to validate DHCP, no `wicked` commands are involved here. I think this will work because
- `setupNetwork` will be called eventually,  before heading to the next page. It uses `wicked` to setup networking.
- If `setupNetwork` succeeded, installer would turn to the next page immediately. If users go back to the mgmt network page, we would load up the IP addresses of each interfaces because `setupNetwork` configured them. Users could now see the IP requested from DHCP server.

bk201 (approved):
Thanks. The error messages are also much more clear than before.


Screen Shot 2021-09-17 at 12 32 23 PM
@weihanglo Merging this PR will cause conflicts to bonding works.

PR #124, opened 2021-09-16
Use default route device as the interface of canal

yaocw2020:
It's a temporary solution before the bond solution completes.

issue: https://github.com/harvester/harvester/issues/1207


guangbochen (approved):

guangbochen:
@bk201 @gitlawr can u PTAL of this update, this is a temp solution before the bond PR is ready, at least for now it will not block QA and e2e test with UI and VLAN network, thanks.

bk201 (approved):
LGTM.
We will add the value back when bonding change is ready.

PR #125, opened 2021-09-16
Fix systemd-timesyncd is not enabled

bk201:
The `systemd-timesyncd` service should be enabled to make time sync work.
https://rancher-sandbox.github.io/cos-toolkit-docs/docs/reference/cloud_init/#stagesstage_idstep_nametimesyncd


guangbochen (approved):

FrankYang0529 (approved):
LGTM

PR #127, opened 2021-09-17
Get management URL from rancher-expose svc for agents

yaocw2020:
issue: https://github.com/harvester/harvester/issues/1210


gitlawr (approved):
LGTM

PR #128, opened 2021-09-17
Set manegement bond network during installation

weihanglo:
## What

We can set management bond network during installation.

- ISO: adding two action items in network connection page
    - Management Bond: Former "Management NIC". Now user can selection multiple NIC to create a NIC bonding
    - Bond Mode: Select bond mode
- PXE booting: Since update the harvester config `install.networks` has been updated. User can specify `install.networks.intefaces` to create a bond.

image


weihanglo:
Got "is attempting to grant RBAC permissions not currently held" for following rules:

```yaml
- apiGroups:
    - subresources.kubevirt.io
  resources:
    - virtualmachineinstances/guestosinfo
    - virtualmachineinstances/filesystemlist
    - virtualmachineinstances/userlist
  verbs:
    - get
```

Need to figure out why.

pkg/console/install_panels.go
```diff
@@ -783,7 +814,7 @@ func addNetworkPanel(c *Console) error {
 		spinner := NewFocusSpinner(g, networkValidatorPanel, fmt.Sprintf("Requesting IP through DHCP..."))
 		spinner.Start()
 		go func(g *gocui.Gui) {
-			addr, err := getIPThroughDHCP(mgmtNetwork.Interface)
+			addr, err := getIPThroughDHCP(mgmtNetwork.Interfaces[0].Name)
```
gitlawr, line 817: Why do we get IP for the slave interface?

pkg/console/validator.go
```diff
@@ -39,23 +39,33 @@ func prettyError(errMsg string, value string) error {
...
-			return nil
+		if i.Name == iface.Name {
+			return checkFlags(i.Flags, iface.Name)
 		}
 	}
-	return prettyError(ErrMsgInterfaceNotFound, name)
+
+	if iface.Name == "" {
+		return prettyError(ErrMsgInterfaceNotFound, iface.Name)
+	}
```
gitlawr, line 67: duplicated check with L44

pkg/console/validator.go
```diff
@@ -119,12 +129,21 @@ func checkIPList(ipList []string) error {
 	return nil
 }
 
-func checkNetworks(networks []config.Network) error {
+func checkNetworks(networks map[string]config.Network) error {
+	if len(networks) == 0 {
+		return errors.New(ErrMsgMgmtInterfaceNotSpecified)
+	}
+
+	if _, ok := networks[config.MgmtInterfaceName]; !ok {
```
gitlawr, line 137: Should we check if networks[config.MgmtInterfaceName].Interfaces is empty

pkg/config/cos.go
```diff
@@ -314,7 +340,7 @@ func UpdateWifiConfig(stage *yipSchema.Stage, wifis []Wifi, run bool) error {
 	}
 
 	if run {
-		stage.Commands = append(stage.Commands, fmt.Sprintf("wicked ifreload %s", strings.Join(interfaces, " ")))
+		// stage.Commands = append(stage.Commands, fmt.Sprintf("wicked ifreload %s", strings.Join(interfaces, " ")))
```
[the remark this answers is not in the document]
weihanglo: Oops... Sorry about that.

pkg/console/validator.go
```diff
@@ -39,23 +39,33 @@ func prettyError(errMsg string, value string) error {
...
+		if i.Name == iface.Name {
+			return checkFlags(i.Flags, iface.Name)
 		}
 	}
-	return prettyError(ErrMsgInterfaceNotFound, name)
+
+	if iface.Name == "" {
+		return prettyError(ErrMsgInterfaceNotFound, iface.Name)
+	}
+	return prettyError(ErrMsgInterfaceNotFound, iface.HwAddr)
```
bk201, line 68: I'd suggest also including the interface name here, HwAddr is not implemented yet.

pkg/console/install_panels.go
weihanglo: Because the bond network has not yet been created while selecting DHCP/Static method. I guess we can query for both master and slave interfaces and only report errors when both fail.

bk201: We should get IP via the bonded device.
The MAC address that is used to request IP is determined by bond mode.

pkg/config/templates/wicked-ifcfg-bond-master
```diff
@@ -0,0 +1,19 @@
...
+BOOTPROTO='{{ .Method }}'
+{{ if eq .Method "static" -}}
+IPADDR={{ .IP }}
+NETMASK={{ .SubnetMask }}
+{{- end }}
+{{ range $i, $interface := .Interfaces -}}
+BONDING_SLAVE_{{ $i }}='{{ $interface.Name }}'
+{{ end }}
+{{  with .BondOption -}}
+{{- $bondMode := or .Mode "balance-tlb" -}}
```
bk201, line 12: We have a bit of inconsistency here:
- From installer,`balance-rr` is used by default (I think because it's the first option)
- From PXE boot, if no bond mode id specified, then `balance-tlb` is used.

pkg/console/install_panels.go
weihanglo: I though it was only a DHCP test. If we really want to query DHCP server with bond network name, we should establish that beforehand, and that will reorder the whole network interface creation.

pkg/config/templates/wicked-ifcfg-bond-master
weihanglo: Folks, which should be the default? Do you have any preference for the order of bond mode options?

guangbochen: I think `blancer-rr` is good for general purposes but must be configured from Switch to support static trunking first, however, the `balance-tlb` does not require any Switch configs so let's just default to `balance-tlb` for now.
http://qnapsupport.net/load-balance-setup-which-is-the-best-port-trunking-option/

guangbochen:
@weihanglo can we please prepare PRs for both the docs https://github.com/harvester/docs/tree/v0.3-dev and ipxe-example of this change, thanks

pkg/console/install_panels.go
weihanglo: What if we just remove this DHCP server validation, since the address is not stored anywhere. Is there any risk?

gitlawr: We need to configure the network for next stages like fetching the remote config file. Can we create the bond after selecting dhcp/static mode?

weihanglo: I moved DHCP check after the network `wicked ifreload` on 3b405e0. Thanks.

pkg/console/install_panels.go
```diff
@@ -802,41 +817,12 @@ func addNetworkPanel(c *Console) error {
...
 		if selected == config.NetworkMethodStatic {
 			return showNext(c, dnsServersPanel, gatewayPanel, addressPanel)
 		}
 
 		c.CloseElements(dnsServersPanel, gatewayPanel, addressPanel)
-		if err := networkValidatorV.Show(); err != nil {
-			return err
-		}
-		spinner := NewFocusSpinner(g, networkValidatorPanel, fmt.Sprintf("Requesting IP through DHCP..."))
-		spinner.Start()
```
gitlawr, line 784: It seems like we remove the spinner showing DHCP request status. Is this a UX regression?

weihanglo: The spinner waiting for DHCP request is merging into "Applying Network Configurations". Shall we display the DHCP waiting status separately?

bk201:
If I have two interfaces on my system: `ens5` and `ens6`
- Create a bond with `ens5` and `ens6` and DHCP it.
- Go back to the panel again, but this time only chose `ens5`.
- Configuration file of `ens6` still exist.

This is not introduced in this PR, but we need a way to reset the config files to the initial state.

pkg/console/install_panels.go
gitlawr: I see. That works

yaocw2020:
What about adding the first one NIC as the default slave?
image

weihanglo:
> What about adding the first one NIC as the default slave?
> image

@yaocw2020 Fair! I'll add that. Thanks

weihanglo:
> This is not introduced in this PR, but we need a way to reset the config files to the initial state.

@bk201. On 79769e7,  I've just added a mechanism that will restore the original network config before each time applying network to live OS. Please check it thanks. 😃 

> What about adding the first one NIC as the default slave?

@yaocw2020. I found that it's not convenient for current widget to provide a default value 😞 . If you think this is necessary in RC release, please tell me. I'll do a workaround. 🙇🏾‍♂️

pkg/console/install_panels.go
```diff
@@ -653,8 +663,23 @@ func addNetworkPanel(c *Console) error {
...
 		logrus.Infof("Network configuration is applied: %s", output)
 
-		c.config.Networks = []config.Network{
-			mgmtNetwork,
+		c.config.Networks = map[string]config.Network{
+			config.MgmtInterfaceName: mgmtNetwork,
+		}
+
+		if mgmtNetwork.Method == config.NetworkMethodDHCP {
+			if addr, err := getIPThroughDHCP(config.MgmtInterfaceName); err != nil {
```
yaocw2020, line 671: If we have reloaded the wicked network configurations, it may not required to get the IP through DHCP here.

yaocw2020:
image

Is it convenient to add the `miimon` option for the bond device?

guangbochen:
> image
> 
> Is it convenient to add the `miimon` option for the bond device?

considering doing it in the new console UI.

pkg/console/install_panels.go
bk201: We better check again because if configuration is not changed, reload doesn't do anything.

gitlawr (approved):
LGTM. The default value is nice but I think we can get it ready later

janeczku:
Does this also cover bond for VLAN network or is that a separate implementation?  https://github.com/harvester/harvester/issues/1048

guangbochen:
> Does this also cover bond for VLAN network or is that a separate implementation? [harvester/harvester#1048](https://github.com/harvester/harvester/issues/1048)

this is for the management bond, the VLAN bond will be separated, and probably will need to do it in 1.0.0 version

guangbochen (approved):

bk201 (approved):
LGTM. Thanks for fixing all minor issues.

PR #129, opened 2021-09-19
fix(install_panels.go): change default ntp server to SUSE

FrankYang0529:
issue https://github.com/harvester/harvester/issues/1248

### DIFF
--- a/pkg/console/install_panels.go
+++ b/pkg/console/install_panels.go
@@ -41,7 +41,7 @@ const (
 var (
 	once          sync.Once
 	userInputData = UserInputData{
-		NTPServers: "ntp.ubuntu.com",
+		NTPServers: "0.suse.pool.ntp.org",
 	}
 	mgmtNetwork = config.Network{}
 )


bk201 (approved):

pkg/console/install_panels.go
```diff
@@ -41,7 +41,7 @@ const (
 var (
 	once          sync.Once
 	userInputData = UserInputData{
-		NTPServers: "ntp.ubuntu.com",
+		NTPServers: "0.suse.pool.ntp.org",
```
guangbochen, line 44: is there any reference of why we choose the `0.suse.pool.ntp.org`, thanks, it seems like subdomain to me

FrankYang0529: Yes, I saw this domain when I install openSUSE. Also, it is default `FallbackNTPServers` in harvester-v0.3.0-preview1.

Screen Shot 2021-09-23 at 12 47 31 PM


Screen Shot 2021-09-23 at 12 37 44 PM

guangbochen: I think we should use the pool ntp server config instead of a single server node, can we check it this works.
https://documentation.suse.com/sles/15-SP1/html/SLES-all/cha-ntp.html#sec-netz-xntp-netconf

FrankYang0529: Do you mean use `pool.ntp.org` as default input? I have checked `suse.pool.ntp.org`. We don't have this domain. We can also use `0.suse.pool.ntp.org,1.suse.pool.ntp.org,2.suse.pool.ntp.org` as default input.

guangbochen: Yep, does the `pool.ntp.org` works? if not probably default to the above multi ntp server config like 0,1,2 is much better, thanks

gitlawr: I think 0.suse.pool.ntp.org is fine. It resolves to a set of timeservers

FrankYang0529: @guangbochen yes, `pool.ntp.org` works. Also, both `pool.ntp.org` and `0.suse.pool.ntp.org` can be resolved to a set of timeservers. Do we need to change default input to `pool.ntp.org`? Thank you.

guangbochen: if it will be resolved to many timeservers then let's just use the SUSE one for now, thanks

guangbochen (approved):

PR #130, opened 2021-09-22
feat(install_panels): enable ntp server after settings

FrankYang0529:
issue: https://github.com/harvester/harvester/issues/935

Enable NTP time sync if NTP servers can be connected.

### DIFF
--- a/pkg/console/util.go
+++ b/pkg/console/util.go
@@ -147,8 +147,9 @@ func enableNTPServers(ntpServerList []string) error {
 		return err
 	}
 
-	_, err = exec.Command("timedatectl", "set-ntp", "true").Output()
+	output, err := exec.Command("timedatectl", "set-ntp", "true").CombinedOutput()
 	if err != nil {
+		logrus.Error(err, string(output))
 		return err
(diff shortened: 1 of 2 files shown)


pkg/console/util.go
```diff
@@ -130,6 +130,31 @@ func validateNTPServers(ntpServerList []string) error {
...
+
+	ntpServersString := fmt.Sprintf("NTP=%s\n", strings.Join(ntpServerList, " "))
+	_, err = file.WriteString(ntpServersString)
+	if err != nil {
+		return err
+	}
+
+	_, err = exec.Command("timedatectl", "set-ntp", "true").Output()
+	if err != nil {
+		return err
```
gitlawr, line 152: Use CombinedOutput() and log the output when we hit an error. The returned error is usually an ExitError that may not have much info for debugging

FrankYang0529: Updated. Thank you.

gitlawr (approved):
LGTM

guangbochen (approved):
LGTM

PR #132, opened 2021-09-24
Switch Live CD bootloader to Grub2

bk201:
Fix `net.ifnames=1` parameter is not set in UEFI boot.
By switching to Grub2, we make sure there is only one menuentry.
Grub2 is also suggested by the cOS team.

Live CD screenshot:

Screen Shot 2021-09-24 at 6 17 10 PM

**Related issue**:
- https://github.com/harvester/harvester/issues/1267
- https://github.com/harvester/harvester/issues/1200


bk201:
I test the change by using BalenaEtcher to create a USB installation disk.
The `net.ifnames=1` parameter is correctly set. Both UEFI and Legacy boot work on my laptop.

package/harvester-os/Dockerfile
```diff
@@ -10,4 +10,4 @@ RUN rm -f /system/oem/05_network.yaml
 
 ARG HARVESTER_INSTALLER_VERSION
 
-RUN echo GRUB_ENTRY_NAME="\"Harvester $HARVESTER_INSTALLER_VERSION\"" >> /etc/os-release
+RUN echo GRUB_ENTRY_NAME="\"HarvesterOS $HARVESTER_INSTALLER_VERSION\"" >> /etc/os-release
```
guangbochen, line 13: `HarvesterOS` actually does not exist, we can leave it as Harvester thanks.

bk201: Removed the term.

package/harvester-os/iso/boot/grub2/grub.cfg
```diff
@@ -0,0 +1,35 @@
...
+    fi
+fi
+if [ "${grub_platform}" = "efi" ]; then
+    echo "Please press 't' to show the boot menu on this console"
+fi
+set font=($root)/boot/x86_64/loader/grub2/fonts/unicode.pf2
+if [ -f ${font} ];then
+    loadfont ${font}
+fi
+menuentry "Harvester Live ${harvester_version}" --class os --unrestricted {
```
gitlawr, line 23: Should we print “Harvester Installer” given that it’s not a live OS for users

guangbochen (approved):
LGTM

package/harvester-os/iso/boot/grub2/grub.cfg
bk201: Make sense. Updated.

gitlawr (approved):
LGTM

PR #133, opened 2021-09-24
Remove duplicate entries in configuration when users reset NTP servers

FrankYang0529:
* Upgrade github.com/tredoe/osutil to v1.3.6, because we can't find v1.1.8 in [release tags](https://github.com/tredoe/osutil/releases).
* Use go-ini to read and write [timesyncd configuration](https://www.freedesktop.org/software/systemd/man/systemd.network.html#) file because it's ini-style file.
* Remove duplicate entries when users reset NTP servers.

issue: https://github.com/harvester/harvester/issues/935


bk201 (approved):
LGTM

gitlawr (approved):
LGTM

PR #134, opened 2021-09-24
Show harvester version on console dashboard

johnliu55tw:
* Add a new version variable `HarvesterVersion` to store version of harvester. Console dashboard shows this version, instead of the installer version.

* Add a new script `version-harvester` to determine and set up `HARVESTER_VERSION` variable for later use.

* Update script `version`: Unset variable `DIRTY` in case it was set somewhere else.

* Update script `build`:
  - Pull harvester source from github if not exists. This will happen if we're calling `make` from harvester-installer, but not from calling `make build-iso` from harvester.

  - Source "version-harvester" script to get the version of harvester.

  - Update both harvester and harvester-installer version variables using ldflags, `HarvesterVersion` and `Version` respectively.

Tested by calling `make` from harvester-installer and `make build-iso` from harvester.

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -1,5 +1,6 @@
 package version
 
 var (
-	Version = "dev"
+	Version          = "dev"
+	HarvesterVersion = "dev" // Will be replaced by ldflags
 )
--- a/scripts/build
+++ b/scripts/build
@@ -5,13 +5,28 @@ TOP_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." &> /dev/null && pwd )"
 SCRIPTS_DIR="${TOP_DIR}/scripts"
 PACKAGE_HARVESTER_OS_DIR="${TOP_DIR}/package/harvester-os"
 
-source ${SCRIPTS_DIR}/version-rke2
-
 cd ${TOP_DIR}
+
+# Pull harvester source and determine harvester version
+harvester_path=../harvester
+if [ ! -d ${harvester_path} ];then
+    echo "No existed harvester source. Pulling..."
+    git clone --branch master --single-branch --depth 1 https://github.com/harvester/harvester.git ../harvester
+fi
+
+source ${SCRIPTS_DIR}/version-harvester $harvester_path
+source ${SCRIPTS_DIR}/version-rke2
+source ${SCRIPTS_DIR}/version
+
+echo "Harvester version: ${HARVESTER_VERSION}"
+echo "Installer version: ${VERSION}"
 
 mkdir -p bin
 
-LINKFLAGS="-X github.com/harvester/harvester-installer/pkg/config.RKE2Version=$RKE2_VERSION $LINKFLAGS"
+LINKFLAGS="-X github.com/harvester/harvester-installer/pkg/config.RKE2Version=$RKE2_VERSION
+           -X github.com/harvester/harvester-installer/pkg/version.Version=$VERSION
+           -X github.com/harvester/harvester-installer/pkg/version.HarvesterVersion=$HARVESTER_VERSION
+           $LINKFLAGS"
 
 if [ "$(uname)" = "Linux" ]; then
     OTHER_LINKFLAGS="-extldflags -static -s"
--- a/scripts/version-harvester
+++ b/scripts/version-harvester
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+HARVESTER_VERSION=$(cd $1; source ./scripts/version &> /dev/null; echo $VERSION)
--- a/pkg/console/dashboard_panels.go
+++ b/pkg/console/dashboard_panels.go
@@ -98,7 +98,7 @@ func logoPanel(g *gocui.Gui) error {
 		}
 		v.Frame = false
 		fmt.Fprintf(v, logo)
-		versionStr := "version: " + version.Version
+		versionStr := "version: " + version.HarvesterVersion
 		logoLength := 74
 		nSpace := logoLength - len(versionStr)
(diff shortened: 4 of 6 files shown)


scripts/build
```diff
@@ -5,13 +5,26 @@ TOP_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." &> /dev/null && pwd )"
...
 cd ${TOP_DIR}
 
+# Pull harvester source and determine harvester version
+harvester_path=../harvester
+if [ ! -d ${harvester_path} ];then
+    echo "No existed harvester source. Pulling..."
+    git clone --branch master --single-branch --depth 1 https://github.com/harvester/harvester.git ../harvester
+fi
+HARVESTER_VERSION=$(${SCRIPTS_DIR}/version-harvester $harvester_path)
+echo "Harvester version: $HARVESTER_VERSION"
```
gitlawr, line 17: Can we move the command substitution to version-harvester so that we can do `source ${SCRIPTS_DIR}/version-harvester`

pkg/version/version.go
```diff
@@ -1,5 +1,6 @@
 package version
 
 var (
-	Version = "dev"
+	Version          = "dev"
+	HarvesterVersion = "none" // Will be replaced by ldflags
```
gitlawr, line 5: Let's use dev as the default as well

guangbochen:
btw @johnliu55tw can u fix the description typo, thanks

scripts/build
johnliu55tw: `version-harvester` needs to source harvester's `version` script to get version information, which would also introduce other variables like `GIT_TAG` and `COMMIT` into the current scope. I'm a bit worried about this because the **harvester-installer**'s `version` script also defines variables with identical names, and this may cause confusion when we `source` both scripts into the current context. By executing `version-harvester` instead of `source`, only the variable we want i.e. the `VERSION` variable would be "returned" and we could assign it to a different variable.

Nevertheless, I did it this way to avoid changing too many things altogether in harvester-installer and harvester repositories. If we prefer the `source` way, I could totally make a change ;)

pkg/version/version.go
johnliu55tw: Done

johnliu55tw:
> btw @johnliu55tw can u fix the description typo, thanks

The PR description? Could you help point out the typo cause I can't tell where it is... Thanks!

guangbochen:
hmm that's how it looks like from my side 
image

scripts/build
gitlawr: Example:
```
# version-harvester
HARVESTER_VERSION=$(source $harvester_path/scripts/version;echo $VERSION)

# scripts/build
source ${SCRIPTS_DIR}/version-harvester
```
The idea is to make the `version-harvester` script self-contained so that the caller can simply source it. There's no much difference from current implementation.

johnliu55tw:
@gitlawr fixed as you suggested. I do notice a minor side-effect: we need to source `version-harvester` before `version` so that the `DIRTY` variable won't be passed to harvester's `version` script. I tweaked the `version` script in **harvester-installer** to unset the variable at the beginning, and I think we should do the same thing to the `version` script of **harvester** too, just in case.

scripts/version-harvester
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+HARVESTER_VERSION=$(cd $1; source ./scripts/version &> /dev/null; echo $VERSION)
```
gitlawr, line 3: Let's try to avoid polluting the envvars. Does it help if we do `sh` instead of `source` here?

johnliu55tw: Hmmm but because `scripts/version` does not return (echo) anything, I think the parent shell can't get anything from subshell if we replace `source` with `sh`? In other words, the `echo $VERSION` won't work.

gitlawr: yep, that doesn't work.
From my testing, envs in the command substitution will not affect the caller context, is it neccessary to `unset DIRTY` in scripts/version?

```
VERSION=v-old
DIRTY=
cat <<EOF > a.sh
DIRTY=1
VERSION=v1
EOF
HARVESTER_VERSION=$(source a.sh &> /dev/null; echo $VERSION)
echo "VERSION: $VERSION"
echo "DIRTY: $DIRTY"
echo "HARVESTER_VERSION: $HARVESTER_VERSION"

# result
VERSION: v-old
DIRTY:
HARVESTER_VERSION: v1
```

johnliu55tw: If `DIRTY` was set in parent shell (which is true in our case because we sourced a version script), subshell will receive the variable:
```bash
VERSION=v-old
DIRTY="itsdirty"  # Set by sourcing a version script in harvester-installer
cat <<EOF > a.sh
VERSION="v1.2.3-${DIRTY}"
EOF
HARVESTER_VERSION=$(source a.sh &> /dev/null; echo $VERSION)
echo "HARVESTER_VERSION: $HARVESTER_VERSION"

# result
HARVESTER_VERSION: v1.2.3-itsdirty
```

In the `version` script, we append `$DIRTY` directly:
```bash
if [[ -z "$DIRTY" && -n "$GIT_TAG" ]]; then
    VERSION=$GIT_TAG
else
    VERSION="${COMMIT}${DIRTY}"
fi
```

johnliu55tw: Oh BTW with or without `DIRTY` in this PR does not affect the behavior. It's only problematic when we source those version scripts in specific order.

gitlawr (approved):
LGTM

bk201 (approved):

PR #136, opened 2021-09-26
Label nodes which supports virtualization

olljanat:
Extend logic which was added on #93 in way that there will additional `harvesterhci.io/kubevirt=true` label on those nodes which can run KubeVirt and host virtual machines.

First step to support VM mode requested on https://github.com/harvester/harvester/issues/1290


olljanat:
Actually most probably this is not needed. Will close and reopen if needed.

PR #138, opened 2021-09-27
fix(util): prolong ntp timeout deadline

FrankYang0529:
issue: https://github.com/harvester/harvester/issues/1248

### DIFF
--- a/pkg/console/util.go
+++ b/pkg/console/util.go
@@ -101,7 +101,7 @@ func validateNTPServers(ntpServerList []string) error {
 		if err != nil {
 			return err
 		}
-		if err := conn.SetDeadline(time.Now().Add(5 * time.Second)); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/console/util.go
```diff
@@ -101,7 +101,7 @@ func validateNTPServers(ntpServerList []string) error {
 		if err != nil {
 			return err
 		}
-		if err := conn.SetDeadline(time.Now().Add(5 * time.Second)); err != nil {
+		if err := conn.SetDeadline(time.Now().Add(15 * time.Second)); err != nil {
```
guangbochen, line 104: I reckon 3 retries with 5 seconds timeout should be sufficient enough, otherwise, it takes too long for the NTP config from the user perspective.

FrankYang0529: Updated it.

guangbochen (approved):
LGTM

bk201 (approved):
LGTM.

PR #139, opened 2021-09-27
Display the NICs' status during ISO installation

yaocw2020:
**Related issue**: https://github.com/harvester/harvester/issues/1280


guangbochen:
@yaocw2020 currently the bond IP is not displayed anywhere after success configuration, wondering if we can put it after the selected NICs with bracket.
image

Jellyfrog:
Please consider including https://github.com/harvester/harvester/issues/1328 here

yaocw2020:
Dependency pull request: https://github.com/harvester/os2/pull/11

weihanglo (commented):
Overall the PR looks good! There are some display issue need to address.

- ✅ Display NIC state
- ✅ Display vendor and product of NICs
- ❌ Display long description in narrow tty columns (maybe no wrap on text?)
- ❌ Do not block the installer (seems `getNetworkInterfaceOptions` now blocks for a while)

image

yasker:
We should also show the mac address of the nic as well, since macs are commonly used to identify the nics.

guangbochen:
We can resolve the mac address and long text enhancement in the next console UI refactoring, @yaocw2020 can u please help to address the following case only first, thanks
[ ] Do not block the installer (seems getNetworkInterfaceOptions now blocks for a while)

weihanglo:
I guess there is no express path to resolve the long description issue at this moment. `Wrap` not works because wrapping only happens inside a `gocui.View` level but the all the options of select widget is a single `gocui.View`.

weihanglo (approved):
LGTM. Thanks for the update.

image

PR #141, opened 2021-09-29
Move DNS setting to optional panel

FrankYang0529:
Remove `DNSNameservers` in `Network`.

issue: https://github.com/harvester/harvester/issues/1296

### DIFF
--- a/pkg/config/cos.go
+++ b/pkg/config/cos.go
@@ -10,7 +10,6 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/harvester/harvester-installer/pkg/util"
 	yipSchema "github.com/mudler/yip/pkg/schema"
 	"github.com/sirupsen/logrus"
 )
@@ -284,8 +283,6 @@ func SaveOriginalNetworkConfig() error {
 // - manipulates nameservers in `/etc/resolv.conf`.
 // - call `wicked ifreload all` if `run` flag is true.
 func UpdateNetworkConfig(stage *yipSchema.Stage, networks map[string]Network, run bool) error {
-	var staticDNSServers []string
-
 	mgmtNetwork, ok := networks[MgmtInterfaceName]
 	if !ok {
 		return errors.New("no management network defined")
@@ -340,19 +337,6 @@ func UpdateNetworkConfig(stage *yipSchema.Stage, networks map[string]Network, ru
 		case NetworkMethodDHCP, NetworkMethodNone:
 			stage.Commands = append(stage.Commands, fmt.Sprintf("rm -f /etc/sysconfig/network/ifroute-%s", name))
 		}
-
-		for _, nameServer := range network.DNSNameservers {
-			if util.StringSliceContains(staticDNSServers, nameServer) {
-				continue
-			}
-			staticDNSServers = append(staticDNSServers, nameServer)
-		}
-	}
-
-	// Set static DNS servers before wicked reload
-	if len(staticDNSServers) > 0 {
-		// Not using stage.Environment because it's run after stage.Commands in Yip
-		stage.Commands = append(stage.Commands, getAddStaticDNSServersCmd(staticDNSServers))
 	}
 
 	if run {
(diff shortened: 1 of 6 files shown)


pkg/config/cos.go
```diff
@@ -284,8 +283,6 @@ func SaveOriginalNetworkConfig() error {
 // - manipulates nameservers in `/etc/resolv.conf`.
 // - call `wicked ifreload all` if `run` flag is true.
 func UpdateNetworkConfig(stage *yipSchema.Stage, networks map[string]Network, run bool) error {
-	var staticDNSServers []string
```
weihanglo, line 287: Just curious, do we intend to remove the ability to configure DNS server for LIVE OS?

weihanglo (commented):
Build the ISO and confirmed that the dnsserver field in each network config is gone.

pkg/config/cos.go
bk201: We need to make the DNS servers text field out of networks and save them to `OS.DNSNameservers`.

FrankYang0529: @bk201 do we need to ask users to input DNS servers if they choose DHCP mode?

bk201: Those are entered as static DNS servers, it also works with DHCP mode.

FrankYang0529: @bk201 , we still can let users configure DNS servers in static mode. After confirming with @weihanglo , I think what he means is that "whether we let users have ability to connect to internet/intranet after they configure network? Instead of waiting util Install Panel". WDYT?

bk201: We need DNS settings to be applied after the network pannel. NTP panel, ssh keys panel, and config panels need this.
Do you think moving the `DNS servers` Input widget out of static IP settings works?
Screen Shot 2021-09-30 at 10 45 10 AM
l?

guangbochen: Yes, that would make sense for some users and make it an optional config

FrankYang0529: Okay. I will add an optional DNS panel if users choose static mode.

bk201: > Okay. I will add an optional DNS panel if users choose static mode.

The optional DNS servers work for both DHCP and static IP modes.

FrankYang0529: Added DNS optional panel. @bk201 @guangbochen @weihanglo please help me review again. Thank you.

pkg/console/install_panels.go
```diff
@@ -40,6 +40,7 @@ var (
 	once          sync.Once
 	userInputData = UserInputData{
 		NTPServers: "0.suse.pool.ntp.org",
+		DNSServers: "1.1.1.1",
```
weihanglo, line 43: 1.1.1.1 is fine to me. Raise it up here for others to double check this is what we want by default.

pkg/console/util.go
```diff
@@ -160,6 +160,23 @@ func enableNTPServers(ntpServerList []string) error {
...
 
+func updateDNSServersAndReloadNetConfig(dnsServerList []string) error {
+	dnsServers := strings.Join(dnsServerList, " ")
+	output, err := exec.Command("sed", "-i", fmt.Sprintf(`s/^NETCONFIG_DNS_STATIC_SERVERS.*/NETCONFIG_DNS_STATIC_SERVERS="%s"/`, dnsServers), "/etc/sysconfig/network/config").CombinedOutput()
+	if err != nil {
+		logrus.Error(err, string(output))
+		return err
+	}
+
+	output, err = exec.Command("netconfig", "update").CombinedOutput()
```
weihanglo, line 171: NIT: would it be better running `netconfig update -m dns` that only updates DNS settings?

weihanglo (approved):
Overall looks great! I like the cleanup of the network page and the config reload.

pkg/console/install_panels.go
bk201: I'd prefer to have no default DNS servers in this field. Users should manually enter servers if they need it.
We don't host DNS servers like `1.1.1.1` or `8.8.8.8`, hard-coding them in the system is misuse.

FrankYang0529: Okay. I remove it. Thank you.

bk201 (approved):
LGTM

PR #142, opened 2021-09-30
Add a new network method none to bypass network configuration

weihanglo:
Related: https://github.com/harvester/harvester/issues/1325

## Solution

- Add a new network method `none` to bypass configuring any DHCP or static network

### Installer console

- [ ] Run installer and it should act as before. Won't show the `none` network method.

### Install via pxe boot

- [ ] Set `install.networks.harvester-mgmt.method` to `none`. Should throw an error.
- [ ] Set `install.networks.harvester-mgmt.method` to `static` or `dhcp`. Should work as before.
- [ ] Set `install.networks.harvester-mgmt.method` to `static` or `dhcp`. And configure another network with `method: none` Should succeed to install and won't generate any ifroute-* files and only ifcfg- with `BOOTPROTO=none` for the extra network.

### DIFF
--- a/pkg/config/constants.go
+++ b/pkg/config/constants.go
@@ -7,6 +7,7 @@ const (
 
 	NetworkMethodDHCP   = "dhcp"
 	NetworkMethodStatic = "static"
+	NetworkMethodNone   = "none"
 
 	MgmtInterfaceName = "harvester-mgmt"
 
--- a/pkg/config/cos.go
+++ b/pkg/config/cos.go
@@ -311,7 +311,9 @@ func UpdateNetworkConfig(stage *yipSchema.Stage, networks map[string]Network, ru
 	}
 
 	for name, network := range networks {
-		if network.Method != NetworkMethodDHCP && network.Method != NetworkMethodStatic {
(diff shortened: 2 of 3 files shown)


bk201 (approved):
LGTM. Test with the following procedure:

- PXE boot with an additional network config:
```
  networks:
    harvester-mgmt:
      interfaces:
      - name: ens5
      method: dhcp
    harvester-vlan:
      interfaces:
      - name: ens6
      method: none
```
- After installation. Go to GUI and enable VLAN feature on interface `harvester-vlan`.
- Create a VLAN network on vlan 5.
- Spawn a VM and use the new-created network as the second network.
- Identify the VM host and capture packets on interface `harvester-vlan`:
  ```
  tcpdump -i harvester-vlan -nn -e vlan
  ```
- Connect to VM and run DHCP on the second interface.
  ```
  dhclient enp2s0
  ```
- Confirm VLAN-tagged (5) DHCP packets are received.

bk201:
@weihanglo could you help to update the corresponding doc (The harvester configuration)? Thanks.

PR #143, opened 2021-10-01
Allow installation on raid-disks

Jellyfrog:
Allow installation on raid-disks such as mdraid / Intel RST

Before:
Screenshot from 2021-10-04 14-33-19
After:
Screenshot from 2021-10-04 15-06-59


Jellyfrog:
Works to install, but not to boot due to initrd not having `intel-rst` kernel module.
Don't really know where to add that, maybe in cos-toolkit?

PR #147, opened 2021-10-05
feat(install_panels): add bond network note

FrankYang0529:
issue: https://github.com/harvester/harvester/issues/1327


pkg/console/constant.go
```diff
@@ -56,6 +57,7 @@ const (
 	proxyNote              = "Note: In the form of \"http://[[user][:pass]@]host[:port]/\"."
 	sshKeyNote             = "For example: https://github.com/<username>.keys"
 	ntpServersNote         = "Note: You can use comma to add more NTP servers."
+	bondNote               = "Note: Bond Mode can be ignored for a single NIC."
```
bk201, line 60: I guess bond mode can't be ignored even for a single NIC. e.g., using `802.3ad` mode without configuring switches might cause problems.

weihanglo (commented):
It's great to see this be implemented without additional rows added 😆 
Like what bk201 said. The wording need some efforts.

pkg/console/constant.go
weihanglo: Or even better, if there is only one NIC existing, do not display bond mode option. Unfortunately, supposed the release date is coming, it is hard to tweak current console UI in a timely manner. 
Maybe change it to something like "Note: Use the default Bond Mode for a single NIC machine" ?

FrankYang0529: @weihanglo, thanks for the suggestion. Sound good to me!

@bk201, thanks for the correction. I didn't know that. What do you think about Weihang's suggestion?

bk201: @yasker Any suggestion to the wording here? Thanks
I guess users need two information:
- Context about why a bond device is always created.
- A preferred bond mode to use when there is just a single NIC for management network

FrankYang0529: @yasker this is what it looks like now. We use the same position with the network validator.

Kapture 2021-10-07 at 10 30 51

yasker: `Management Bond` field will still be confusing to users since they won't know why they have to use one nic for the bond. Use `Management Interface` instead of `Management Bond`. 

I see the note is per page, so let's do
> Note:
> Select one or more NICs for the management interface. 
> Use the default value for the Bond Mode if only one NIC is selected.

FrankYang0529: `Management Interface` is too long for the current layout. I try to update the input form. It looks like the following. WDYT? @yasker @bk201 @weihanglo 

Screen Shot 2021-10-07 at 12 24 53 PM

yasker: Let's do `Management NICs` then.

FrankYang0529: Updated it. Thank you.

PR #149, opened 2021-10-06
Enable crash kernel for target OS

bk201:
- Calculate crash kernel memory parameters at post-installation time.
- Generate a kdump initrd file before kdump service starts.
  mkdumprd and dracut kdump module don't work well with current file system layout.
  Customize the fstab file in the result initrd by ourself.

Required OS change: https://github.com/harvester/os2/pull/17
Related issue: https://github.com/harvester/harvester/issues/1335

** How to test **
Build a new ISO with the change and boot into the OS
- [ ] `cat /proc/cmdline`: `crashkernel=244M,high crashkernel=72M,low` parameters should be included (values varies from machine to machine)
- [ ] `echo c > /proc/sysrq-trigger` to simulate a crash. After rebooting, a crash dump should be store in`/var/crash` directory.

### DIFF
--- a/package/harvester-os/files/usr/sbin/gen-kdump-initrd
+++ b/package/harvester-os/files/usr/sbin/gen-kdump-initrd
@@ -1,0 +1,52 @@
+#!/bin/bash -ex
+
+KERNEL_VERSION=$(uname -r)
+INITRD_KDUMP_NAME="initrd-$KERNEL_VERSION-kdump"
+INITRD_CACHE_DIR="/usr/local/cache/kdump"
+
+if [ -e /boot/$INITRD_KDUMP_NAME ]; then
+  echo /boot/$INITRD_KDUMP_NAME exists. Skip generating a new one.
+  exit 0
+fi
+
+if [ -e $INITRD_CACHE_DIR/$INITRD_KDUMP_NAME ]; then
+  echo "Using cache: $INITRD_CACHE_DIR/$INITRD_KDUMP_NAME"
+  cp $INITRD_CACHE_DIR/$INITRD_KDUMP_NAME /boot/$INITRD_KDUMP_NAME
+  exit 0
+fi
+
+mkdir -p $INITRD_CACHE_DIR
+
+TMP_DIR=$(mktemp -d)
+trap "rm -rf $TMP_DIR" exit
+
+TMP_INITRD=$TMP_DIR/initrd
+INITRD_EXTRACTED_DIR=$TMP_DIR/extracted
+DRACUT_CONFDIR=$TMP_DIR/dracut.conf.d
+
+mkdir -p $INITRD_EXTRACTED_DIR
+mkdir -p $DRACUT_CONFDIR
+
+
+# specify a new dracut config dir to avoid including cOS configs in /etc/dracut.conf.d
+cat > $DRACUT_CONFDIR/99-tmp.conf <<EOF
+early_microcode="no"
+host_only="yes"
+EOF
+/usr/bin/dracut --confdir $DRACUT_CONFDIR --force --omit "plymouth resume usrmount" --add kdump $TMP_INITRD $KERNEL_VERSION
+
+# extract initrd and replace fstab and kdump config
+cd $INITRD_EXTRACTED_DIR
+xzcat $TMP_INITRD | cpio -idm
+
+PERSISTENT_PARTITION=$(lsblk -n -o UUID /dev/disk/by-label/COS_PERSISTENT)
+cat > etc/fstab <<EOF
+UUID=$PERSISTENT_PARTITION /kdump/mnt0 auto defaults 0 0
+EOF
+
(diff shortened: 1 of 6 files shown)


package/harvester-os/files/usr/sbin/gen-kdump-initrd
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash -ex
+
+KERNEL_VERSION=$(uname -r)
+INITRD_KDUMP_NAME="initrd-$KERNEL_VERSION-kdump"
+INITRD_CACHE_DIR="/usr/local/cache/kdump"
```
weihanglo, line 5: Just wonder. Would it needs a upgradable path for dump kernel?

package/harvester-os/files/usr/sbin/gen-kdump-initrd
```diff
@@ -0,0 +1,52 @@
...
+mkdir -p $INITRD_EXTRACTED_DIR
+mkdir -p $DRACUT_CONFDIR
+
+
+# specify a new dracut config dir to avoid including cOS configs in /etc/dracut.conf.d
+cat > $DRACUT_CONFDIR/99-tmp.conf <<EOF
+early_microcode="no"
+host_only="yes"
+EOF
+/usr/bin/dracut --confdir $DRACUT_CONFDIR --force --omit "plymouth resume usrmount" --add kdump $TMP_INITRD $KERNEL_VERSION
```
weihanglo, line 36: Not a big deal. Just curious about little stuff:

- How were the omitted modules be determine?
- Could we pass `--hostonly` and `--no-early-microcode` instead of creating a conf file?

package/harvester-os/files/usr/sbin/gen-kdump-initrd
```diff
@@ -0,0 +1,52 @@
...
+
+# extract initrd and replace fstab and kdump config
+cd $INITRD_EXTRACTED_DIR
+xzcat $TMP_INITRD | cpio -idm
+
+PERSISTENT_PARTITION=$(lsblk -n -o UUID /dev/disk/by-label/COS_PERSISTENT)
+cat > etc/fstab <<EOF
+UUID=$PERSISTENT_PARTITION /kdump/mnt0 auto defaults 0 0
+EOF
```
weihanglo, line 45: NIT: maybe this works?

```suggestion
cat > etc/fstab <<EOF
/dev/disk/by-label/COS_PERSISTENT /kdump/mnt0 auto defaults 0 0
EOF
```

Just not sure about the usage of this mount and the modification of `KDUMP_SAVEDIR`. Could you explain them more? Thanks.

package/harvester-os/files/usr/sbin/gen-kdump-initrd
```diff
@@ -0,0 +1,52 @@
...
+
+sed -i 's,KDUMP_SAVEDIR.*,KDUMP_SAVEDIR="file:///mnt0/.state/var-crash.bind",' etc/sysconfig/kdump
+
+# pack initrd again
+find . -print0| sort -z | cpio --null -H newc --quiet -o > $TMP_DIR/initramfs.img 
+cat $TMP_DIR/initramfs.img | xz -0 --check=crc32 --memlimit-compress=50% >> $INITRD_CACHE_DIR/$INITRD_KDUMP_NAME
+cp $INITRD_CACHE_DIR/$INITRD_KDUMP_NAME /boot/$INITRD_KDUMP_NAME
```
weihanglo, line 52: I am totally novice to kdump. What the logic behind the  dump capture kernel finding the right initramfs we provided?

weihanglo (commented):
Overall looks well! Just some random questions from a kdump newbie.

package/harvester-os/files/usr/sbin/gen-kdump-initrd
bk201: They are the same thing, but since I need to override the confdir, I put those in the config files.
It also shortens the command and makes adding new params easier in the future.

package/harvester-os/files/usr/sbin/gen-kdump-initrd
bk201: That's my original version.
But I found in some case user have multiple partitions with the `COS_PERSISTENT` label (e.g. another disk was installed before).
I need to make sure the dump is saving to the current OS's persistent partition.

bk201:
@johnliu55tw @weihanglo 
I create a grub entry to enable the crash kernel.  Users can select the (debug) one to enable it.
Screen Shot 2021-10-08 at 2 55 14 PM

weihanglo (approved):
LGTM. Confirmed the followings:

- [x] There is a new debug boot grub option
- [x] no `crashkernel` parameter for a normal boot 
- [x] kdump.service is disabled for a normal boot
- [x] trigger a crash then it will reboot in 10 seconds without any dump in `/var/crash` for a normal boot
- [x] Got `crashkernel` parameter for a debug boot
- [x] kdump.service is active for a debug boot
- [x] trigger a crash then the dump can be seen in `/var/crash` for a debug boot

PR #151, opened 2021-10-07
Use map to store bond options

johnliu55tw:
## Todos
- [x] Test on manual installation
- [x] Test on PXE instalaltion
- [x] Update documents: https://github.com/harvester/docs/pull/28

## Notes
While installing via ISO and a remote config is supplied, the bond options `mode` and `miimon` will not be overwritten by the remote config due to the merging policy we have right now. The remote config can only be used to supply additional options. I think this is the default behavior.

Related issue: https://github.com/harvester/harvester/issues/1324

### DIFF
--- a/pkg/console/install_panels.go
+++ b/pkg/console/install_panels.go
@@ -805,14 +805,17 @@ func addNetworkPanel(c *Console) error {
 
 	// askBondModeV
 	askBondModeV.PreShow = func() error {
-		if mgmtNetwork.BondOption.Mode == "" {
+		if mgmtNetwork.BondOptions == nil {
 			askBondModeV.Value = config.BondModeBalanceTLB
 		}
 		return nil
 	}
 	askBondModeVConfirm := func(g *gocui.Gui, v *gocui.View) error {
 		mode, err := askBondModeV.GetData()
-		mgmtNetwork.BondOption.Mode = mode
+		mgmtNetwork.BondOptions = map[string]string{
+			"mode":   mode,
+			"miimon": "100",
+		}
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


weihanglo:
You may want to update https://github.com/harvester/ipxe-examples/ as well.

weihanglo:
Cool. Then no need to update 👍🏾

pkg/console/install_panels.go
```diff
@@ -1263,6 +1265,17 @@ func addInstallPanel(c *Console) error {
...
 			}
+
+			// Adding default NIC bonding options for every network config if missing
+			for netName, network := range c.config.Install.Networks {
+				if network.BondOptions == nil {
+					msg := fmt.Sprintf("Adding default NIC bonding options for \"%s\"", netName)
+					logrus.Info(msg)
+					printToPanel(c.Gui, msg, installPanel)
+					c.config.Install.Networks[netName] = network.AddDefaultBondOptions()
+				}
```
weihanglo, line 1276: There is a subtle that not every NIC config is a bond. We may not need this *for every network* but management NIC only, though it wouldn't affect anything if a `Network` doesn't contain any interfaces with bond options on it.

weihanglo: You may want to check the length of `network.interfaces` to determine this behavior

pkg/console/install_panels.go
```diff
@@ -805,14 +805,16 @@ func addNetworkPanel(c *Console) error {
...
-		if mgmtNetwork.BondOption.Mode == "" {
+		if mgmtNetwork.BondOptions == nil {
 			askBondModeV.Value = config.BondModeBalanceTLB
 		}
 		return nil
 	}
 	askBondModeVConfirm := func(g *gocui.Gui, v *gocui.View) error {
 		mode, err := askBondModeV.GetData()
-		mgmtNetwork.BondOption.Mode = mode
+		mgmtNetwork.BondOptions = make(map[string]string)
```
guangbochen, line 815: create a slice with a default length is always preferred.

pkg/console/install_panels.go
johnliu55tw: @weihanglo What do you think about moving default options to here? https://github.com/harvester/harvester-installer/blob/ff0bd2f4a11e24f5d762546399c93c0c7a2f7337/pkg/config/cos.go#L368

PR #162, opened 2021-11-10
[DNM] Add "SystemSettings" config

johnliu55tw:
**Note**: Do not merge as we might switch to ManagedChart.

Add "SystemSettings" config and renders it as Setting CR resources with special annotation for harvester controller to load.

### DIFF
--- a/pkg/config/cos.go
+++ b/pkg/config/cos.go
@@ -15,11 +15,12 @@ import (
 )
 
 const (
-	cosLoginUser       = "rancher"
-	manifestsDirectory = "/var/lib/rancher/rke2/server/manifests/"
-	canalConfig        = "rke2-canal-config.yaml"
-	harvesterConfig    = "harvester-config.yaml"
-	ntpdService        = "systemd-timesyncd"
+	cosLoginUser            = "rancher"
+	manifestsDirectory      = "/var/lib/rancher/rke2/server/manifests/"
+	canalConfig             = "rke2-canal-config.yaml"
+	harvesterConfig         = "harvester-config.yaml"
+	harvesterSystemSettings = "harvester-settings.yaml"
+	ntpdService             = "systemd-timesyncd"
 
 	networkConfigDirectory = "/etc/sysconfig/network/"
 	ifcfgGlobPattern       = networkConfigDirectory + "ifcfg-*"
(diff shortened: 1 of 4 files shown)


pkg/config/templates/settings.yaml
```diff
@@ -0,0 +1,9 @@
+{{- range $name, $value := .SystemSettings }}
+---
+apiVersion: harvesterhci.io/v1beta1
+kind: Setting
+metadata:
+  annotations:
+    harvesterhci.io/preconfigValue: {{ $value }}
```
bk201, line 7: Could you check if the passed-in `$value` is boolean or number works?
I feel we might need to quote here.

pkg/config/cos.go
```diff
@@ -130,10 +130,33 @@ func ConvertToCOS(config *HarvesterConfig) (*yipSchema.YipConfig, error) {
...
+	if len(config.SystemSettings) != 0 {
+		settingCrs, err := render("settings.yaml", config)
+		if err != nil {
+			return nil, err
+		}
+
+		fsStages = append(fsStages, yipSchema.Stage{
+			Files: []yipSchema.File{
+				{
+					Path:        "/var/lib/rancher/rke2/server/manifests/settings.yaml",
```
bk201, line 145:
```suggestion
					Path:        "/var/lib/rancher/rke2/server/manifests/harvester-settings.yaml",
```

It would be nice to add a prefix. The directory is a bit crowded.

```
node1:~ # ls /var/lib/rancher/rke2/server/manifests/
harvester-config.yaml      monitoring-crd.yaml        monitoring.yaml            rke2-canal-config.yaml     rke2-coredns.yaml          rke2-multus.yaml
harvester.yaml             monitoring-dashboard.yaml  rancher/                   rke2-canal.yaml            rke2-metrics-server.yaml
```

I also consider moving all harvester things to a new directory `/var/lib/rancher/rke2/server/manifests/harvester`.
But considering we might switch back to ManagedChart for the upgrade, we can skip it now.

johnliu55tw: Ok, I'll rename it to `harvester-settings.yaml`.

pkg/config/templates/settings.yaml
johnliu55tw: Good catch! Although installer doesn't report any error, the annotation would disappear if not quoted.

johnliu55tw: @bk201 Updated. Please help review again, thanks!

johnliu55tw:
We have switched to `ManagedChart`. Close this one.

PR #163, opened 2021-11-10
Check iso_url in automatic install mode like iPXE

w13915984028:
Problem:
 When iso_url is left blank in harvester automatic install like iPXE, no proper check is done, system goes into unexpected state

Solution:
 Add check in harv-install shell script,  when iso_url is blank in automatic mode, stop install and show error info.

Related issue:  harvester/harvester#1439

Test plan:
1. Set proper iPXE boot environment
2. In config file (e.g. config-create.yaml), set iso_url be blank;  (BTW: given the iso_url is wrong/unreachable, there has already been some checks and error/info prompt)
3. Boot the VM or baremetal via iPXE
4. Expect: Install will stop, and show info "Failed to read iso_url, it must be set in automatic install mode"

### DIFF
--- a/package/harvester-os/files/usr/sbin/harv-install
+++ b/package/harvester-os/files/usr/sbin/harv-install
@@ -59,6 +59,11 @@ get_iso()
         get_url ${COS_INSTALL_ISO_URL} ${TEMP_FILE}
         ISO_DEVICE=$(losetup --show -f $TEMP_FILE)
         mount -o ro ${ISO_DEVICE} ${ISOMNT}
+    else
+        if [ -n "$COS_INSTALL_AUTOMATIC" ]; then
+            echo "Failed to read iso_url, it must be set in automatic install mode"
+            return 1
+        fi
     fi
 }
 
--- a/package/harvester-os/files/usr/sbin/harv-install
+++ b/package/harvester-os/files/usr/sbin/harv-install
@@ -59,11 +59,6 @@ get_iso()
         get_url ${COS_INSTALL_ISO_URL} ${TEMP_FILE}
         ISO_DEVICE=$(losetup --show -f $TEMP_FILE)
         mount -o ro ${ISO_DEVICE} ${ISOMNT}
-    else
-        if [ -n "$COS_INSTALL_AUTOMATIC" ]; then
-            echo "Failed to read iso_url, it must be set in automatic install mode"
-            return 1
-        fi
     fi
 }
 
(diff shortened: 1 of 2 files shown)


package/harvester-os/files/usr/sbin/harv-install
```diff
@@ -59,6 +59,9 @@ get_iso()
         get_url ${COS_INSTALL_ISO_URL} ${TEMP_FILE}
         ISO_DEVICE=$(losetup --show -f $TEMP_FILE)
         mount -o ro ${ISO_DEVICE} ${ISOMNT}
+    else
+        echo "Failed to read iso_url, it must be set in automatic install mode"
+        return 1
```
bk201, line 64: get_iso is also called in ISO installation mode.
So this breaks the ISO installation.

w13915984028: from source code:
get_iso --> get_url, which either "curl remote" or " cp local"

I guess:  for ISO install, it "cp local"
 in both cases, get_iso is called, and "$COS_INSTALL_ISO_URL" should not be empty normally. 

Will do an ISO install to check.

```
get_url()
{
    FROM=$1
    TO=$2
    case $FROM in
        ftp*|http*|tftp*)
            n=0
            attempts=5
            until [ "$n" -ge "$attempts" ]
..
                curl -o $TO -fL ${FROM} && break
     ..
        *)
            cp -f $FROM $TO
            ;;
    esac
}
```

w13915984028: Verified, the ISO install failed. @bk201  

the "$COS_INSTALL_ISO_URL" is NULL when ISO install.

needs further check, thanks.

w13915984028: @bk201 Please take a look of the latest commit, thanks.

https://github.com/harvester/harvester-installer/pull/163/commits/52a21c58089d37e2a23932cbfee629d0d79c03bf

w13915984028:
Both PXE and ISO installation works as expected.

image

package/harvester-os/files/usr/sbin/harv-install
bk201: @w13915984028 The fix looks nice. But do you think it's better to put the checks in the validation code for consistency? Since we do most of the configuration validation there, other contributors might be easier to know what validations are done.

I did test my suggestion in https://github.com/harvester/harvester/issues/1439#issuecomment-965068891 and it works.

w13915984028: > @w13915984028 The fix looks nice. But do you think it's better to put the checks in the validation code for consistency? Since we do most of the configuration validation there, other contributors might be easier to know what validations are done.
> 
> I did test my suggestion in [harvester/harvester#1439 (comment)](https://github.com/harvester/harvester/issues/1439#issuecomment-965068891) and it works.


@bk201 I also did a quick test with your change, it really works.  
image


Sure, it is better to put it together with other checks.

The "validateConfig" runs after "Automatic" filed is set, my previous understanding of code flow was a bit wrong. thanks.

Please submit your change.

PR #165, opened 2021-11-12
Check iso_url in automatic install mode like iPXE

w13915984028:
Problem:
When iso_url is left blank in harvester automatic installation like iPXE, no proper check is done, system goes into unexpected state

Solution:
Add check in validator, when iso_url is blank in automatic mode, stop installation and show related error message.

Related issue: harvester/harvester#1439

Test plan:

Set proper iPXE boot environment
In config file (e.g. config-create.yaml), set iso_url be blank; (BTW: given the iso_url is wrong/unreachable, there has already been some checks and error/info prompt)
Boot the VM / baremetal via iPXE
Expect: Install will stop, and show info `iso_url is required in automatic installation`


guangbochen (approved):
LGTM

bk201 (approved):
LGTM

PR #166, opened 2021-11-15
Use managed chart to deploy Harvester

johnliu55tw:
## Overview
- Remove `manifests/` directory and move these charts and resources to `pkg/config/templates/` as Rancherd's `bootstrapResources`. We need   them to be Go templates because we need to patch some of them during the installation process. If they are deployed by cOS immutably, we can't update them.

- Add `HarvesterChartVersion` and `MonitoringChartVersion` to `config.go` because we need to inject these versions into the templates. These versions are provided by the build scripts.

- ~Fix Rancher monitoring issue: ManagedChart's resources will have a prefix "mcc-" added to their name when they are installed. This would break monitoring dashboard because GUI expects a service named "rancher-monitoring-grafana" to provide Grafana service, not "mcc-rancher-monitoring-grafana". To fix this issue, we need to add a Service with correct name and points to Grafana Pods.~ We can use `releaseName` field of `ManagedChart` to specify the release name, to avoid the `mcc-` prefix.

- Update installer to adapt to ManagedChart resources.

See also https://github.com/harvester/harvester-installer/pull/108 for a reference.

## TODOs
- [x] Create another Rancher config for `bootstrapResource`. Right now it's in `rancherd-config.yaml` with other Rancherd configs.
- [x] Remove hard-coded harvester ManagedChart `version`
- [x] Rancher monitoring still uses RKE2's manifests directory for deploying. Maybe we could use ManagedChart also.
- [x] Check `build-bundle` for the effect of changing `CHARTS_DIR`.
- [x] Fix Rancher monitoring issue

## Known issues
- VIP on the terminal dashboard appears a bit too late.
  - The `Harvester management URL` shows the real IP instead of VIP for a while after cluster status shows `Ready`
    image
    After a couple of minutes it shows the VIP:
    image


pkg/config/cos.go
```diff
@@ -446,3 +462,31 @@ func getAddStaticDNSServersCmd(servers []string) string {
...
+		"13-monitoring.yaml",
+	} {
+		rendered, err := render("rancherd-"+templateName, config)
+		if err != nil {
+			return nil, err
+		}
+
+		bootstrapConfs[templateName] = rendered
+	}
+	// It's not a template but I still put it here for consistency
```
bk201, line 483: Normally it's no harm to render a file without actions in it.
But given that `12-monitoring-dashboard.yaml` contains a long JSON text and there are also `{{ }}` text in it,  +1 to treat the file separately.

guangbochen:
I'm wondering after upgrading the helmChart from the manifest to the `managedChart` will we be able to persist the previous data including LH volumes?

pkg/config/config.go
```diff
@@ -104,9 +104,11 @@ type HarvesterConfig struct {
...
 	Token     string `json:"token,omitempty"`
 
-	OS             `json:"os,omitempty"`
-	Install        `json:"install,omitempty"`
-	RuntimeVersion string `json:"runtimeVersion,omitempty"`
+	OS                     `json:"os,omitempty"`
+	Install                `json:"install,omitempty"`
+	RuntimeVersion         string `json:"runtimeVersion,omitempty"`
+	HarvesterChartVersion  string `json:"harvesterChartVersion,omitempty"`
+	MonitoringChartVersion string `json:"monitoringChartVersion,omitempty"`
```
gitlawr, line 111: Why do we want these options in the harvester config? They seem to be fixed for each Harvester release.

bk201: The purpose was to test different RKE2 versions from PXE boot installation.
If they are not required, we can drop them.

johnliu55tw: For `HarvesterChartVersion`, a release of Harvester will have a version string different from `0.0.0-dev` ([REF: v0.3.0](https://github.com/harvester/harvester/blob/842a9942011fa8d113897b910efd1fec9271666e/deploy/charts/harvester/Chart.yaml#L3)), so we need to inject this value into `ManagedChart` for it to work correctly.

As for `MonitoringChartVersion`, this was a variable in `build-bundle` script ([REF](https://github.com/harvester/harvester-installer/blob/be59d22082c0a0de4789ffba8cf6113ebc7d1233/scripts/build-bundle#L20)) so it can be changed during build time. To keep this behavior, I add it here so to inject the version into the `ManagedChart`.

johnliu55tw:
> I'm wondering after upgrading the helmChart from the manifest to the managedChart will we be able to persist the previous data including LH volumes?

@guangbochen Good question, I need to test how `ManagedChart` would interact with the existing HelmChart.

bk201:
Overall the cluster looks good, but I noticed two issues in bundle deployments:
Screen Shot 2021-11-16 at 4 39 06 PM

I saw the longhorn storage class one before but didn't figure out why yet.

bk201:
@johnliu55tw For a ManagedChart, it looks like we can specify `spec.releaseName` to remove the `mcc-` prefix.
https://github.com/rancher/rancher/blob/d2c5498762b984c8cb0462c9aa894e6943aec526/pkg/apis/management.cattle.io/v3/managed_chart.go#L23

johnliu55tw:
> I'm wondering after upgrading the helmChart from the manifest to the managedChart will we be able to persist the previous data including LH volumes?

@guangbochen Good question, I need to test how `ManagedChart` would interact with the existing HelmChart.

> @johnliu55tw For a ManagedChart, it looks like we can specify `spec.releaseName` to remove the `mcc-` prefix. https://github.com/rancher/rancher/blob/d2c5498762b984c8cb0462c9aa894e6943aec526/pkg/apis/management.cattle.io/v3/managed_chart.go#L23

Great! I'll check this out, thanks!

johnliu55tw:
> I saw the longhorn storage class one before but didn't figure out why yet.
@bk201 I saw this error occasionally too, but it always went away after a while. I think this is not entirely related to ManagedChart?

bk201:
> I'm wondering after upgrading the helmChart from the manifest to the `managedChart` will we be able to persist the previous data including LH volumes?

I did some tests on 0.3.0 to migrate from RKE2 HelmChart to ManagedChart.
Creating a ManagedChart like this to take over ownership:
```
apiVersion: management.cattle.io/v3
kind: ManagedChart
metadata:
  name: harvester
  namespace: fleet-local
spec:
  takeOwnership: true           # <-- seems to be important, but I found it seems to not necessary for Harvester chart
  chart: harvester
  version: 0.0.0-dev
  releaseName: harvester        # <-- Without this, the release name will be prefixed with `mcc-`
...
```

And we also need to clean RKE2 manifests. Otherwise, RKE2 will re-deploy the HelmChart again on restart. (RKE2 **deletes** a release and re-deploy if it can't successfully upgrade a chart, which makes all VMs gone)
```
        sudo rm -f /var/lib/rancher/rke2/server/manifests/harvester.yaml
        sudo rm -f /var/lib/rancher/rke2/server/manifests/monitoring-crd.yaml
        sudo rm -f /var/lib/rancher/rke2/server/manifests/monitoring-dashboard.yaml
        sudo rm -f /var/lib/rancher/rke2/server/manifests/monitoring.yaml
```

This procedure can be applied to other charts too.

guangbochen:
I'm wondering if there is an option to not delete the helmChart release if by any means it was failed. So that we won't take risks of losing the VMs and volumes of the cluster. If any error occurs we should be able to either upgrade or resolve the chart error manually instead. Will need some investigation on it, thanks.

bk201 (approved):
LGTM. I wonder if we should publish the `rancher/harvester-cluster-repo` image.
it will be preloaded from the installer anyway.

guangbochen:
> LGTM. I wonder if we should publish the `rancher/harvester-cluster-repo` image. it will be preloaded from the installer anyway.

is there any cases that we will use this image independently for testing or validation, if yes we can manage to do it even for future usage, thanks

pkg/config/cos.go
```diff
@@ -446,3 +462,31 @@ func getAddStaticDNSServersCmd(servers []string) string {
...
 }
+
+// Returns Rancherd bootstrap resources
+// map: fileName -> fileContent
+func genBootstrapResources(config *HarvesterConfig) (map[string]string, error) {
+	bootstrapConfs := make(map[string]string, 4)
+
+	for _, templateName := range []string{
+		"10-harvester.yaml",
+		"11-monitoring-crd.yaml",
```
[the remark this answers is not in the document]
bk201: When Rancherd applies these bootstrap resources, the KubeVirt might not be up yet.
(It also applies with `--validate=false`)
https://github.com/rancher/rancherd/blob/5fd1a272e0e58f3a656a893ece0b5d26b0d74e54/pkg/resources/resources.go#L194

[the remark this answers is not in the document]
johnliu55tw: Just did some tests and I think there's no guarantee of which ManagedChart would be installed first, even if we order them like so.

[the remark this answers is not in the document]
guangbochen: sure, then it should be fine for us to leave what it is, thanks

guangbochen (approved):
LGTM

package/harvester-os/templates/91-harvester-bootstrap-repo.yaml
```diff
@@ -0,0 +1,39 @@
...
+- apiVersion: apps/v1
+  kind: Deployment
+  metadata:
+    name: harvester-cluster-repo
+    namespace: cattle-system
+  spec:
+    selector:
+      matchLabels:
+        app: harvester-cluster-repo
+    replicas: 1
```
gitlawr, line 11: Do we need HA for the repository, what if it's down during an upgrade?

bk201: Just tested it, draining the node that contains the pod will re-create a new pod on another node.
So maybe it's not needed? We'll preload images on each node before upgrading Harvester.

gitlawr (approved):
LGTM

PR #167, opened 2021-11-16
Add Node labels config

johnliu55tw:
* Use `kubelet-arg` in RKE2 config to bootstrap node labels. See https://github.com/rancher/rancherd/issues/16.

* Remove `node-label+` from `rke2-90-harvester-agent.yaml` because we already have those default labels defined in `rancherd-config.yaml`.

## TODOs
- [x] Add documents for this config: https://github.com/harvester/docs/pull/76

## Related issues
- https://github.com/harvester/harvester/issues/1416

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -2,8 +2,10 @@ package config
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/imdario/mergo"
+	"k8s.io/apimachinery/pkg/util/validation"
 )
 
 const (
@@ -98,6 +100,7 @@ type OS struct {
 	Wifi           []Wifi            `json:"wifi,omitempty"`
 	Password       string            `json:"password,omitempty"`
 	Environment    map[string]string `json:"environment,omitempty"`
+	Labels         map[string]string `json:"labels,omitempty"`
 }
 
 type HarvesterConfig struct {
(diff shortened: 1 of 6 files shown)


pkg/config/config.go
```diff
@@ -100,13 +101,18 @@ type OS struct {
...
+}
+
 type HarvesterConfig struct {
 	ServerURL string `json:"serverUrl,omitempty"`
 	Token     string `json:"token,omitempty"`
 
 	OS             `json:"os,omitempty"`
 	Install        `json:"install,omitempty"`
 	RuntimeVersion string `json:"runtimeVersion,omitempty"`
+	Node           `json:"node,omitempty"`
```
bk201, line 115: Harvester config is[ originated from k3os config](https://github.com/rancher/k3os#k3oslabels). Maybe putting under `os`?

pkg/config/config.go
```diff
@@ -145,3 +151,19 @@ func (c *HarvesterConfig) String() string {
 	}
 	return fmt.Sprintf("%+v", *s)
 }
+
+func (c *HarvesterConfig) GetKubeletArgs() []string {
+	// node-labels=key1=val1,key2=val2
+	labelStrs := make([]string, 0, len(c.Labels))
+	for key, val := range c.Labels {
+		labelStrs = append(labelStrs, fmt.Sprintf("%s=%s", key, val))
```
gitlawr, line 159: Can we add some validation? e.g., `k8s.io/apimachinery/pkg/util/validation.IsQualifiedName/IsValidLabelValue`

pkg/config/config.go
johnliu55tw: Didn't know it's from k3os. Will move `labels` to under `os`.

pkg/config/config.go
johnliu55tw: Good idea. Will add this validation.

gitlawr (approved):
LGTM

bk201 (approved):

PR #168, opened 2021-11-17
Support using symbolic link as the install device

bk201:
Users can use a symbolic link like `/dev/disk/by-id/xxx` to select an install device.
This is useful when device enumeration (`/dev/sda`...) is not persistent across reboots.

**Related issue**
https://github.com/harvester/harvester/issues/1462

**Test plan**
1. In a running OS, select a symbolic link to a disk. You can run the following command to see the links:
   ```
   ls /dev/disk/by-id -alh
   ```
   
   To test in a VM with Virtio disks, you can choose links in `/dev/disk/by-path/` because Virtio disks don't have ID.
2. Use symbolic link name as `os.install.device` to perform a PXE boot installation.

### DIFF
--- a/package/harvester-os/files/usr/sbin/harv-install
+++ b/package/harvester-os/files/usr/sbin/harv-install
@@ -246,8 +246,14 @@ trap cleanup exit
 # For PXE Boot
 get_iso
 
+# Follow the symbolic link to the real device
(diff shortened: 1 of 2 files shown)


gitlawr (approved):
LGTM

johnliu55tw (approved):
LGTM

PR #169, opened 2021-11-18
Use managedchart to deploy setting CR

johnliu55tw:
Now we have `ManagedChart`, we could easily deploy these Setting CR not worrying RKE2 manifests reboot overwriting issue.

## TODOs
- [x] Revert "preconfig-loader" commit: https://github.com/harvester/harvester/pull/1563

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -109,9 +109,10 @@ type HarvesterConfig struct {
 
 	OS                     `json:"os,omitempty"`
 	Install                `json:"install,omitempty"`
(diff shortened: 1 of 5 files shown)


gitlawr (approved):
LGTM. Pretty neat now.

bk201 (approved):
Test OK and LGTM.

Do you think there is a better term for the settings?
I'm not quite sure, but `systemSettings` feels like something to be applied on a certain node.

guangbochen (approved):

PR #171, opened 2021-11-22
Use VIP in RKE2 Server URL for agent nodes

bk201:
An agent node should use VIP to reach HA server nodes because
a server node might fail.

**Related issue**
https://github.com/harvester/harvester/issues/1521

### DIFF
--- a/package/harvester-os/files/etc/systemd/system/rke2-agent.service.d/override.conf
+++ b/package/harvester-os/files/etc/systemd/system/rke2-agent.service.d/override.conf
@@ -1,0 +1,2 @@
+[Service]
+ExecStartPre=/usr/sbin/harv-update-rke2-server-url agent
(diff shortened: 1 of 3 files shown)


guangbochen (approved):
LGTM

johnliu55tw:
The overall logic LGTM, but I'm wondering maybe we don't need to add `rke2-server.service.d/override.conf` at all? Because `90-harvester-vip.yaml` is created dynamically during bootstrap stage, so for the server node, this file would never exist.

johnliu55tw (approved):

gitlawr:
Is this a temporary workaround? Do we have a related issue to track the upstream fix?

bk201:
> The overall logic LGTM, but I'm wondering maybe we don't need to add `rke2-server.service.d/override.conf` at all? Because `90-harvester-vip.yaml` is created dynamically during bootstrap stage, so for the server node, this file would never exist.

We promote agent nodes to server nodes. It's used to remove the file on a promoted node.

bk201:
> Is this a temporary workaround? Do we have a related issue to track the upstream fix?

Rancher doesn't have the knowledge of VIP. I can create a feature request to track this.

gitlawr (approved):
LGTM

PR #173, opened 2021-11-24
Enable ingress controller

gitlawr:
https://github.com/harvester/harvester/issues/761

### DIFF
--- a/pkg/config/templates/rancherd-10-harvester.yaml
+++ b/pkg/config/templates/rancherd-10-harvester.yaml
@@ -20,6 +20,45 @@ bootstrapResources:
       release: rancher-monitoring
     name: rancher-monitoring-operator
     namespace: cattle-monitoring-system
+- apiVersion: helm.cattle.io/v1
+  kind: HelmChartConfig
+  metadata:
+    name: rke2-ingress-nginx
+    namespace: kube-system
+  spec:
+    valuesContent: |-
+      controller:
+        admissionWebhooks:
+          port: 8444
+        publishService:
+          pathOverride: kube-system/ingress-expose
+        extraArgs:
+          default-ssl-certificate: cattle-system/tls-rancher-internal
+- apiVersion: v1
+  kind: Secret
+  metadata:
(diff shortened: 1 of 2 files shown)


pkg/config/templates/rancherd-10-harvester.yaml
```diff
@@ -20,6 +20,45 @@ bootstrapResources:
...
+          default-ssl-certificate: cattle-system/tls-rancher-internal
+- apiVersion: v1
+  kind: Secret
+  metadata:
+    name: tls-ingress
+    namespace: cattle-system
+  type: Opaque
+  data:
+    tls.crt: Cg==
+    tls.key: Cg==
```
guangbochen, line 45: Q: is pre-render the empty `tls-ingress` secret required?

pkg/config/templates/rancherd-10-harvester.yaml
```diff
@@ -20,6 +20,45 @@ bootstrapResources:
...
+- apiVersion: helm.cattle.io/v1
+  kind: HelmChartConfig
+  metadata:
+    name: rke2-ingress-nginx
+    namespace: kube-system
+  spec:
+    valuesContent: |-
+      controller:
+        admissionWebhooks:
+          port: 8444
```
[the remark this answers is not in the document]
gitlawr: It's a hostPort, it defaults to 8443. Set it to 8444 to avoid conflicting with the rancher hostPort.

pkg/config/templates/rancherd-10-harvester.yaml
gitlawr: We can either create it here or create it on the Harvester server side.

guangbochen: both works for me, we can leave as what it is, thanks.

guangbochen (approved):
LGTM

johnliu55tw (approved):
LGTM

bk201 (approved):

PR #174, opened 2021-11-25
Allow OWASP password special characters as token 

johnliu55tw:
Properly escape the token value to allow using OWASP password special characters as cluster token value.

See https://owasp.org/www-community/password-special-characters

### Positive test case
1. Use ISO to install Harvester in "create" mode
2. Enter any combination of alphanumeric and [OWASP password special characters](https://owasp.org/www-community/password-special-characters) as Cluster Token
3. Use ISO to install Harvester in "join" mode
4. Enter the same cluster token.
5. Make sure it can join a cluster using the same token value.

### Negative test case
**NOTE:** It's not that easy to enter non-ASCII characters using installer GUI, so here I recommend using iPXE to perform the negative test case.

1. Use iPXE to install Harvester: Change the token value of config file to something invalid, such as `Äöê`
2. Verify that the installation process failed with an error message showing the token value is invalid.

## TODOs
- [ ] Add a troubleshooting guide for "Entered incorrect cluster token during the installation process.": https://github.com/harvester/docs/pull/81

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -190,3 +190,41 @@ func TestHarvesterSystemSettingsRendering_AsEmptyArrayIfNoSetting(t *testing.T) 
 	assert.NotNil(t, bootstrapResources)
 	assert.Equal(t, 0, len(bootstrapResources))
 }
+
+func TestHarvesterTokenRendering(t *testing.T) {
+	// Test the Token value is escaped correctly
+	testCases := []struct {
+		name  string
+		token string
+	}{
+		{
+			name:  "Test OWASP password special characters",
+			token: " !\"#$%&'()*+,-./:;<=>?@[\\]^_`{|}~",
+		},
+		{
+			name:  "Test mixed characters",
+			token: "Hello, I opened a new bar! It's called \"FOOBAR\". \\YES/",
+		},
+	}
+
+	for _, testCase := range testCases {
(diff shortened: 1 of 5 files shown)


gitlawr:
Do you have any reference regarding how rke2 validates the token?

johnliu55tw:
> Do you have any reference regarding how rke2 validates the token?

Based on my tests [here](https://github.com/harvester/harvester/issues/812#issuecomment-931842041), I think RKE2 has no special restriction on the token's value.

gitlawr:
Why do we add the restriction if the token is valid then? Complexity helps to improve token strength

guangbochen:
I leave a comment previously in https://github.com/harvester/harvester/issues/812#issuecomment-963798589 as I think it would be tedious for us to support all special cases in the token validation, so adding a validation probably would be easier as most people won't care it is managed by RKE2 or another system.

johnliu55tw:
@gitlawr @guangbochen The current implementation is to properly escape the token value while rendering it into the template `rancherd-config.yaml` so that we could support using OWASP password special characters as token value. The validation is to make sure we only have ASCII printable characters, not UTF-8 characters.

guangbochen (approved):
LGTM

bk201 (approved):
LGTM. Nice enhancement.

gitlawr (approved):
LGTM

PR #175, opened 2021-11-25
feat: add vlan option to network

FrankYang0529:
issue: https://github.com/harvester/harvester/issues/1390

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -35,6 +35,7 @@ type Network struct {
 	Gateway      string             `json:"gateway,omitempty"`
 	DefaultRoute bool               `json:"-"`
 	BondOptions  map[string]string  `json:"bondOptions,omitempty"`
+	VlanID       int                `json:"vlanId,omitempty"`
 }
 
 type HTTPBasicAuth struct {
--- a/pkg/config/cos.go
+++ b/pkg/config/cos.go
@@ -341,6 +341,9 @@ func UpdateNetworkConfig(stage *yipSchema.Stage, networks map[string]Network, ru
 		} else {
 			err = updateNIC(stage, name, &network)
 		}
+		if network.VlanID != 0 {
+			err = updateVLAN(stage, name, &network)
+		}
 		if err != nil {
 			return err
 		}
@@ -361,7 +364,7 @@ func UpdateNetworkConfig(stage *yipSchema.Stage, networks map[string]Network, ru
 	}
 
 	if run {
-		stage.Commands = append(stage.Commands, fmt.Sprintf("wicked ifreload all"))
+		stage.Commands = append(stage.Commands, "wicked ifreload all")
 
 		// in case wicked config is not changed and netconfig is not called
 		stage.Commands = append(stage.Commands, "netconfig update")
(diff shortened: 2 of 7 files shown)


pkg/config/config.go
```diff
@@ -28,13 +28,15 @@ const (
 )
 
 type Network struct {
+	Name         string             `json:"name,omitempty"`
```
bk201, line 31: Do we need to introduce the `Name` field? We use the network map key as the network name.
This will give users the chance to have a mismatch between the network key and the `Name` field.

pkg/config/cos.go
```diff
@@ -434,6 +442,23 @@ func updateBond(stage *yipSchema.Stage, name string, network *Network) error {
 	return nil
 }
 
+func updateVLAN(stage *yipSchema.Stage, name string, network *Network) error {
+	n := networkWithName{Name: name, Network: *network}
```
weihanglo, line 446: NIT: anonymous struct might be better for local use?

```suggestion
	n := struct {
		Name    string
		Network Network
	}{
		Name: name,
		Network: *network,
	}
```

pkg/config/templates/wicked-ifcfg-vlan
```diff
@@ -0,0 +1,13 @@
+STARTMODE='onboot'
+BOOTPROTO='{{ .Method }}'
```
weihanglo, line 2: Shouldn't it be `.Network.Method`? Or use `with` action for convenience.

pkg/config/cos.go
FrankYang0529: Updated it. Thank you.

pkg/config/templates/wicked-ifcfg-vlan
FrankYang0529: I update the struct like the following, so we can keep wicked files structure same.

```
struct {
  Name string
  Network
}
```

pkg/config/config.go
FrankYang0529: Removed it. Thank you.

pkg/console/validator.go
```diff
@@ -201,6 +202,11 @@ func checkNetworks(networks map[string]config.Network) error {
 		default:
 			return prettyError(ErrMsgNetworkMethodUnknown, network.Method)
 		}
```
w13915984028, line 204: VlanID > 4094, is error

https://en.wikipedia.org/wiki/IEEE_802.1Q
VLAN identifier (VID)
A 12-bit field specifying the VLAN to which the frame belongs. The values of 0 and 4095 (0x000 and 0xFFF in hexadecimal) are reserved. All other values may be used as VLAN identifiers, allowing up to 4,094 VLANs. The reserved value 0x000 indicates that the frame does not carry a VLAN ID

pkg/console/validator.go
```diff
@@ -35,6 +35,7 @@ var (
 	ErrMsgForceMBROnUEFI               = "cannot force MBR on UEFI system"
 
 	ErrMsgNetworkMethodUnknown = "unknown network method"
+	ErrMsgNetworkVLANID        = "vlan id should be 1 to 4096"
```
w13915984028, line 38: 1~4094  or  0~4094

pkg/console/validator.go
FrankYang0529: Updated it. Thanks :)

pkg/console/validator.go
FrankYang0529: Updated it. Thanks 😀

bk201 (approved):
LGTM

weihanglo (approved):
Haven't setup a VLAN on KVM to test, but tested with templating and it rendered correctly. I think it's good to move forward to test in bare metal.

PR #178, opened 2021-11-30
Prompt users to use VIP to join a cluster

bk201:
Users should use VIP to join a cluster because a server node might fail.

**Related issue**
https://github.com/harvester/harvester/issues/1521


guangbochen (approved):

guangbochen (commented):
LGTM

PR #180, opened 2021-11-30
Bump OS base image to 20211130

bk201:
- Partition creation is enhanced in yip.
- Depends on a fixed OS snapshot image instead of the unpredictable `dev-head` tag.


guangbochen (approved):
LGTM

johnliu55tw (approved):

PR #181, opened 2021-12-02
Get VIP through DHCP for PXE boot

FrankYang0529:
Issue https://github.com/harvester/harvester/issues/1410

### Test Case

* Comment `vip` and `vip_hw_addr` in `ipxe-examples/vagrant-pxe-harvester/ansible/roles/harvester/templates/config-create.yaml.j2`
* Start vagrant-pxe-harvester
* Run `kubectl get cm -n harvester-system vip`
  * Check whether we can get `ip` and `hwAddress` in it
* Run `ip a show harvester-mgmt`
  * Check whether there is two IP in it and one is vip.

### DIFF
--- a/pkg/console/install_panels.go
+++ b/pkg/console/install_panels.go
@@ -1364,7 +1364,23 @@ func addInstallPanel(c *Console) error {
 					return
 				}
 				logrus.Info("Local config (merged): ", c.config)
+
+				if needToGetVIPFromDHCP(c.config.VipMode, c.config.Vip, c.config.VipHwAddr) {
+					printToPanel(c.Gui, "Configuring network...", installPanel)
+					if _, err := applyNetworks(c.config.Networks); err != nil {
+						printToPanel(c.Gui, fmt.Sprintf("can't apply networks: %s", err), installPanel)
+						return
+					}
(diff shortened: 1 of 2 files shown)


pkg/console/install_panels.go
```diff
@@ -1384,7 +1384,23 @@ func addInstallPanel(c *Console) error {
 					return
 				}
 				logrus.Info("Local config (merged): ", c.config)
+
+				if strings.ToLower(c.config.VipMode) == config.NetworkMethodDHCP && (c.config.Vip == "" || c.config.VipHwAddr == "") {
+					if _, err := applyNetworks(c.config.Networks); err != nil {
```
weihanglo, line 1389: Maybe print `Configuring network" before applying? I remember this take some time.

weihanglo (commented):
Look good overall! I'll do a e2e test ASAP.

pkg/console/install_panels.go
FrankYang0529: Updated it. Thank you.

weihanglo (approved):
Build from harvester master with this PR. Works like a breeze!

bk201 (approved):
LGTM

PR #183, opened 2021-12-06
Ability to lookup mac addresses

ibrokethecloud:
When trying to install Harvester on public metal cloud providers like Equinix Metal, we run into an issue where the network device names are not predictable. This is because even with the same instance types, based on where the network adapter is attached on the PCI bus, the device name will change.

One option is to disable predictable device naming during PXE boot, which doesnt work, since the harvester installation will revert the installed OS back to predicatable device naming and the node never gets ready post installation since the device names now have changed.

A simpler fix may be to leave the NetworkInterface `Name` empty and only provide a `HwAddr`. 

This PR allows users to do the same, but looking up the HwAddr and querying the OS interfaces to get the correct interface name associated with the HwAddr.

This makes it a lot easier to configure and consume Harvester in a public cloud provider with metal instances since the HwAddr for interfaces can be queried easily.

A sample create.yaml with this configuration would be as follows:

```
token: token
os:
  hostname: node1
  password: rancher

install:
  mode: create
  automatic: true
  debug: true
  device: /dev/sda
  iso_url: http://172.16.135.50:8080/harvester-v0.3.0-amd64.iso
  poweroff: false
  vip: 172.16.128.12
  vip_mode: static
  networks:
    harvester-mgmt:
      interfaces:
      - name: ""
        hwAddr: '0c:c4:7a:6b:84:20'
      default_route: true
      method: dhcp
      bond_options:
        mode: balance-tlb
        miimon: 100
```

Ref: https://github.com/harvester/harvester/issues/1641


yasker:
is `name` still a required field? If we don't know the NIC name, then set it to `""` is a bit misleading. Maybe better to only have the hwaddr there.

ibrokethecloud:
@yasker `name` is not required. That was just an example I was using. 

This would be the sample minimal definition for `interfaces`

```
  networks:
    harvester-mgmt:
      interfaces:
      - hwAddr: '0c:c4:7a:6b:84:20'
      default_route: true
      method: dhcp
      bond_options:
        mode: balance-tlb
        miimon: 100

```

pkg/console/install_panels.go
```diff
@@ -1393,6 +1393,23 @@ func addInstallPanel(c *Console) error {
 				c.config.TTY = getFirstConsoleTTY()
 			}
 
+			// lookup MAC Address to populate device names where needed
+			for key, network := range c.config.Networks {
```
johnliu55tw, line 1397: Nit: How about directly assign the updated `NetworkInterface` object?

```go
for key, network := range c.config.Networks {
	for idx, iface := range network.Interfaces {
		if err := iface.FindNetworkInterfaceName(); err != nil {
			logrus.Error(err)
			printToPanel(c.Gui, err.Error(), installPanel)
			return
		}
		c.config.Networks[key].Interfaces[idx] = iface
	}
}
```

johnliu55tw (approved):
Overall LGTM but a nit comment.

guangbochen:
@ibrokethecloud can please amend the commit to rerun the CI validation, previously there was a CI issue but it was resolved now, thanks.

ibrokethecloud:
@guangbochen CI looks a lot better now

bk201 (approved):
LGTM

PR #185, opened 2021-12-08
Cleanup vagrants leftovers before provisioning a new cluster

guangyee:
If a job exited abruptly, there could be old vagrants leftovers which must be
cleaned up prior to starting the next provisioning job.


bk201 (approved):
LGTM.

guangbochen (approved):

PR #186, opened 2021-12-08
[DNM] Patch cos for NVME drive fix

johnliu55tw:
Directly patch `cos` command for fixing https://github.com/harvester/harvester/issues/1627 in case we can't get the backported cos-installer `0.21.2` in time.

See cOS PR https://github.com/rancher-sandbox/cOS-toolkit/pull/937


johnliu55tw:
installer 0.21.2 It's published.

PR #187, opened 2021-12-09
Copy ifcfg and resolv.conf generated by dracut "ifcfg" module

johnliu55tw:
Copying `ifcfg-<iface>` and `resolv.conf` generated by dracut module `ifcfg`, which reads the artifacts created by dracut's `network` module.

## Gotchas
`ifcfg` module has some issues that we have to work around:
- It generates incorrect ifcfg for interfaces using DHCP. See https://bugzilla.opensuse.org/show_bug.cgi?id=1193518.
- The dracut `network` module creates DNS server file `net.<iface>.resolv.conf` for each interface based on the `ip=` kernel cmdline arguments, which would then be picked up by `ifcfg` module and merged into one single `resolv.conf`. However, when the interface is configured to use DHCP, the DNS server file would have a suffix `ipvX` such as `net.ens5.resolv.conf.ipv4` which cannot be recognized by `ifcfg` module, thus no `resolv.conf` will be generated. We need to manually merge these `net.<iface>.resolv.conf.ipvX` files into one.

## Known issues
- When supplying multiple `ip=...` kernel cmdline arguments, **only one of them** will be configured by dracut, therefore only the configured interface would have ifcfg generated. So for now, we can't support multiple `ip=...` kernel cmdline arguments.

## Dependencies
- https://github.com/harvester/os2/pull/25

## Related issues
- https://github.com/harvester/harvester/issues/1590


gitlawr (approved):
LGTM

bk201 (approved):
LGTM

guangbochen (approved):

PR #189, opened 2021-12-10
Update proxy options for ingress

gitlawr:
Address https://github.com/harvester/harvester/issues/1380#issuecomment-988260681

Set unlimited request size and disable request buffering(https://github.com/kubernetes/ingress-nginx/issues/6242)

### DIFF
--- a/pkg/config/templates/rancherd-10-harvester.yaml
+++ b/pkg/config/templates/rancherd-10-harvester.yaml
@@ -28,6 +28,9 @@ bootstrapResources:
   spec:
     valuesContent: |-
       controller:
+        config:
+          proxy-body-size: "0"
+          proxy-request-buffering: "off"
         admissionWebhooks:
           port: 8444
         publishService:


pkg/config/templates/rancherd-10-harvester.yaml
```diff
@@ -48,6 +48,9 @@ bootstrapResources:
   metadata:
     name: rancher-expose
     namespace: cattle-system
+    annotations:
+      nginx.ingress.kubernetes.io/proxy-body-size: "0"
+      nginx.ingress.kubernetes.io/proxy-request-buffering: "off"
```
guangbochen, line 53: prefer to set those values via the `HelmChartConfig` using https://github.com/kubernetes/ingress-nginx/blob/9c0a39636da11b7e262ddf0b4548c79ae9fa1667/charts/ingress-nginx/values.yaml#L27, e.g,

```YAML
  spec:
    valuesContent: |-
      controller:
        config:
          proxy-body-size: "0"
          proxy-request-buffering: "off"
```

gitlawr:
Updated

guangbochen (approved):
LGTM

PR #191, opened 2021-12-14
Increase OEM partition waiting timeout to 120s

bk201:
Increase the timeout value because it takes longer for a device to appear
on some systems.

**Related issue**
https://github.com/harvester/harvester/issues/1667


johnliu55tw (approved):
LGTM. However, the only thing I checked is to make sure kernel parameter `rd.cos.oemtimeout=120` is setup correctly. Supposedly this should fix the problem, but @guangbochen might need your environment to actually verify the issue.

guangbochen:
Yep, will test it once the master-head is build.

guangbochen (approved):

PR #193, opened 2021-12-14
Disable systemd messages before starting the installer

bk201:
The systemd messages can be hidden with the kernel parameter `systemd.show_status=0`.
But I don't think it's a good idea because the booting messages prior to the Harvester dashboard/installer are quite useful.
So the idea is to turn off the messager before starting the installer and turn it on after the installer is stooped. This makes messages before shutdown or reboots visible.

References:
https://www.freedesktop.org/software/systemd/man/systemd.html#systemd.show_status
https://www.freedesktop.org/software/systemd/man/systemd.html#SIGRTMIN+21

To get current show status:
```
# busctl get-property org.freedesktop.systemd1 /org/freedesktop/systemd1 org.freedesktop.systemd1.Manager ShowStatus
b false
```

**Related Issue**
https://github.com/harvester/harvester/issues/1132

### DIFF
--- a/package/harvester-os/files/usr/bin/setup-installer.sh
+++ b/package/harvester-os/files/usr/bin/setup-installer.sh
@@ -18,6 +18,15 @@ EnvironmentFile=-/etc/rancher/installer/env
 # NOTE: it doesn't work for serial console
 ExecStartPre=/usr/bin/setterm --msg off
 
+# Disable systemd messages before starting the installer
+# See https://www.freedesktop.org/software/systemd/man/systemd.html#SIGRTMIN+21
+# Send SIGRTMIN+21
+ExecStartPre=/usr/bin/kill -s 55 1
+
+# Enable systemd messager after stopping the installer
+# Send SIGRTMIN+20
+ExecStopPost=/usr/bin/kill -s 54 1
+
 # clear the original command in getty@.service
 ExecStart=
 


bk201:
@johnliu55tw I'm not sure if this can solve the issue, I'll apply the patch on my tests. Please take a look, thanks!

johnliu55tw:
I'm thinking if systemd message only output to tty1, can we start Harvester console on other virtual console and then we switch to that one to prohibit systemd messages?

bk201:
> I'm thinking if systemd message only output to tty1, can we start Harvester console on other virtual console and then we switch to that one to prohibit systemd messages?

Could you test if that works? We also need to consider headless situations like there is only serial console available.

johnliu55tw:
Yeah for the serial console case I think my proposal would not fit. I'll first verify if this PR could fix the issue.

johnliu55tw (approved):
Tested using our [ipxe-example/vagrant-pxe-harvester](https://github.com/harvester/ipxe-examples) and verified systemd messages no longer overlapped with our console dashboard. Some notes here:

- It's difficult to reproduce when using ISO to install
- It's easier to reproduce using ipxe-example for some reason
- I haven't found an effective way to reproduce this issue. I need to rebuild the ISO every time I modified the installer's code and reinitiate the whole cluster using ipxe-example.

guangbochen (approved):
LGTM

bk201:
@johnliu55tw Using a different TTY is still more elegant. I think we can move forward with this approach if possible.

PR #197, opened 2021-12-20
Fix awk expression in normalize_image

futuretea:
NF=1 should be NF==1


gitlawr:
LGTM, can you get the CI rerun and passed?

bk201 (approved):

PR #200, opened 2021-12-20
debug CI

bk201:
Signed-off-by: Kiefer Chang <kiefer.chang@suse.com>

PR #201, opened 2021-12-20
Build version defaults to master

gitlawr:
Fix the vagrant CI failure. See https://github.com/harvester/harvester-installer/pull/200#issuecomment-997611719


bk201 (approved):
Thanks for the fix.

guangbochen (approved):

PR #204, opened 2021-12-24
Preserve DHCP IP received during installation time

johnliu55tw:
- Set hostname before configure network using `hostnamectl`

- Copy wicked runtime data (`/var/lib/wicked`) to the installed system.

## Related issues
- https://github.com/harvester/harvester/issues/1682

### DIFF
--- a/pkg/console/network.go
+++ b/pkg/console/network.go
@@ -2,6 +2,7 @@ package console
 
 import (
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"os/exec"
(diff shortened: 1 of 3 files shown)


pkg/console/network.go
```diff
@@ -133,3 +134,15 @@ func listNetworkHardware() (map[string]networkHardwareInfo, error) {
 
 	return m, nil
 }
+
+func setHostname(hostname string) error {
+	// NOTE: Can't use Golang's syscall.Sethostname
+	// because it sets the "trasient hostname", not "static hostname".
+	// wicked uses static hostname to make the DHCP request.
+	cmd := exec.Command("hostnamectl", "set-hostname", fmt.Sprintf("%q", hostname))
```
bk201, line 142: LGTM.

FWIW. yip also sets hosts file.
https://github.com/mudler/yip/blob/master/pkg/plugins/hostname.go#L50-L59

bk201 (approved):

w13915984028:
LGTM

w13915984028 (approved):

pkg/console/network.go
johnliu55tw: This seems like a neater solution to set the hostname! I'll send another PR for this.

PR #206, opened 2022-01-04
Fix: CTRL-C prevents rebooting after installation

johnliu55tw:
Use context to interrupt cos-install-shutdown script

## Related issue
- https://github.com/harvester/harvester/issues/1761

### DIFF
--- a/pkg/console/util.go
+++ b/pkg/console/util.go
@@ -2,6 +2,7 @@ package console
 
 import (
 	"bufio"
+	"context"
 	"crypto/tls"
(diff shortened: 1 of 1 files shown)


FrankYang0529 (approved):
Verify we can prevent reboot if we press ctrl-c when panel shows `* Rebooting system in 5 seconds (CTRL+C to cancel)`. LGTM.

guangbochen (approved):
LGTM

bk201 (approved):
LGTM

PR #208, opened 2022-01-05
Stop installer if DNS is unset for static IP

johnliu55tw:
If DNS servers are not provided by the config file in PXE installation or from ISO Installer while the management network has static IP, the installer would stop users from proceeding with the installation.

### ISO Installation
1. Install Harvester using ISO
2. While configuring networking, set up a **static IP address**
3. Verify that while configuring DNS servers, if no DNS servers are provided, the installer will **show a warning message and stop users from proceeding with the installation**
4. Verify that if DNS servers are given, installation can proceed
5. Go back to the network configuration page. This time set it up with **DHCP**
6. Verify that if no DNS servers are provided, the installer would still allow users to proceed

### PXE Installation
1. Create a config file with static networking, such as:
    ```
    ...
    install:
      ...
      networks:
        harvester-mgmt:
          interfaces:
          - name: ens6  # The management interface name
          method: static
          ip: 192.168.0.20
          subnet_mask: 255.255.255.0
          gateway: 192.168.0.1
    ...
    ```
2. Start the iPXE installation
3. Verify that installer will stop and show a warning message about "DNS servers are required for static IP address"
4. Add `dns_nameservers` settings to the config file:
    ```
    ...
    os:
      ...
      dns_nameservers:
      - 192.168.0.1
      - 8.8.8.8
    ...
    ```
5. Restart the PXE installation and verify that installation would not be stopped.

## Related issues
- https://github.com/harvester/harvester/issues/1412

### DIFF
--- a/pkg/console/constant.go
+++ b/pkg/console/constant.go
@@ -61,7 +61,7 @@ const (
 	proxyNote              = "Note: In the form of \"http://[[user][:pass]@]host[:port]/\"."
 	sshKeyNote             = "For example: https://github.com/<username>.keys"
 	ntpServersNote         = "Note: You can use comma to add more NTP servers."
-	dnsServersNote         = "Note: You can use comma to add more DNS servers."
+	dnsServersNote         = "Note: You can use comma to add more DNS servers. Leave blank to use default DNS."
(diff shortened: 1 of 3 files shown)


FrankYang0529 (approved):
Verify static method with empty DNS will show the warning message. LGTM.

bk201 (approved):
LGTM

PR #212, opened 2022-01-10
[WIP] Refactor: Move partition layout logic into cos.go

johnliu55tw:
* Upgrade yip to 0.9.25 so we could use latest yip/schema package to define the "layout" stage.

* Move function `console.createPartitionLayout` to `config.createPartitioningStage` in cos.go as this make much more sense.


johnliu55tw:
I'll do the refactoring along with other PR.

PR #215, opened 2022-01-12
Specify OEM label in kernel parameter

bk201:
The label needs to be explicitly specified after this change in cOS:
https://github.com/rancher-sandbox/cOS-toolkit/pull/894

This fixes the problem that partition with `HARV_LH_DEFAULT` label is not mounted to `/var/lib/longhorn`.


johnliu55tw (approved):
LGTM

guangbochen (approved):

PR #216, opened 2022-01-12
VIP interface isn't needed after kube-vip v0.4.1

yaocw2020:
After kube-vip v0.4.1, since the kube-vip can use the VIP based on the default route automatically, the VIP interface parameter is not required anymore.

Related issue: 
https://github.com/harvester/harvester/issues/1681

Dependency: https://github.com/harvester/harvester/pull/1805

### DIFF
--- a/pkg/config/templates/rancherd-10-harvester.yaml
+++ b/pkg/config/templates/rancherd-10-harvester.yaml
@@ -123,16 +123,6 @@ bootstrapResources:
         enabled: true
       kube-vip:
         enabled: true
-        {{- if .Vip }}
-        config:
-          {{- $vipInterface := "harvester-mgmt" }}
-          {{- $mgmtNetwork := (index .Networks "harvester-mgmt") }}
-          {{- if ne $mgmtNetwork.VlanID 0 }}
-            {{- $vlanStr := print "vlan" $mgmtNetwork.VlanID }}
-            {{- $vipInterface = (print $vlanStr "@harvester-mgmt") }}
-          {{- end }}
-          vip_interface: {{ $vipInterface }}
-        {{- end }}
       kube-vip-cloud-provider:
         enabled: true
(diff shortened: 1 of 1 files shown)


yaocw2020:
@johnliu55tw The VIP interface configurations deleted include VLAN configurations. Does it affect the original function?

johnliu55tw:
> @johnliu55tw The VIP interface configurations deleted include VLAN configurations. Does it affect the original function?

@yaocw2020 I think it's related to https://github.com/harvester/harvester-installer/pull/175. cc @FrankYang0529

bk201:
@yaocw2020 Will this work when upgrading from 1.0.0? If the existing managedChart still contains those values.

FrankYang0529:
> > @johnliu55tw The VIP interface configurations deleted include VLAN configurations. Does it affect the original function?
> 
> @yaocw2020 I think it's related to #175. cc @FrankYang0529

Since we should redesign network architecture for https://github.com/harvester/harvester/issues/1390, I think it's okay to revert all changes in https://github.com/harvester/harvester-installer/pull/175.

yaocw2020:
> > > @johnliu55tw The VIP interface configurations deleted include VLAN configurations. Does it affect the original function?
> > 
> > 
> > @yaocw2020 I think it's related to #175. cc @FrankYang0529
> 
> Since we should redesign network architecture for [harvester/harvester#1390](https://github.com/harvester/harvester/issues/1390), I think it's okay to revert all changes in #175.

As for what issue 1390 requests, the management NIC should also be VLAN. We may not need to revert all changes but consider more about other components.

yaocw2020:
> @yaocw2020 Will this work when upgrading from 1.0.0? If the existing managedChart still contains those values.

No. Those values can not be reserved.

guangbochen (approved):
LGTM

johnliu55tw (approved):

PR #219, opened 2022-01-17
Remove duplicated Rancher images

bk201:
- Remove duplicated images. The bump-rancher script doesn't know how to remove duplicated images list in the rancher image-list file. We need to do this manually.

- Replacing `rancher/mirrored-jettech-kube-webhook-certgen` by `mirrored-ingress-nginx-kube-webhook-certgen`. See https://github.com/rancher/rancher/issues/35891#issuecomment-995989510 for more info.


bk201:
We'll need a way to determine the right versions of Rancher images.

w13915984028 (approved):
LGTM

BTW: We may consider enhance the bump script, which strip the version info from rancher-images.txt, then check if some image has duplicated/multi lines, and give warning in case existing.

johnliu55tw (approved):
Verified: Those removed images don't exist in the cluster.

guangbochen (approved):
LGTM and will track remove duplicate images as an enhancement on the https://github.com/harvester/harvester/issues/1851

PR #220, opened 2022-01-18
Use VIP:443 to join a cluster

bk201:
VIP:8443 doesn't work when VIP doesn't coexist with rancher pods.

**Related Issue**
https://github.com/harvester/harvester/issues/1830

### DIFF
--- a/package/harvester-os/files/usr/sbin/harv-update-rke2-server-url
+++ b/package/harvester-os/files/usr/sbin/harv-update-rke2-server-url
@@ -6,17 +6,24 @@ RKE2_VIP_CONFIG_FILE=/etc/rancher/rke2/config.yaml.d/90-harvester-vip.yaml
 
 update_agent_conf()
 {
-  export SERVER_URL="$(yq -e e .serverurl /oem/harvester.config | sed -E 's,^https://(.*):8443,https://\1:9345,')"
+  server_url=$(yq -e e .serverurl $HARVESTER_CONFIG_FILE)
 
-  if [ -z "$SERVER_URL" ]; then
-    echo "[Error] fail to get server URL with VIP."
+  if [ -z "$server_url" ]; then
+    echo "[Error] fail to get server URL in Harvester config."
     exit 1
   fi
 
+  if ! echo $server_url | grep -E '^https://(.*):(8443|443)'; then
+    echo "[Error] server port is not 8443 or 443."
+    exit 1
+  fi
+
+  export RKE2_SERVER_URL="$(echo $server_url | sed -E 's,^https://(.*):(8443|443),https://\1:9345,')"
+
   if [ -e $RKE2_VIP_CONFIG_FILE ]; then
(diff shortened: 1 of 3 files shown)


package/harvester-os/files/usr/sbin/harv-update-rke2-server-url
```diff
@@ -6,17 +6,24 @@ RKE2_VIP_CONFIG_FILE=/etc/rancher/rke2/config.yaml.d/90-harvester-vip.yaml
...
+  server_url=$(yq -e e .serverurl /oem/harvester.config)
 
-  if [ -z "$SERVER_URL" ]; then
-    echo "[Error] fail to get server URL with VIP."
+  if [ -z "$server_url" ]; then
+    echo "[Error] fail to get server URL in Harvester config."
     exit 1
   fi
 
+  if ! echo $server_url | grep -E '^https://(.*):(8443|443)'; then
```
gitlawr, line 16: Is it valid to use a server url without the port?

gitlawr:
Any concern for compatibility as in early discussion?

package/harvester-os/files/usr/sbin/harv-update-rke2-server-url
bk201: If the port is not included, rancherd can't get any needed info to join a cluster. Thus RKE2 will not be deployed and this script won't be run.

There are two places that involve the server URL:
- Text UI path: port is always appended to VIP.
- PXE boot path: the user needs to explicitly specify the full URL including the port number.

bk201:
> Any concern for compatibility as in early discussion?

The only thing I am aware of so far is the RKE2 server replacement is the `harv-update-rke2-server-url` script. To maintain backward compatibility, port `8843` is also considered: `(8443|443)`.

We can ignore the rancherd server URL (generated from the harvester server URL) because old nodes are already bootstrapped.

bk201:
The vagrant-pxe-harvester setting needs to adapt as well, see: https://github.com/harvester/ipxe-examples/pull/41

The CI succeeded with port `8443` because VIP is on the first node.

gitlawr (approved):
LGTM

weihanglo (commented):
Tested ISO installation with [the following steps](https://github.com/harvester/harvester/issues/1830#issuecomment-1015177805), and it behaves as expected

package/harvester-os/files/usr/sbin/harv-update-rke2-server-url
```diff
@@ -6,17 +6,24 @@ RKE2_VIP_CONFIG_FILE=/etc/rancher/rke2/config.yaml.d/90-harvester-vip.yaml
 
 update_agent_conf()
 {
-  export SERVER_URL="$(yq -e e .serverurl /oem/harvester.config | sed -E 's,^https://(.*):8443,https://\1:9345,')"
+  server_url=$(yq -e e .serverurl /oem/harvester.config)
```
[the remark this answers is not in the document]
bk201: nice catch. Fixed!

weihanglo (approved):
LGTM

PR #221, opened 2022-01-19
Download ISO to OS system disk

johnliu55tw:
For PXE installation, we need to download and mount the ISO in order to preload container images. However, as the ISO gets larger, we can no longer download it into the /tmp folder on machines with little RAM installed. We would have to download it somewhere else.

* Download the ISO onto the system disk, after we mounted it for the purpose of preloading images. The ISO is too large to fit into /tmp for machines with little RAM installed.

* Fix the issue of the installer won't error out if encountered errors while running `harv-install` script.

Signed-off-by: John Liu <john.liu@suse.com>

### DIFF
--- a/package/harvester-os/files/usr/sbin/harv-install
+++ b/package/harvester-os/files/usr/sbin/harv-install
@@ -18,19 +18,36 @@ umount_target() {
 cleanup2()
 {
     sync
+    [ -n "$_COS_INSTALL_ISO_URL" ] && umount "$ISOMNT" || true
+    [ -n "$ISOTEMP" ] && rm -f "$ISOTEMP"
     umount_target || true
     umount ${STATEDIR}
-    [ -n "_COS_INSTALL_ISO_URL" ] && umount ${ISOMNT} || true
 }
 
 cleanup()
 {
     EXIT=$?
-    if [ -n "$ISOTEMP" ]; then
-        rm -f "$ISOTEMP"
-    fi
     cleanup2 2>/dev/null || true
     return $EXIT
+}
+
+check_url()
+{
+    local url=$1
+    case $url in
+        tftp*)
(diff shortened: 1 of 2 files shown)


package/harvester-os/files/usr/sbin/harv-install
```diff
@@ -55,9 +59,9 @@ get_iso()
 {
     if [ -n "$_COS_INSTALL_ISO_URL" ]; then
         ISOMNT=$(mktemp -d -p /tmp cos.XXXXXXXX.isomnt)
-        TEMP_FILE=$(mktemp -p /tmp cos.XXXXXXXX.iso)
-        get_url ${_COS_INSTALL_ISO_URL} ${TEMP_FILE}
-        ISO_DEVICE=$(losetup --show -f $TEMP_FILE)
+        ISOTEMP=$(mktemp -p ${TARGET}/usr/local cos.XXXXXXXX.iso)
```
bk201, line 62: Since is persistent storage, it would be nice to remove the temp file when everything is done.

johnliu55tw: The file will be removed in `cleanup()` function.

weihanglo (approved):
Succeeded to install via iPXE and there is not leftover ISO file. LGTM

Jellyfrog:
Maybe more future-proof to create a tmp folder instead, then cleanup() can simply delete the whole folder.

johnliu55tw:
> Maybe more future-proof to create a tmp folder instead, then cleanup() can simply delete the whole folder.

Totally agree it's more future-proof, but right now the ISO is the only thing we need to download onto the disk, so I'd prefer to implement this "on-disk tmp folder" until we need to put more files here.

bk201 (approved):
LGTM.

weihanglo (approved):

PR #223, opened 2022-01-28
Add env var 'LOGFILE' to configure log file path

johnliu55tw:


bk201 (approved):
We might eventually move the log to stdout since we run the installer as a systemd service.

guangbochen (approved):
LGTM

PR #224, opened 2022-01-28
Add Installer UI for selecting the default data disk

johnliu55tw:
# Description

- Add Data Disk page. It will only show if the machine has two or more disks.

- Refactor: createVerticalLocator now takes interface `Element`.

- Refactor: add `ShowElement()` to Console to show the element using element's name. We could leverage this method to focus (move the cursor to) the element.

- Fix: Panel.SetContent() should only set its displaying content and ignore other View-related manipulation if View has not been created yet.

- Fix: Select.SetData() should not error out if View is not created yet. It should update its `Value` field and skip the View manipulation.

# Test plan
## Case 1: Machines with two or more disks can assign the data disk
1. Create a VM with **two or more** disks. **All of them should be larger than 140G (recommended size)**
2. Boot the installation ISO
3. Choose the first disk, say `sda`, as the installation target (This is the OS disk)
4. Verify that installer shows the page "Choose disk for storing VM data"

    image

5. Choose the second disk, say `sdb`, as the data disk
6. Proceed with the installation
7. After the installation is finished, wait for the system to reboot and verify the following:
    - The data disk `/dev/sdb` is mounted to `/var/lib/harvester/defaultdisk/` as the default disk. Create a VM and verify Longhorn is storing data into `/var/lib/harvester/defaultdisk/`. You can create a huge file in the VM and see the space of `/dev/sdb` is getting consumed.
    - `/var/lib/longhorn` still exists, and Longhorn binaries are stored here.
        ```
        $ ls -l /var/lib/longhorn/engine-binaries
        total 4
        drwxr-xr-x 2 root root 4096 Jan 19 08:14 longhornio-longhorn-engine-v1.2.3
        ```

## Case 2: Machines with two or more disks can choose the same disk as OS and data disk
1. Repeat the step 1 ~ 4 of Case 1
2. Choose the option `Use the installation disk (...)` as the data disk

    image

3. Proceed with the installation
4. After the installation is finished, wait for the system to reboot and verify the following:
    - A partition labeled `HARV_LH_DEFAULT` is created on the OS disk `/dev/sda` and mounted to `/var/lib/harvester/defaultdisk/`.
    - `/var/lib/longhorn` still exists, and Longhorn binaries are stored here.
        ```
        $ ls -l /var/lib/longhorn/engine-binaries
        total 4
        drwxr-xr-x 2 root root 4096 Jan 19 08:14 longhornio-longhorn-engine-v1.2.3
        ```

## Case 3: Machines with only one disk cannot choose data disk
1. Create a VM with only one disk, which should be greater than 140G
2. Boot the installation ISO
3. Choose the only existed disk, say `sda`, as the installation target (This is the OS disk)
4. Verify that installer **doest not** show the page "Choose disk for storing VM data". Instead, installer would jump to "Configure network connection" directly.
5. Proceed with the installation
6. Follow step 4 in Case 2 to verify the installation.

## Case 4: Test the validator
Both OS disk and data disk share the same validation criteria: **Soft limit = 140G and hard limit = 60G**:
- If the disk is smaller than the soft limit (140G), installer will show a warning message, but users can still proceed after pressing Enter again
- If the disk is smaller than the hard limit (60G), installer will block the installation. Users must choose another disk that fulfills the criteria in order to proceed.
- If the data disk is provided by a remote config or it's installed via PXE, only the hard limit would be established, i.e. installer would stop the installation if either of the disks is smaller than 60G.

## Case 5: Assigning data disk using `install.data_disk` Harvester config
1. Prepare a Harvester config including the `data_disk` config:
    ```
    ...
    install:
      data_disk: /dev/sdb  # Symbolic-link path is also supported, e.g. /dev/disk/by-id/scsi-1IET_00010001
      ...
    ...
    ```
2. Create a PXE installation environment with the above config
3. Start the PXE installation on a machine/VM that has two or more disks, of those disks must have one that matches that path (`/dev/sdb` in the above example config)
4. Follow step 7 in Case 1 to verify the installation.

@guangbochen Do you think we should add a size validation for the DataDisk?


guangbochen:
@johnliu55tw I think we can add a proper pop-up warning message if the data disk size is smaller than the minimum recommended size, but users can still continue the installation if needed.

pkg/console/util.go
```diff
@@ -561,8 +557,8 @@ func validateDiskSizeSoft(devPath string) error {
 	if err != nil {
 		return err
 	}
-	if diskSizeBytes>>30 < softMinDiskSizeGiB {
-		return fmt.Errorf("Disk size smaller than recommended size %dGB", softMinDiskSizeGiB)
+	if diskSizeBytes>>30 < config.SoftMinDiskSizeGiB {
+		return fmt.Errorf("Disk size smaller than recommended size %dGB", config.SoftMinDiskSizeGiB)
```
bk201, line 561:
```suggestion
		return fmt.Errorf("Disk size is smaller than the recommended size: %dGB", config.SoftMinDiskSizeGiB)
```
nit

bk201 (approved):
LGTM!

guangbochen (approved):
LGTM

PR #225, opened 2022-02-02
Ability to process kernel arguments containing harvester-mgmt NetworkInterface details

ibrokethecloud:
The PR is in relation to feature request: https://github.com/harvester/harvester/issues/1889

With this minor tweak to the harvester-installer, the mass PXE booting of nodes is easier as the ${net0/mac} PXE variable can be passed to kernel arguments to allow harvester-installer to use hwAddr based address lookup to find the correct interface.

This makes it easy to PXE boot a full multi node harvester cluster quickly using something like terraform. A sample of the same using a custom squashfs build is here: https://github.com/ibrokethecloud/harvester-equinix-terraform-sample

### DIFF
--- a/pkg/config/read.go
+++ b/pkg/config/read.go
@@ -1,7 +1,6 @@
 package config
 
 import (
-	"encoding/json"
 	"fmt"
 	"strings"
 
@@ -18,10 +17,6 @@ const (
 func ReadConfig() (HarvesterConfig, error) {
 	result := NewHarvesterConfig()
 	data, err := util.ReadCmdline(kernelParamPrefix)
-	if err != nil {
-		return *result, err
-	}
-	err = ToNetworkInterface(data)
 	if err != nil {
 		return *result, err
 	}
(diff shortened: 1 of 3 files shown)


pkg/config/read.go
```diff
@@ -20,6 +21,10 @@ func ReadConfig() (HarvesterConfig, error) {
 	if err != nil {
 		return *result, err
 	}
+	err = ToNetworkInterface(data)
```
bk201, line 24: Since this hack is kernel parameter-specific, would it be better to put the logic in `parseCmdLine` function?
https://github.com/harvester/harvester-installer/blob/e0c7293a149bd385399b41766ed2fe7fc2a46006/pkg/util/cmdline.go#L13

ibrokethecloud: thanks for the feedback @bk201. Please let me know if it looks better now.

pkg/config/read.go
```diff
@@ -46,3 +51,33 @@ func mapToEnv(prefix string, data map[string]interface{}) []string {
 	}
 	return result
 }
+
+func ToNetworkInterface(data map[string]interface{}) error {
+	if installInterface, ok := data["install"]; ok {
+		if networksInterface, ok := installInterface.(map[string]interface{})["networks"]; ok {
+			if mgmtInterface, ok := networksInterface.(map[string]interface{})["harvester-mgmt"]; ok {
+				if interfaces, ok := mgmtInterface.(map[string]interface{})["interfaces"]; ok {
```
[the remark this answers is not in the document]
ibrokethecloud: Done. Moved to using values for handling the data translation.

johnliu55tw (approved):
LGTM

bk201 (approved):

PR #226, opened 2022-02-08
validate system settings to avoid unexpected settings

tjjh89017:
related to https://github.com/harvester/harvester/issues/1739

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -35,24 +35,33 @@ const (
 	NormalCosPartSizeGiB = 50
 )
 
-// refer: https://docs.harvesterhci.io/latest/settings/settings/
+// refer: https://github.com/harvester/harvester/blob/master/pkg/settings/settings.go
 func GetSystemSettingsAllowList() []string {
 	return []string{
 		"additional-ca",
-		"backup-target",
+		"api-ui-version",
 		"cluster-registeration-url",
-		"http-proxy",
-		"log-level",
-		"overcommit-config",
 		"server-version",
-		"ssl-certificates",
-		"ssl-parameters",
 		"ui-index",
+		"ui-path",
 		"ui-source",
+		"voluem-snapshot-class",
+		"backup-target",
+		"upgradable-versions",
 		"upgrade-checker-enabled",
 		"upgrade-checker-url",
-		"auto-disk-provision-paths",
+		"log-level",
+		"ssl-certificates",
+		"ssl-parameters",
+		"support-bundle-image",
+		"support-bundle-namespaces",
+		"support-bundle-timeout",
+		"default-sotrage-class",
+		"http-proxy",
 		"vm-force-reset-policy",
+		"overcommit-config",
+		"vip-pools",
+		"auto-disk-provision-paths",
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/console/validator.go
```diff
@@ -274,6 +276,27 @@ func checkToken(token string) error {
...
 
+func checkSystemSettings(systemSettings map[string]string) error {
+	if systemSettings == nil {
+		return nil
+	}
+
+	allowList := config.GetSystemSettingsAllowList()
+	for systemSetting, _ := range systemSettings {
+		isValid := false
+		for _, allowSystemSetting := range allowList {
```
FrankYang0529, line 287: NIT: Can use a map to avoid looping `allowList` for each setting.

```
func GetAllowedSystemSettings() map[string]interface{} {
	return map[string]interface{}{
		"additional-ca":             nil,
        ...
    }
}

func checkSystemSettings(systemSettings map[string]string) error {
	if systemSettings == nil {
		return nil
	}

	allowedSystemSettings := GetAllowedSystemSettings()
	for systemSetting, _ := range systemSettings {
		if _, ok := allowedSystemSettings[systemSetting]; !ok {
			return errors.Errorf(ErrMsgSystemSettingsUnknown, systemSetting)
		}
	}
	return nil
}
```

FrankYang0529 (approved):
LGTM

pkg/console/validator.go
tjjh89017: I found this benchmark
`map` take lots of time of it.
so i chose linear search and keep readability
https://stackoverflow.com/questions/15323767/does-go-have-if-x-in-construct-similar-to-python

pkg/config/config.go
```diff
@@ -35,6 +35,27 @@ const (
 	NormalCosPartSizeGiB = 50
 )
 
+// refer: https://docs.harvesterhci.io/latest/settings/settings/
+func GetSystemSettingsAllowList() []string {
+	return []string{
```
johnliu55tw, line 40: The complete list of settings is actually here: https://github.com/harvester/harvester/blob/f303b27ef7479f5727a1d6674b2d543f307575d0/pkg/settings/settings.go#L22-L45. Although some of them are not user-facing, IMO we should allow all of them to be configurable using this `SystemSettings`.

pkg/console/validator.go
```diff
@@ -274,6 +276,27 @@ func checkToken(token string) error {
 	return nil
 }
 
+func checkSystemSettings(systemSettings map[string]string) error {
+	if systemSettings == nil {
+		return nil
+	}
+
+	allowList := config.GetSystemSettingsAllowList()
+	for systemSetting, _ := range systemSettings {
```
johnliu55tw, line 285: NIT: You don't necessarily need the `_` here if key is all you need:
```suggestion
	for systemSetting := range systemSettings {
```

pkg/config/config.go
tjjh89017: that's also my question.
IMO, we should keep allow list only including we already fully describe in the docs.
and refuse some non user-facing settings.

guangbochen (approved):
LGTM

pkg/config/config.go
tjjh89017: @guangbochen 
after discussing with @johnliu55tw , we decide to copy allow list from all settings including non user-facing.
plz wait for new commit before merging 
thanks a lot

johnliu55tw (approved):
LGTM!

PR #227, opened 2022-02-08
add ClusterNetworks config in installer

tjjh89017:
cluster network only allow one type of entry in same cluster.
also check dup name or not

address https://github.com/harvester/harvester/issues/1739

## Test method

1. Install a node of new cluster with ISO or PXE and put the config below into `config.yaml` and provide it to installer 
```yaml
cluster_networks:
  vlan:
    enable: true
    description: "some description about this cluster network"
    config:
      defaultPhysicalNIC: ens3
```

2. verify no error occurs and finish the installation
3. verify `ClusterNetwork` resource named `vlan` is created.
```bash
kubectl get clusternetworks.network.harvesterhci.io vlan -o yaml
```

4. verify the fields of `ClusterNetwork` resource named `vlan`  are correct.
```bash
kubectl get clusternetworks.network.harvesterhci.io vlan -o yaml
```

Result should contain `name`, `enable`, `config`, `defaultPhysicalNIC` with pre-defined value.
```
apiVersion: network.harvesterhci.io/v1beta1
config:
  defaultPhysicalNIC: ens3
description: "some description about this cluster network"
enable: true
kind: ClusterNetwork
metadata:
  creationTimestamp: "2022-01-25T07:18:21Z"
  finalizers:
  - wrangler.cattle.io/harvester-clusternetwork-controller
  generation: 6
  name: vlan
  resourceVersion: "252459"
  uid: 9d2c473e-f11a-474b-9189-8889df1076c6

```

5. verify UI in `Settings` page VLAN entry. Click into vlan option setup page and verify that show `ClusterNetwork: vlan`. Please refer to the picture and check the resource type in the red frame
image

6. verify UI of VLAN setup page in Settings page. 
    - verify `Enabled`, `Disabled` button.
    - verify it shows correct interface name in `Default Network Interface`
    - verify changing value in `Default Network Interface` and apply config
        - after setting new value for `Default Network Interface`, verify step 4 again to confirm that resource is updated

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -117,6 +117,11 @@ type OS struct {
 	Labels         map[string]string `json:"labels,omitempty"`
 }
 
+type ClusterNetwork struct {
+	Description string            `json:"description,omitempty"`
+	Config      map[string]string `json:"config,omitempty"`
+}
+
 type HarvesterConfig struct {
 	ServerURL string `json:"serverUrl,omitempty"`
 	Token     string `json:"token,omitempty"`
@@ -127,6 +132,7 @@ type HarvesterConfig struct {
 	HarvesterChartVersion  string            `json:"harvesterChartVersion,omitempty"`
 	MonitoringChartVersion string            `json:"monitoringChartVersion,omitempty"`
 	SystemSettings         map[string]string `json:"systemSettings,omitempty"`
+	ClusterNetworks        map[string]ClusterNetwork `json:"clusterNetworks,omitempty"`
 }
 
 func NewHarvesterConfig() *HarvesterConfig {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -118,6 +118,8 @@ type OS struct {
 }
 
 type ClusterNetwork struct {
+	// by default: false
+	Enable      bool              `json:"enable,omitempty"`
 	Description string            `json:"description,omitempty"`
 	Config      map[string]string `json:"config,omitempty"`
 }
--- a/pkg/config/templates/rancherd-21-harvester-clusternetworks.yaml
+++ b/pkg/config/templates/rancherd-21-harvester-clusternetworks.yaml
@@ -1,0 +1,20 @@
+bootstrapResources: {{ if not .ClusterNetworks -}} [] {{- else }}
+{{- range $name, $network := .ClusterNetworks }}
+- apiVersion: harvesterhci.io/v1beta1
+  kind: ClusterNetwork
+  enable: true
+  metadata:
+    name: {{ printf "%q" $name }}
+{{- with $network }}
+{{- if .Description }}
+  description: {{ printf "%q" .Description }}
+{{- end }}
+{{- if .Config }}
+  config:
+{{- range $name, $value := .Config }}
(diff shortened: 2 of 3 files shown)


pkg/config/templates/rancherd-21-harvester-clusternetworks.yaml
```diff
@@ -0,0 +1,20 @@
+bootstrapResources: {{ if not .ClusterNetworks -}} [] {{- else }}
+{{- range $name, $network := .ClusterNetworks }}
+- apiVersion: harvesterhci.io/v1beta1
+  kind: ClusterNetwork
+  enable: true
```
yaocw2020, line 5: Why is it always true?

pkg/config/config.go
```diff
@@ -127,6 +132,7 @@ type HarvesterConfig struct {
 	HarvesterChartVersion  string            `json:"harvesterChartVersion,omitempty"`
 	MonitoringChartVersion string            `json:"monitoringChartVersion,omitempty"`
 	SystemSettings         map[string]string `json:"systemSettings,omitempty"`
+	ClusterNetworks        map[string]ClusterNetwork `json:"clusterNetworks,omitempty"`
```
yaocw2020, line 135: I guess there is a golint issue.

pkg/config/templates/rancherd-21-harvester-clusternetworks.yaml
yaocw2020, line 5: Why is it always true?

pkg/config/config.go
tjjh89017: golint only told me that need comment for that entry.

pkg/config/templates/rancherd-21-harvester-clusternetworks.yaml
tjjh89017: let me add an option for "enable"

i thought if they will put config for installer, and they will only want to enable it right away

yaocw2020: We can set it as `true` if the VLAN ID and default physical nic are not empty. What do you think?

tjjh89017: i'm thinking not to detect "config" field for all kind of clusternetworks.
because that will need to setup lots of validate function for all kinds in installer if we have further network type to support. it will cause we have two place that need to be fixed if we have some change (installer and network-controller)

Maybe we can setup "false" as default and put some words in docs for notice?

yaocw2020: Agree

tjjh89017:
add some example and config reference into docs PR.
for notice, default value and valid config for `config` map

https://github.com/harvester/docs/pull/114/commits/000d2a5f72da4106942a37be2ffa5be1bde1820f

guangbochen (approved):
LGTM, can we please add proper validation steps to the PR description.

yaocw2020 (approved):
LGTM

guangbochen (approved):
LGTM, also should add both manual and e2e tests of this feature change, thanks.

PR #230, opened 2022-02-22
Add VLAN ID panel

johnliu55tw:


pkg/console/network.go
```diff
@@ -108,29 +107,3 @@ func getNICState(name string) int {
 	}
 	return NICStateUP
 }
-
-type networkHardwareInfo struct {
```
johnliu55tw, line 112: Remove unused codes.

johnliu55tw:
Closed due to that VLAN ID implementation might be changed in the future.

PR #231, opened 2022-02-22
change the apiVersoin to correct path

tjjh89017:
assign incorrect apiVersion
already correct it


guangbochen (approved):
LGTM

PR #234, opened 2022-02-25
Support MTU config

johnliu55tw:
- Add config `MTU` to Network configuration
- Support merging the "Networks" field of HarvesterConfig (because [Mergo](https://github.com/imdario/mergo#usage) does not support merging struct inside maps)

1. Boot the machine via Harvester ISO
2. Supply a remote config with the following YAML:
    ```
    install:
      networks:
        harvester-mgmt:
          mtu: 1400
    ```
    This will set the `harvester-mgmt` interface to have MTU `1400`. Change this value as you wish (default is `1500`).

3. Proceed with the installation
4. SSH into the Harvester node and use the command `ip link show <interface>` to verify:
    - `harvester-mgmt` interface has the MTU `1400` (or whatever value you provided)
    - `enp1s0` (The slave interface of `harvester-mgmt`) also has the MTU `1400` (or whatever value you provided)
5. Reboot the node and verify that the MTU setting is preserved over system reboots.

## Related issues
- https://github.com/harvester/harvester/issues/1934

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -74,6 +74,7 @@ type Network struct {
 	DefaultRoute bool               `json:"-"`
 	BondOptions  map[string]string  `json:"bondOptions,omitempty"`
 	VlanID       int                `json:"vlanId,omitempty"`
+	MTU          int                `json:"mtu,omitempty"`
 }
 
 type HTTPBasicAuth struct {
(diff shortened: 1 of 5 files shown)


weihanglo (approved):
Tested and looks reasonable to me!

image

pkg/config/config.go
```diff
@@ -229,6 +230,29 @@ func (c *HarvesterConfig) GetKubeletArgs() ([]string, error) {
 	return []string{}, nil
 }
 
+func (c *HarvesterConfig) Merge(other HarvesterConfig) error {
+	// We need to manually merge the "Networks" field (map[string]Network) beacuse Mergo won't
```
bk201, line 234:
```suggestion
	// We need to manually merge the "Networks" field (map[string]Network) because Mergo won't
```
nit: My editor found this typo.

johnliu55tw: Fixed, thanks!

PR #241, opened 2022-03-08
Allow to installing on 64 GB USB Flash Disks

mehmetaydogduu:
HardMinDiskSizeGiB value decreased to 56

Because usable capacity is 57 GB on most devices.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -30,7 +30,7 @@ const (
 
 const (
 	SoftMinDiskSizeGiB   = 140
-	HardMinDiskSizeGiB   = 60
+	HardMinDiskSizeGiB   = 56
 	MinCosPartSizeGiB    = 25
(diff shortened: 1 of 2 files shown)


bk201:
cc @johnliu55tw Does decreasing the hard limit down potentially break anything? Or we can treat any disk size < soft limit as development or trying purpose.

Hi, @mehmetaydogduu Thanks for your contribution, it would be nice to [create an issue](https://github.com/harvester/harvester/issues/new/choose) and link this PR to that issue to ease the tracking. Thanks!

johnliu55tw:
@bk201 With disk size less than current hard-limit 60G might cause the monitoring system to stop working.

bk201:
The disk size is crucial for operating the cluster, please see https://docs.harvesterhci.io/v1.3/install/requirements#hardware-requirements for hardware requirements. Thanks for the contribution.

PR #243, opened 2022-03-10
Use harvester code info to patch harvester chart version and harvester image tag

futuretea:
We want to build iso for [harvester v1.0 branch](https://github.com/harvester/harvester/tree/v1.0)

**Problem:**

1. If not a harvester-installer release, the default version of harvester-installer is always `master`. 
each iso build on different branches has the same name.
1. If not a harvester release, the default image tag of harvester chart is always `master-head`. 
1. If not a harvester release, the default chart version of harvester chart is always `0.0.0-dev`. 
1. If not a harvester release, the default chart appVersion of harvester chart is always `v0.0.0-dev`. 

**Solution:**

1. In harvester, build chart version, image tag in `script/version`
1. In harvester-installer, try to use `DRONE_BRANCH` as the iso version
1. In harvester-installer, Add a script `patch-harvester` to patch harvester chart version and harvester image tag uses the harvester code info.

**Related issues**
https://github.com/harvester/harvester/issues/1817

**Dependent PR**
https://github.com/harvester/harvester/pull/1990

**Test plan**

### DIFF
--- a/scripts/version
+++ b/scripts/version
@@ -6,8 +6,8 @@ if [ -n "$(git status --porcelain --untracked-files=no)" ]; then
 fi
 
 COMMIT=$(git rev-parse --short HEAD)
-COMMIT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
-VERSION=${DRONE_TAG:-${COMMIT_BRANCH}}
+VERSION=${DRONE_TAG:-${DRONE_BRANCH}}
+VERSION=${VERSION:-"master"}
 
 if [[ -n "$DIRTY" ]]; then
     VERSION="${COMMIT}${DIRTY}"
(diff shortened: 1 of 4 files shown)


weihanglo (approved):
LGTM

bk201 (approved):
LGTM

scripts/version
```diff
@@ -6,7 +6,8 @@ if [ -n "$(git status --porcelain --untracked-files=no)" ]; then
 fi
 
 COMMIT=$(git rev-parse --short HEAD)
-VERSION=${DRONE_TAG:-master}
+VERSION=${DRONE_TAG:-${DRONE_BRANCH}}
+VERSION=${VERSION:-"master"}
```
guangbochen, line 10: is this redundant as it will always have a `DRONE_BRANCH`?

guangbochen (approved):

PR #251, opened 2022-03-22
Fix Ready status

futuretea:
Signed-off-by: futuretea <Hang.Yu@suse.com>

- check the Ready status by
```bash
curl -fk https://managementURL/version
```

- remove unused logics: `getFirstReadyMasterIP`

**Related issues**
https://github.com/harvester/harvester/issues/2022

**Test plan**

- Install harvester on a machine
- wait for it to show the 1st ready status
- try to access harvester via the URL
- harvester should only show ready status once it is fully ready.

### DIFF
--- a/pkg/console/dashboard_panels.go
+++ b/pkg/console/dashboard_panels.go
@@ -41,8 +41,9 @@ const (
 )
 
 type state struct {
-	installed bool
-	firstHost bool
+	installed     bool
+	firstHost     bool
+	managementURL string
 }
 
 var (
@@ -216,10 +217,7 @@ func doSyncManagementURL(g *gocui.Gui) {
 	managementIP := getVIP()
 	if managementIP != "" {
 		managementURL = fmt.Sprintf("https://%s", managementIP)
-	} else {
-		if managementIP = getFirstReadyMasterIP(); managementIP != "" {
-			managementURL = fmt.Sprintf("https://%s:%s", managementIP, harvesterNodePort)
-		}
+		current.managementURL = managementURL
 	}
 
 	g.Update(func(g *gocui.Gui) error {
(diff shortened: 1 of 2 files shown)


pkg/console/dashboard_panels.go
```diff
@@ -330,6 +313,17 @@ func chartIsInstalled() bool {
...
 
+func isAPIReady(managementURL, path string) bool {
+	if !strings.HasPrefix(current.managementURL, "https://") {
+		return false
+	}
+	command := fmt.Sprintf(`curl -k %s%s`, managementURL, path)
+	cmd := exec.Command("/bin/sh", "-c", command)
+	cmd.Env = os.Environ()
+	_, err := cmd.CombinedOutput()
+	return err == nil
```
weihanglo, line 324: Do we need to check HTTP status here? Or a response is sufficient?

futuretea: changed to 'curl -fk'

pkg/console/dashboard_panels.go
```diff
@@ -330,6 +313,17 @@ func chartIsInstalled() bool {
 	return processed >= 1
 }
 
+func isAPIReady(managementURL, path string) bool {
+	if !strings.HasPrefix(current.managementURL, "https://") {
+		return false
+	}
+	command := fmt.Sprintf(`curl -fk %s%s`, managementURL, path)
+	cmd := exec.Command("/bin/sh", "-c", command)
```
yaocw2020, line 321: What about using the net/http package instead of the hardcode command?

futuretea: curl is simple and strong

weihanglo (approved):
LGTM

guangbochen (approved):
LGTM

PR #253, opened 2022-03-23
fixed some typos in the validate code

joeyloman:
After reinstalling Harvester with version 1.0.1-rc1 I ran into validation check errors during installation which were caused by these typo's. They now match again with the settings in the Harvester code base.


bk201 (approved):
Nice spotting! Thanks for the fix.

cc @tjjh89017 We might need to figure out a more sustainable way to track harvester settings.

tjjh89017 (approved):
LGTM

PR #255, opened 2022-03-23
Generate hybrid ISO

bk201:
Bump `luet-makeiso` to 0.4.0 to generate a valid hybrid ISO.
Hopefully, it solves the issue that some users can't boot a USB installation media.

For more information, see https://github.com/rancher-sandbox/cOS-toolkit/issues/1148

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -1,5 +1,5 @@
 FROM quay.io/costoolkit/releases-green:luet-toolchain-0.21.2 as luet
-FROM quay.io/costoolkit/releases-green:luet-makeiso-toolchain-0.3.8-16 as makeiso
+FROM quay.io/costoolkit/releases-green:luet-makeiso-toolchain-0.4.0 as makeiso
 
 FROM registry.suse.com/bci/bci-base:15.3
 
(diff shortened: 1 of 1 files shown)


FrankYang0529 (approved):
LGTM

Dockerfile.dapper
```diff
@@ -17,7 +17,7 @@ ARG DAPPER_HOST_ARCH
 ENV ARCH $DAPPER_HOST_ARCH
 
 RUN zypper -n rm container-suseconnect && \
-    zypper -n install go1.16 git curl docker gzip tar wget zstd squashfs xorriso awk
+    zypper -n install go1.16 git curl docker gzip tar wget zstd squashfs xorriso awk mtools dosfstools
```
weihanglo, line 20: Can you explain why these two tools are required?
I feel like we need to record the reason when adding tools otherwise it will become hard to remove afterwards

weihanglo (approved):
LGTM

Dockerfile.dapper
bk201: Sure! Rebased and add the reason in the comment.

weihanglo (approved):

PR #257, opened 2022-03-30
Bump rancher to v2.6.4-harvester1

bk201:
Bump rancher to `v2.6.4-harvester1`.

Known issues
- https://github.com/rancher/rancher/issues/37113
- https://github.com/rancher/rancher/issues/37114
  - **Need to test the PR with Internet connectivity**


guangbochen (approved):
LGTM. cc @TachunLin @lanfon72 because of the above Rancher bugs upgrade with air-gap env will not be supported, we can verify it after the upstream fix, thanks.

bk201:
@guangbochen Not only upgrade, but it's also about the bootstrap of new clusters.
Internet is required.

weihanglo (approved):
LGTM

- ✅ Can join cluster
- ✅ Can run VMs in the cluster

PR #259, opened 2022-03-30
Add workaround for https://github.com/rancher/rancher/issues/37113

bk201:
- We need to temporarily hard-code the generated rancher-webhook versions in harvester-release.yaml.
- Also specify the minimum rancher-webhook chart versions via env variables when creating applying the rancher chart.

To test upgrade, the PR is needed: https://github.com/harvester/harvester/pull/2097


johnliu55tw (approved):
Upgraded from v1.0.0 and rancher-webhook chart is being upgraded to `1.0.3+up0.2.5`.

guangbochen (approved):
LGTM

PR #261, opened 2022-03-31
Bump Rancher to v2.6.4-harvester2

guangbochen:
**Description:**
Bump Rancher to `v2.6.4-harvester2`.


weihanglo (approved):
Since it is like "two different tags referring to the same thing", I lean towards merging it for now.

bk201 (approved):

PR #265, opened 2022-04-06
Check whether ntpServers input is empty

FrankYang0529:
### Test steps

1. Install Harvester by ISO.
2. In the NTP page, remove default input.
3. Press Enter and we can see a warning message "Empty NTP Server is not recommended. Press Enter to continue.".
4. Press Enter again and we can go to the next page.
5. Press Esc and we can go back to the NTP page.
6. Press Enter and we can see the warning message.
7. Input a valid NTP server.
8. Press Enter. There should not have any error message and we can go to next page.

related issue:
https://github.com/harvester/harvester/issues/2119

### DIFF
--- a/pkg/console/install_panels.go
+++ b/pkg/console/install_panels.go
@@ -1759,10 +1759,8 @@ func addNTPServersPanel(c *Console) error {
 				return err
 			}
 
-			// Go to next page when user give an empty input or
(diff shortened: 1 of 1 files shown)


bk201 (approved):
LGTM

guangbochen (approved):
LGTM

PR #267, opened 2022-04-06
Add boot test document

tlehman:
In the course of trying to resolve [Issue 2023 on the Harvester repo](https://github.com/harvester/harvester/issues/2023), I found a reliable way to test the two different boot modes of the build artifact of this repo. Currently the legacy BIOS boot mode __does not work__.

The reason there was some confusion about this was that newer motherboards with UEFI and the CSM (Compatibility Support Module) can technically boot the installer, but older motherboards without any UEFI option cannot.

These two tests I documented here reliably reproduce this issue, and they should be part of the installer repo.

### DIFF
--- a/tests/boot.md
+++ b/tests/boot.md
@@ -1,0 +1,30 @@
+## Testing the ISO image
+
+This repo uses [luet-makeiso](https://github.com/Luet-lab/luet-makeiso) to produce a hybrid ISO image.
+A hybrid ISO image should be able to boot on older computers that boot with legacy BIOS as well as 
+newer computers that boot with UEFI BIOS.
+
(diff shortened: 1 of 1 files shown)


noahgildersleeve:
@tlehman All manual tests go into the harvester/tests repo in the tests/docs/content/manual/ folder. It’s documented in the README.md. All new tests go into there in the tests/docs/content/manual/_incoming/ . We move the tests out of incoming when we publish a new release so that we can keep track of new tests per release. We use a static site builder called hugo that creates the page. You can look at one of the other tests for the format. It’s mostly Markdown except for the header which is a little different.

PR #268, opened 2022-04-07
Downgrade RKE2 to v1.21.11+rke2r1

bk201:
Depends on https://github.com/harvester/os2/pull/35 (Note I include a temporary image in commits)

**Related issue**: https://github.com/harvester/harvester/issues/2112

### DIFF
--- a/scripts/package-harvester-os
+++ b/scripts/package-harvester-os
@@ -14,7 +14,7 @@ source ${SCRIPTS_DIR}/version-rancher
 source ${SCRIPTS_DIR}/version-harvester ${TOP_DIR}/../harvester
 source ${SCRIPTS_DIR}/version-monitoring
 
-BASE_OS_IMAGE="rancher/harvester-os:20220406"
+BASE_OS_IMAGE="rancher/harvester-os:20220407"
 HARVESTER_OS_IMAGE=rancher/harvester-os:$VERSION
 
 cd ${PACKAGE_HARVESTER_OS_DIR}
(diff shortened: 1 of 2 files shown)


guangbochen (approved):
LGTM and can we update the base OS image, thanks. 
Also loop @TachunLin  @lanfon72 because of the [ETCD data inconsistency issue](https://groups.google.com/a/kubernetes.io/g/dev/c/B7gJs88XtQc/m/rSgNOzV2BwAJ?pli=1) we need to downgrade the RKE2 version to v1.21.11 which uses the ETCD 3.4.x, which means we will need to run the upgrade regression test probably with the `v1.0-head.ISO` image later, thanks.

scripts/package-harvester-os
```diff
@@ -14,7 +14,7 @@ source ${SCRIPTS_DIR}/version-rancher
 source ${SCRIPTS_DIR}/version-harvester ${TOP_DIR}/../harvester
 source ${SCRIPTS_DIR}/version-monitoring
 
-BASE_OS_IMAGE="rancher/harvester-os:20220406"
+BASE_OS_IMAGE="bk201z/harvester-os:20220406-60b74cd"
```
weihanglo, line 17: Build succeeded and we can now replace it.

https://drone-publish.rancher.io/harvester/os2/46/1/2


```suggestion
BASE_OS_IMAGE="bk201z/harvester-os:20220407-ae38642"
```

weihanglo (approved):
LGTM

scripts/package-harvester-os
weihanglo: Need a tag on harvester/os2 repo

bk201: Sure, The image is published and used.

PR #270, opened 2022-04-08
Run "rke2-killall.sh" before shutdown

johnliu55tw:
Create a `rke2-shutdown.service` to run script `rke2-killall.sh` to stop all the running pods (containers) before system shutdown or reboot. One of the benefits is that system shutdown and reboot would be much faster.

The key component is the `Before=umount.target` so that the service will run `rke2-killall.sh` before systems started to unmount everything, including those Pods' volumes.

## Test plan
1. Deploy Harvester (one node would be sufficient). Wait for it to be Ready
2. Switch to Shell by pressing `F12`
3. Initiate the **graceful** shutdown process, e.g. by running the command `poweroff`
4. The system should be shut down **fairly quickly compared to the old version of Harvester**. One way to tell the difference is that in **an older Harvester**, the terminal would show the following message while shutting down:
image
and the shutdown process would be blocked for around 1 minute. **With this PR, those messages will not show at all.**
5. You can further test `reboot` command or operation such as *Press power button* to initiate the graceful shutdown/reboot process, and they all shouldn't be affected by the message shown above.

## TODOs
- [x] Upgrade: Validated that when upgrading from an older version of Harvester, the `rke2-shutdown.service` will be installed and enabled.

## Related
- https://github.com/harvester/harvester/issues/1876
- https://github.com/rancher/rke2/issues/2411#issue-1117223202 The `rke2-shutdown` service is inspired by the `systemd-unit.txt` provided by the author of the issue

### DIFF
--- a/package/harvester-os/files/etc/systemd/system/rke2-shutdown.service
+++ b/package/harvester-os/files/etc/systemd/system/rke2-shutdown.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=Kill RKE2 Containers for clean unmount
+DefaultDependencies=no
+RefuseManualStart=yes
+Before=reboot.target halt.target shutdown.target poweroff.target umount.target final.target
+
+[Install]
+RequiredBy=reboot.target halt.target shutdown.target poweroff.target umount.target final.target
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=-/usr/local/bin/rke2-killall.sh
--- a/package/harvester-os/files/etc/systemd/system/rke2-shutdown.service
+++ b/package/harvester-os/files/etc/systemd/system/rke2-shutdown.service
@@ -10,4 +10,4 @@ RequiredBy=reboot.target halt.target shutdown.target poweroff.target umount.targ
 [Service]
 Type=oneshot
 RemainAfterExit=yes
-ExecStart=-/usr/local/bin/rke2-killall.sh
+ExecStart=/usr/sbin/rke2-kill-containers.sh
--- a/package/harvester-os/files/system/oem/91_rke2-shutdown.yaml
+++ b/package/harvester-os/files/system/oem/91_rke2-shutdown.yaml
@@ -1,0 +1,7 @@
+name: "Enable RKE2 shutdown service"
+stages:
+  initramfs:
+    - name: "enable rke2-shutdown.service"
+      if: 'grep -q root=LABEL=COS_ACTIVE /proc/cmdline && [ -n "$(blkid -L COS_ACTIVE)" ]'
+      commands:
+        - systemctl enable rke2-shutdown.service
--- a/package/harvester-os/files/usr/sbin/rke2-kill-containers.sh
+++ b/package/harvester-os/files/usr/sbin/rke2-kill-containers.sh
@@ -1,0 +1,50 @@
+#!/bin/bash -x
+
+# make sure we run as root
+if [ ! $(id -u) -eq 0 ]; then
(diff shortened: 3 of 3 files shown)


tjjh89017 (commented):
LGTM

tjjh89017 (approved):

package/harvester-os/files/etc/systemd/system/rke2-shutdown.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Kill RKE2 Containers for clean unmount
+Documentation=man:systemd-halt.service(8)
```
bk201, line 3: nit: not quite sure the doc here is valid.

package/harvester-os/files/etc/systemd/system/rke2-shutdown.service
```diff
@@ -0,0 +1,14 @@
...
+DefaultDependencies=no
+RefuseManualStart=yes
+Before=reboot.target halt.target shutdown.target poweroff.target umount.target final.target
+
+[Install]
+RequiredBy=reboot.target halt.target shutdown.target poweroff.target umount.target final.target
```
bk201, line 9: Test and work well.
Is this unit required to be run **after** rke2-agent or rke2-server service is stopped?

package/harvester-os/files/etc/systemd/system/rke2-shutdown.service
```diff
@@ -0,0 +1,14 @@
...
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=-/usr/local/bin/rke2-killall.sh
```
bk201, line 14: Also `rke2-killall.sh` sends `KILL (-9)` signals to processes, but rke2-server/agent's systemd services's `ExecStopPost` option sends `TERM` signals. Just wondering if it's safer to use TERM.

package/harvester-os/files/etc/systemd/system/rke2-shutdown.service
johnliu55tw: I feel like if it's a user that stopped/restarted the service, the user should decide whether `rke2-killall.sh` should be run.

package/harvester-os/files/etc/systemd/system/rke2-shutdown.service
tjjh89017: using TERM might be a good idea to give the processes some time to clean up and terminate.
but we need to have another script to do it?

bk201: If that's oneliner, we can also embed it in the systemd unit file. There are many cleanups in the rke2-killall.sh script, maybe some are unnecessary.

package/harvester-os/files/etc/systemd/system/rke2-shutdown.service
johnliu55tw: FWIW, I found some discussions about this issue, and seems like "stopping K3s/RKE2 does not stop running containers" is specifically designed to allow non-disruptive upgrade of K3s/RKE2.

- https://github.com/k3s-io/k3s/issues/2400#issuecomment-1040795816

bk201: During an upgrade, we will always wait for RKE2 upgraded first and shut down later.

bk201:
Thanks, John. This was addressed in https://github.com/harvester/harvester-installer/pull/480, so close it.

PR #272, opened 2022-04-15
Generate supportconfig for failed installations

ibrokethecloud:
This PR is related to issue: https://github.com/harvester/harvester/issues/1864

The aim is to auto generate a supportconfig when `debug` is enabled in the installation, and the installation fails.

The PR also introduces two custom plugins to copy the rke2.log and console.log into the generated supportconfig to make it easier to troubleshoot failures.

To test:
* patch /usr/sbin/harv-install to `exit 1`
* pass a minimal config via a http endpoint to enable debug. This would look as follows:
```yaml
install:
  debug: true
```
* on installation failure, harvester-installer will trigger the generation of support config, and also run the two packaged plugins

image

image

### DIFF
--- a/pkg/console/util.go
+++ b/pkg/console/util.go
@@ -449,10 +449,13 @@ func doInstall(g *gocui.Gui, hvstConfig *config.HarvesterConfig, webhooks Render
 		webhooks.Handle(EventInstallFailed)
 		printToPanel(g, fmt.Sprintf(installFailureMessage, defaultLogFilePath), installPanel)
 		if hvstConfig.Debug {
-			err = execute(ctx, g, []string{}, "/sbin/supportconfig")
-			if err != nil {
+			printToPanel(g, "support config is being generated as running in debug mode, this can take a few minutes...", installPanel)
+			fileSuffix := fmt.Sprintf("harvester_%s", rand.String(5))
+			scErr := executeSupportconfig(ctx, fileSuffix)
+			if scErr != nil {
 				printToPanel(g, fmt.Sprintf("support bundle collection failed %v", err), installPanel)
 			}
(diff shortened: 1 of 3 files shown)


bk201:
Probably we can move the installation failure messages to the end and suggest user that there is a support bundle.
Otherwise, the user might be a bit confused about why a support bundle is generated.

ibrokethecloud:
@bk201 I have disabled logging of supportconfig to console.

Now we only provide info that supportconfig is being generated and location of same.

Sample output is as follows:

image

pkg/console/util.go
```diff
@@ -448,6 +448,15 @@ func doInstall(g *gocui.Gui, hvstConfig *config.HarvesterConfig, webhooks Render
...
 		printToPanel(g, fmt.Sprintf(installFailureMessage, defaultLogFilePath), installPanel)
+		if hvstConfig.Debug {
+			printToPanel(g, "support config is being generated as running in debug mode, this can take a few minutes...", installPanel)
+			fileSuffix := fmt.Sprintf("harvester_%s", rand.String(5))
+			scErr := executeSupportconfig(ctx, fileSuffix)
+			if scErr != nil {
+				printToPanel(g, fmt.Sprintf("support bundle collection failed %v", err), installPanel)
+			}
+			printToPanel(g, fmt.Sprintf("support config is available at /var/log/scc_%s.txz", fileSuffix), installPanel)
+		}
```
bk201, line 459: LGTM.
nit: we can make the name more consistent. Either `Support config` or`Support bundle` for all strings.

ibrokethecloud: fixed it up

bk201 (approved):

johnliu55tw (approved):
LGTM

PR #274, opened 2022-04-21
Added a diff to monitoring managed chart

ibrokethecloud:
The PR adds a diff to the managed chart for monitoring and allows fleet to ignore changes to CA certificate injected at runtime.

This is related to issue: https://github.com/harvester/harvester/issues/2152

Post this change the fleet bundle associated with monitoring managed chart is reconciled correctly.

### DIFF
--- a/pkg/config/templates/rancherd-13-monitoring.yaml
+++ b/pkg/config/templates/rancherd-13-monitoring.yaml
@@ -10,6 +10,18 @@ bootstrapResources:
     version: {{ .MonitoringChartVersion }}
     defaultNamespace: cattle-monitoring-system
     repoName: harvester-charts
+    diff:
+      comparePatches:
+        - apiVersion: admissionregistration.k8s.io/v1
+          kind: ValidatingWebhookConfiguration
+          name: rancher-monitoring-admission
+          operations:
+            - { "op": "remove", "path": "/webhooks" }
+        - apiVersion: admissionregistration.k8s.io/v1
+          kind: MutatingWebhookConfiguration
+          name: rancher-monitoring-admission
+          operations:
+            - { "op": "remove", "path": "/webhooks" }
     targets:
     - clusterName: local
       clusterSelector:


pkg/config/templates/rancherd-13-monitoring.yaml
```diff
@@ -10,6 +10,18 @@ bootstrapResources:
     version: {{ .MonitoringChartVersion }}
     defaultNamespace: cattle-monitoring-system
     repoName: harvester-charts
+    diff:
```
guangbochen, line 13: @ibrokethecloud  expect the code wondering if we have the diff usage doc anywhere from the Rancher side?

guangbochen: probably referred to https://fleet.rancher.io/troubleshooting/#fleet-deployment-stuck-in-modified-state

guangbochen (commented):
LGTM in regards to a fresh Harvester install, @bk201 any concerns about the upgrade part? IIRC this template will be applied once only?

pkg/config/templates/rancherd-13-monitoring.yaml
w13915984028: @ibrokethecloud Good fix, I will do a local validation. thanks.

bk201:
> LGTM in regards to a fresh Harvester install, @bk201 any concerns about the upgrade part? IIRC this template will be applied once only?

Yes, this only works for new clusters.
To make it work for the previous version, we need to add a [migration](https://github.com/harvester/harvester/blob/master/package/upgrade/migrations/README.md) in the `harvester/harvester` repo.
- Create a file `package/upgrade/migrations/managed_charts/v1.0.1.sh`, so the file is called when the previous version is `1.0.1`.
- Implement the change, check https://github.com/harvester/harvester/blob/c6640cf139b835584234eedf5b724f62d2ae6220/package/upgrade/migrations/managed_charts/v1.0.0.sh#L34 for example.

To know how this script is called, please check: https://github.com/harvester/harvester/blob/c6640cf139b835584234eedf5b724f62d2ae6220/package/upgrade/upgrade_manifests.sh#L330-L332

pkg/config/templates/rancherd-13-monitoring.yaml
w13915984028: @ibrokethecloud Verified: your fix solves the "modified" state in newly installed cluster. Thanks.

ibrokethecloud: yep @guangbochen that is the correct document.

ibrokethecloud:
@bk201 I have created the associated PR for harvester https://github.com/harvester/harvester/pull/2182

johnliu55tw (approved):
LGTM

w13915984028 (approved):
LGTM

PR #275, opened 2022-05-11
Resolve the device path before generating partition layout

johnliu55tw:
Without resolving it first, elemental installation would fail.

## Test plan
1. Install Harvester with any node using PXE boot installation with setting `install.device` to /dev/disk/by-path instead of `/dev/vda`
2. Make sure installation succeeded

## Related issues
- https://github.com/harvester/harvester/issues/2256

### DIFF
--- a/pkg/config/cos.go
+++ b/pkg/config/cos.go
@@ -537,6 +537,11 @@ func CreateRootPartitioningLayout(devPath string) (*yipSchema.YipConfig, error) 
 		return nil, err
 	}
 
+	resolvedDevPath, err := filepath.EvalSymlinks(devPath)
+	if err != nil {
+		return nil, err
+	}
+
 	yipConfig := yipSchema.YipConfig{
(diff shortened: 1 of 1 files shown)


pkg/config/cos.go
```diff
@@ -537,6 +537,11 @@ func CreateRootPartitioningLayout(devPath string) (*yipSchema.YipConfig, error)
 		return nil, err
 	}
 
+	resolvedDevPath, err := filepath.EvalSymlinks(devPath)
```
bk201, line 540: Does `EvalSymlinks` also work for normal device files?

johnliu55tw: Yes. If it's not a symlink it would just return the original value.

bk201 (approved):
LGTM

weihanglo (approved):
Looks reasonable. Thanks!

PR #278, opened 2022-05-12
Allow users setup VLAN ID for mgmt interface

tjjh89017:
please refer https://github.com/harvester/harvester/issues/2236

Old config:
```yaml
... omit ...
install:
  ... omit ...
  networks:
    harvester-mgmt:
      interfaces:
      - name: eth0
      - name: eth1
      method: dhcp
    bond-public:
      interfaces:
      - name: eth2
      - name: eth3
      method: none
  ... omit ...
```

New config
```yaml
... omit ...
install:
  ... omit ...
  management_interface: # It will create `br-mgmt.101` as mgmt interface
    interfaces:
    - name: eth0
    - name: eth1
    method: dhcp
    vlan_id: 101
  ... omit ...
# disallow other network interfaces to be configured in the installation stage.
# Installer will have other future PR to support VLAN network pre-configuration.
```


pkg/config/testdata/harvester-config.yaml
```diff
@@ -28,12 +28,11 @@ os:
...
-  networks:
-    harvester-mgmt:
-      interfaces:
-      - name: ens0
-      - name: ens3
-      method: dhcp
+  management_interface:
+    interfaces:
+    - name: ens0
+    - name: ens3
```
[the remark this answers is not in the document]
tjjh89017: This is for direct config testing only.
I will not add vlan id on it because so far we cannot test vlan id in PXE mode.

pkg/console/install_panels.go
```diff
@@ -893,6 +899,7 @@ func addNetworkPanel(c *Console) error {
 		c.CloseElements(
 			hostNamePanel,
 			askInterfacePanel,
+			askVlanIDPanel,
```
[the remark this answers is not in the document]
tjjh89017: I think I should fix it with L810, L812, but not here.
I expected VLAN ID is following `askInterfacePanel`

pkg/console/install_panels.go
```diff
@@ -1045,6 +1056,44 @@ func addNetworkPanel(c *Console) error {
...
+			return "", nil
+		}
+		var vlanID int
+		vlanID, err = strconv.Atoi(vlanIDStr)
+		if err != nil {
+			return "VLAN ID should be a number 1 ~ 4094.", err
+		}
+		// 0 is unset
+		if vlanID < 0 || vlanID > 4094 {
+			return "VLAN ID should be a number 1 ~ 4094.", nil
```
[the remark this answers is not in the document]
tjjh89017: I'm not sure how to improve this one.
Could you give me an example for this.
Tks

pkg/console/install_panels.go
```diff
@@ -1520,19 +1574,22 @@ func addInstallPanel(c *Console) error {
...
+			tmpInterfaces := []config.NetworkInterface{}
+			for _, iface := range c.config.ManagementInterface.Interfaces {
+				if err := iface.FindNetworkInterfaceName(); err != nil {
+					logrus.Error(err)
+					printToPanel(c.Gui, err.Error(), installPanel)
+					return
 				}
-				network.Interfaces = tmpInterfaces
-				c.config.Networks[key] = network
+				if err := iface.FindNetworkInterfacHwAddr(); err != nil {
```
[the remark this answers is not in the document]
tjjh89017: I will consider to leave them separately at this moment.
But I found a typo here.
Let me think about it to merge

pkg/console/install_panels.go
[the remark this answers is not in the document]
tjjh89017: I merge it.

pkg/console/install_panels.go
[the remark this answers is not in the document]
tjjh89017: I merge it.

tjjh89017:
I will update PXE example for this PR change once we confirm the break change here.
Thanks a lot

bk201:
This PR also address https://github.com/harvester/harvester/issues/1647

pkg/config/config.go
```diff
@@ -93,9 +95,9 @@ type Webhook struct {
 }
 
 type Install struct {
-	Automatic bool               `json:"automatic,omitempty"`
-	Mode      string             `json:"mode,omitempty"`
-	Networks  map[string]Network `json:"networks,omitempty"`
+	Automatic           bool    `json:"automatic,omitempty"`
+	Mode                string  `json:"mode,omitempty"`
+	ManagementInterface Network `json:"managementInterface,omitempty"`
```
bk201, line 100: How about naming it `ManagementNetwork`? Since it's a Network type and inside it, there is already a field `Interfaces`.

pkg/console/install_panels.go
```diff
@@ -1045,6 +1060,44 @@ func addNetworkPanel(c *Console) error {
...
+		var vlanID int
+		vlanID, err = strconv.Atoi(vlanIDStr)
+		if err != nil {
+			return ErrMsgVLANShouldBeANumberInRange, err
+		}
+		// 0 is unset
+		if vlanID < 0 || vlanID > 4094 {
+			return ErrMsgVLANShouldBeANumberInRange, fmt.Errorf(ErrMsgVLANShouldBeANumberInRange)
+		}
+		mgmtNetwork.VlanID = vlanID
```
bk201, line 1088: By typing an invalid value, user can't see the error.
The installer simply restarts.

pkg/config/templates/rke2-canal-config.yaml
```diff
@@ -6,4 +6,4 @@ metadata:
 spec:
   valuesContent: |-
     flannel:
-      iface: harvester-mgmt
+      iface: ""
```
bk201, line 9: Is it OK to remove this HelmChartConfig resource if the iface is empty?

pkg/console/validator.go
```diff
@@ -165,57 +165,50 @@ func checkIPList(ipList []string) error {
...
+			return err
 		}
 	}
 
-	for _, network := range networks {
-		for _, iface := range network.Interfaces {
-			if err := checkInterface(iface); err != nil {
-				return err
-			}
+	// TODO check VLAN ID in 0-4094 (0 is unset)
```
bk201, line 186: Nice to have the check here.

pkg/console/install_panels.go
```diff
@@ -1501,11 +1554,16 @@ func addInstallPanel(c *Console) error {
...
 						printToPanel(c.Gui, fmt.Sprintf("can't apply networks: %s", err), installPanel)
 						return
 					}
-					vip, err := getVipThroughDHCP(config.MgmtInterfaceName)
+					mgmtName := config.MgmtInterfaceName
+					vlanID := c.config.ManagementInterface.VlanID
+					if vlanID >= 2 && vlanID <= 4094 {
+						mgmtName = fmt.Sprintf("%s.%d", mgmtName, vlanID)
+					}
+					vip, err := getVipThroughDHCP(mgmtName)
```
bk201, line 1566: This is re-used in another place, might be good to implement it as a method in Network.

pkg/config/config.go
tjjh89017: I'm not sure
ManagementNetwork might conflict with the term of the network of Canal.

pkg/console/validator.go
tjjh89017: fixed

pkg/console/install_panels.go
tjjh89017: fixed

pkg/console/install_panels.go
tjjh89017: fixed

pkg/config/config.go
bk201: It's a field name in Harvester configuration; why does it conflict with Canal things?

And do we need to remove the original `Networks` map? We can keep the `Networks` map but only allow the `harvester-mgmt` key here. This provides backward compatibility. For users who don't have other networks configuration, they don't need to alter their existing configuration files.

tjjh89017: > Centralized Gateway[¶](https://kubeovn.github.io/docs/v1.10.x/en/guide/subnet/#centralized-gateway)

It will confuse us, we will not sure if we are talking about Pod's Management Network or Management Interface (node's network)

> And do we need to remove the original Networks map? We can keep the Networks map but only allow the harvester-mgmt key here. This provides backward compatibility. For users who don't have other networks configuration, they don't need to alter their existing configuration files.

I will prefer to remove `harvester-mgmt` term in future version.
Make sure it will not confuse anyone.

bk201: It will not because [we have the document](https://docs.harvesterhci.io/v1.0/install/harvester-configuration/#installnetworks). We don't present the field to the end-user in the installer UI either. If it confuses anyone, we can add a better document or UI terms.

Harvester is already GA for almost one year, and we already have a user base. So keeping more compatibility here would be appreciated.

tjjh89017: > Harvester is already GA for almost one year, and we already have a user base. So keeping more compatibility here would be appreciated.

I didn't really get it about the naming.
If we remove the `harvester-mgmt` and it will never be showed in Linux system.
And if we keep the naming in installer config, won't it confuse the users?

It will only affect in fresh installation but not upgrading.
Nodes from previous version will not be affected at all.

It will be a little bit weird in docs.
we ask users to write config with `harvester-mgmt`, but this keyword doesn't mean anything.

bk201: The key here is compatibility and not bringing too much impact to existing users.
I understand we do try our best to migrate existing configurations in the upgrade, but we also need to take care of existing users who have already an infrastructure to use Harvester. For example:
- Customers/users who already have their infrastructure code to spawn Harvester on demand.
- Our QA: they do have infrastructure codes to deploy Harvester repeatedly.
- Our lab maintainer.
- Equinix Metal users.
- Any developers who re-deploy Harvester daily.

Even if we keep the original `Networks`, we still need to check if there are additional configured network keys because they are no longer supported. What I really suggest is to minimize impact. Not arguing the naming.

tjjh89017: I will suggest, if we put so many changes in this verison, we should consider to take this chance to migrate all we should.
From my side, I will not consider to leave some legacy entries if we don't need it and that is not for a patch build.

and also you mentioned the additional network keys are not supported anymore, that means some users still need to change the script or be noticed some changes to make sure the scripts will fit the new installer config format.

I also minimize impact for the future change to make sure there is no other will confuse with that entry and code.

But you reminded me, that we should ask users to add `scheme_version` to their config.yaml in the future if we consider to have scheme version for future migration plan.

pkg/config/config.go
```diff
@@ -155,8 +157,9 @@ type ClusterNetwork struct {
 }
 
 type HarvesterConfig struct {
-	ServerURL string `json:"serverUrl,omitempty"`
-	Token     string `json:"token,omitempty"`
+	SchemeVersion uint32 `json:"schemeVersion,omitempty"`
```
guangbochen, line 160: Can u specify what the `SchemeVersion` is used for?

pkg/config/config.go
guangbochen: both design has pros and cons, if upgrade capability and customer support is our main concerns we may just ask feedbacks internally, and be aware of need to adding those changes to the ipxe repo and existing infra code.

pkg/config/cos.go
```diff
@@ -325,49 +303,57 @@ func SaveOriginalNetworkConfig() error {
...
-		if err != nil {
-			return err
+	if len(mgmtInterface.Interfaces) > 0 {
+		bondMgmt := Network{
+			Interfaces:  mgmtInterface.Interfaces,
+			Method:      NetworkMethodNone,
+			BondOptions: mgmtInterface.BondOptions,
+			MTU:         mgmtInterface.MTU,
 		}
+		err = updateBond(stage, MgmtBondInterfaceName, &bondMgmt)
```
guangbochen, line 330: prefer to return the error in here directly.

pkg/config/cos.go
```diff
@@ -407,10 +377,27 @@ func updateBond(stage *yipSchema.Stage, name string, network *Network) error {
...
+	stage.Directories = append(stage.Directories, yipSchema.Directory{
+		Path:        "/etc/wicked/scripts",
+		Permissions: 0644,
+		Owner:       0,
+		Group:       0,
+	})
+
+	var postUpScript string
+	var err error
+	postUpScript, err = render("wicked-setup-bond.sh", MgmtBondInterfaceName)
```
guangbochen, line 390:
```suggestion
	postUpScript, err := render("wicked-setup-bond.sh", MgmtBondInterfaceName)
```

pkg/config/cos.go
```diff
@@ -444,6 +431,92 @@ func updateBond(stage *yipSchema.Stage, name string, network *Network) error {
...
+	// setup pre up script
+	stage.Directories = append(stage.Directories, yipSchema.Directory{
+		Path:        "/etc/wicked/scripts",
+		Permissions: 0644,
+		Owner:       0,
+		Group:       0,
+	})
+
+	var preUpScript string
+	preUpScript, err = render("wicked-setup-bridge.sh", MgmtBondInterfaceName)
```
guangbochen, line 453: prefer to use `:=` assignment if they are not top level variables.

pkg/console/util.go
```diff
@@ -409,6 +409,9 @@ func doInstall(g *gocui.Gui, hvstConfig *config.HarvesterConfig, webhooks Render
 		return err
 	}
 	defer os.Remove(cosConfigFile)
+	
+	// save current scheme version to harvester.config
+	hvstConfig.SchemeVersion = config.SchemeVersion
```
guangbochen, line 414: Who will maintains the schemeVersion? if users can modifiy it how can we verify it accoridngly?

pkg/config/config.go
tjjh89017: Currently, we only logged the version but do nothing about this.
It's reserved for the future.
if we want to do some migration in the future.

pkg/config/cos.go
tjjh89017: ok

pkg/config/cos.go
tjjh89017: ok

pkg/config/cos.go
tjjh89017: ok

pkg/console/util.go
tjjh89017: Once we change the config struct, we should increase the scheme version.
and users shouldn't change it on their own.
It will be referred to by the future migration script.
Currently, we could not verify it if users change it manually.

pkg/config/constants.go
```diff
@@ -9,7 +9,8 @@ const (
 	NetworkMethodStatic = "static"
 	NetworkMethodNone   = "none"
 
-	MgmtInterfaceName = "harvester-mgmt"
+	MgmtInterfaceName     = "mgmt-br"
+	MgmtBondInterfaceName = "mgmt-bo"
```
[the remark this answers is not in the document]
tjjh89017: Canwu asked me to use mgmt-br and mgmt-bo for mgmt interface.
I will check with him again

tjjh89017:
add validater to check if users didn't provide `management_interface` entry and guide users to check docs for new scheme config.

pkg/config/constants.go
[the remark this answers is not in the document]
yaocw2020: The prefix would be the name of the cluster network, thus mgmt is reasonable.

pkg/config/config.go
yaocw2020: Please add an annotation.

pkg/config/cos.go
```diff
@@ -325,49 +303,57 @@ func SaveOriginalNetworkConfig() error {
...
-	if !ok {
-		return errors.New("no management network defined")
+func UpdateManagementInterfaceConfig(stage *yipSchema.Stage, mgmtInterface Network, run bool) (string, error) {
+	if len(mgmtInterface.Interfaces) == 0 {
+		return "", errors.New("no slave defined for management network bond")
 	}
-	if len(mgmtNetwork.Interfaces) == 0 {
-		return errors.New("no slave defined for management network bond")
+
+	switch mgmtInterface.Method {
```
yaocw2020, line 315: What about using if expression rather than switch here because we do nothing in the case branch?

pkg/config/cos.go
```diff
@@ -325,49 +303,57 @@ func SaveOriginalNetworkConfig() error {
...
-			err = updateNIC(stage, name, &network)
+	if len(mgmtInterface.Interfaces) > 0 {
+		bondMgmt := Network{
+			Interfaces:  mgmtInterface.Interfaces,
+			Method:      NetworkMethodNone,
+			BondOptions: mgmtInterface.BondOptions,
+			MTU:         mgmtInterface.MTU,
 		}
+		err = updateBond(stage, MgmtBondInterfaceName, &bondMgmt)
 		if err != nil {
```
yaocw2020, line 331: The coding mode shown below is more recommend.
``` go
if err := updateBond(stage, MgmtBondInterfaceName, &bondMgmt); err != nil {
    return "", err
}
```

pkg/config/cos.go
```diff
@@ -325,49 +303,57 @@ func SaveOriginalNetworkConfig() error {
...
+	if err = updateBridge(stage, MgmtInterfaceName, &mgmtInterface); err != nil {
+		return "", err
+	}
+
+	name := MgmtInterfaceName
+	if mgmtInterface.VlanID >= 2 && mgmtInterface.VlanID <= 4094 {
+		name = fmt.Sprintf("%s.%d", name, mgmtInterface.VlanID)
+	}
+
+	switch mgmtInterface.Method {
```
yaocw2020, line 345: What about merging the 315 line switch expression with this switch?

pkg/config/cos.go
```diff
@@ -325,49 +303,57 @@ func SaveOriginalNetworkConfig() error {
...
+		stage.Files = append(stage.Files, yipSchema.File{
+			Path:        fmt.Sprintf("/etc/sysconfig/network/ifroute-%s", name),
+			Content:     fmt.Sprintf("default %s - %s\n", mgmtInterface.Gateway, name),
+			Permissions: 0600,
+			Owner:       0,
+			Group:       0,
+		})
+	case NetworkMethodDHCP, NetworkMethodNone:
+		stage.Commands = append(stage.Commands, fmt.Sprintf("rm -f /etc/sysconfig/network/ifroute-%s", name))
 	}
```
yaocw2020, line 357: I believe `case default:` is necessary for the switch expression.

pkg/console/util.go
yaocw2020: Please add an annotation.

pkg/config/templates/wicked-setup-bridge.sh
```diff
@@ -0,0 +1,20 @@
...
+
+case $ACTION in
+	pre-up)
+		# enable vlan-aware
+		ip link set $INTERFACE type bridge vlan_filtering 1
+		;;
+
+	post-up)
+		# inherit MAC address from {{ . }}
+		ip link set dev $INTERFACE address $(ip -json link show dev {{ . }} | jq -j '.[0]["address"]')
```
w13915984028, line 14: `$(ip -json link show dev {{ . }} | jq -j '.[0]["address"]')`

please consider add check and log, to make sure the MAC is really got; and log essential info when failed

pkg/console/validator.go
```diff
@@ -41,6 +41,8 @@ var (
 	ErrMsgNetworkMethodUnknown = "unknown network method"
 
 	ErrMsgSystemSettingsUnknown = "unknown system settings: %s"
+
+	ErrMsgManagementInterfaceNotFound = "`networks` is deprecated, please use `management_interface` for new config and refer https://docs.harvesterhci.io/v1.1/install/harvester-configuration/#installmanagement_interface"
```
w13915984028, line 45: it seems uncommen to add "`" in code for message, please consider to remove them

pkg/config/cos.go
```diff
@@ -444,6 +429,90 @@ func updateBond(stage *yipSchema.Stage, name string, network *Network) error {
 	return nil
 }
 
+func updateBridge(stage *yipSchema.Stage, name string, mgmtNetwork *Network) error {
+	// add Bridge named MgmtInterfaceName and attach Bond named MgmtBondInterfaceName to bridge
+
+	needVlanInterface := false
+	// pvid is always 1, if vlan id is 1, it means untagged vlan.
+	if mgmtNetwork.VlanID >= 2 && mgmtNetwork.VlanID <= 4094 {
```
w13915984028, line 437: it's better to define a const for 4094, it's scattered in the code

pkg/config/cos.go
tjjh89017: we will need 3 conditions to match.
but switch-case could use 1 case to cover

pkg/config/cos.go
tjjh89017: We need the mgmt name after updateBridge and updateBond.
So we cannot merge it.

pkg/config/config.go
bk201: The user can know configuration changes:
Screen Shot 2022-09-20 at 2 49 28 PM

For the scheme version part, could we do the same?
Screen Shot 2022-09-20 at 2 46 26 PM

pkg/config/cos.go
tjjh89017: We do nothing in this default case.
Just skip it.

pkg/config/config.go
tjjh89017: fixed

pkg/config/templates/wicked-setup-bridge.sh
tjjh89017: move it to bond post up script

pkg/console/validator.go
tjjh89017: removed

pkg/config/cos.go
tjjh89017: I will not change this in this stage

tjjh89017: There are some ranges between 0 to 4094 and 2 to 4094.
It will confuse if we only set a variable for 4094 but not for 0 and 2.
and it's hard to give a name for 0 and 2.

yaocw2020 (approved):
LGTM

w13915984028 (approved):
LGTM, thanks.

bk201 (approved):
lgtm

guangbochen (approved):

PR #282, opened 2022-05-27
VIP panel: wrap info panel

bk201:
The response message from DHCP servers could be more than one line.
Allow the panel to wrap text.

Related issue: https://github.com/harvester/harvester/issues/2224

**Test plans**
- Disable DHCPD and select `Automatic (DHCP)` method in the VIP panel. The error message should wrap.
Screen Shot 2022-05-27 at 4 00 46 PM

- Select the `Static` method and enter a very long text, the error message should wrap too.
Screen Shot 2022-05-27 at 4 23 15 PM

- This PR also cleans the left-over error messages when jumping back to the DNS panel and jumping to the VIP panel again
  -  In the VIP panel, enter some error text in Static mode. An error message should display.
  -  Jump back to the DNS panel.
  -  Jump to the VIP panel. The error message should be clear.

### DIFF
--- a/pkg/console/install_panels.go
+++ b/pkg/console/install_panels.go
@@ -1595,6 +1595,8 @@ func addUpgradePanel(c *Console) error {
 }
 
 func addVIPPanel(c *Console) error {
+	setLocation := createVerticalLocator(c)
+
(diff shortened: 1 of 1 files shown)


weihanglo (approved):
It works. Thanks for the fix!

joshmeranda:
Looks good to me, thanks!

joshmeranda (approved):

PR #283, opened 2022-05-30
VIP must be different from management NIC's IP

weihanglo:
related: https://github.com/harvester/harvester/issues/2229

Assert the difference between VIP and management NIC's IP.

### Test plan

0. Build the installer and run it.
1. Select `Static` as your IPv4 Method
2. Type a IP for management NIC
    
3. Select `Static` as your VIP mode
4. Type the same IP as NIC management IP
    
5. Expect to see the error message **"VIP must not be the same as management NIC's IP"**.


tjjh89017 (approved):
LGTM

chrisho:
👍

chrisho (commented):
LGTM

PR #287, opened 2022-06-08
fix2361 increase grafana resource limit

w13915984028:
(1) Expose grafana pod resource setting managedchart rancher-monitoring 
(2) Increase grafana pod resource setting
current:
```
  resources:
    limits:
      memory: 200Mi --
      cpu: 200m
    requests:
      memory: 100Mi --
      cpu: 100m
```

after:
```
   resources: 
     limits:
       cpu: 200m
       memory: 500Mi --
     requests:
       cpu: 100m
       memory: 200Mi --

```

https://github.com/harvester/harvester/issues/2361

### DIFF
--- a/pkg/config/templates/rancherd-13-monitoring.yaml
+++ b/pkg/config/templates/rancherd-13-monitoring.yaml
@@ -39,6 +39,13 @@ resources:
           type: pvc
           accessModes:
           - ReadWriteOnce
+        resources:
+          limits:
+            cpu: 200m
+            memory: 500Mi
+          requests:
+            cpu: 100m
+            memory: 200Mi
       prometheus:
         prometheusSpec:
           evaluationInterval: 1m


pkg/config/templates/rancherd-13-monitoring.yaml
```diff
@@ -39,6 +39,13 @@ resources:
           type: pvc
           accessModes:
           - ReadWriteOnce
+        resources:
+          limits:
+            cpu: 200m
+            memory: 500Mi
```
Vicente-Cheng, line 45: just curious, how we know 500Mi is the suitable value for grafana?

Vicente-Cheng: yap, I feel it to be a bit higher :p 

And it's also good because we could modify it via `kubectl` after this commit.
Thanks for your explanation.

Vicente-Cheng (approved):

bk201 (approved):
lgtm! We also need to handle the upgrade part. An example to change prometheus values in v1.0.0 -> v1.0.1: https://github.com/harvester/harvester/blob/96b907149989c6e96e84be51de6ce0b42ad4487e/package/upgrade/migrations/managed_charts/v1.0.0.sh#L19-L26

guangbochen (approved):
LGTM, @w13915984028 please help to add the upgrade part support, thanks.

PR #289, opened 2022-06-10
console: dashboard: show node ip for identify nodes

tjjh89017:
please refer https://github.com/harvester/harvester/issues/2087
Test case in https://github.com/harvester/harvester/issues/2087

It will check the interface with the default route to have the primary/first IP from it.
If not, it will fall back to harvester-mgmt and get IP from it.

### Test plan

1. build harvester-installer with `make`
2. copy harvester-installer with `scp bin/harvester-installer rancher@<IP>:<PATH>
3. Switch harvester console to TTY2 or other (press CTRL+ALT+F2)
4. execute harvester-installer in current TTY `DEBUG=true HARVESTER_DASHBOARD=true TTY=/dev/tty <PATH>/harvester-installer`
5. check the screen has Node IP and IP address (refer the figure below)
Screenshot_node1_2022-06-14_10:53:17

### DIFF
--- a/pkg/console/dashboard_panels.go
+++ b/pkg/console/dashboard_panels.go
@@ -69,7 +69,15 @@ func (c *Console) layoutDashboard(g *gocui.Gui) error {
 		v.Wrap = true
 		go syncManagementURL(context.Background(), g)
 	}
-	if v, err := g.SetView("status", maxX/2-40, 14, maxX/2+40, 18); err != nil {
+	if v, err := g.SetView("nodeIP", maxX/2-40, 14, maxX/2+40, 18); err != nil {
+		if err != gocui.ErrUnknownView {
+			return err
+		}
+		v.Frame = false
+		v.Wrap = true
+		go syncNodeIP(context.Background(), g)
+	}
+	if v, err := g.SetView("status", maxX/2-40, 18, maxX/2+40, 22); err != nil {
 		if err != gocui.ErrUnknownView {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


bk201 (approved):
LGTM.

pkg/console/dashboard_panels.go
```diff
@@ -249,6 +257,65 @@ func getVIP() string {
 	return outStr
 }
 
+func syncNodeIP(ctx context.Context, g *gocui.Gui) {
+	// sync IP at the beginning
+	doSyncNodeIP(g)
+
+	syncDuration := 30 * time.Second
```
Vicente-Cheng, line 264: Does it mean that  If I change the node IP after beginning, I need  30 seconds to check the current IP?
If so, is the value too long?

tjjh89017: Currently, we disallow users to change node IP.
These 30 seconds should be only needed for booting up if the network is preparing.

Vicente-Cheng: OK, keep in mind that. Thanks for explaining :)

Vicente-Cheng (approved):
LGTM

bk201:
As the suggestion in the demo, we can display the node name too.

pkg/console/dashboard_panels.go
```diff
@@ -249,6 +257,75 @@ func getVIP() string {
...
+		device   string
+	)
+
+	// find node hostname
+	cmd = `hostname | tr -d '\r\n'`
+	out, err = exec.Command("/bin/sh", "-c", cmd).Output()
+	hostname = string(out)
+	if err != nil || hostname == "" {
+		logrus.Warnf("node didn't have a hostname")
+		hostname = ""
```
Vicente-Cheng, line 304: Is any chance the hostname empty and valid?
I am not sure how we could do that :p 
If the empty hostname is invalid, maybe we should do some error handling here?

tjjh89017: We should not change it.
Hostname should not be empty.
I think we could only show error msg in console or log.
hmm

Vicente-Cheng: OK, looks like the only thing we could do is show the error msg.

Vicente-Cheng (approved):
LGTM

bk201 (approved):

PR #295, opened 2022-07-05
WIP: add rke2-coredns helm chart config

FrankYang0529:
https://github.com/harvester/harvester/issues/2336


FrankYang0529:
Close the PR since it's fixed in upstream.

PR #296, opened 2022-07-05
Bump baseOS to 20220705

tjjh89017:

### DIFF
--- a/scripts/package-harvester-os
+++ b/scripts/package-harvester-os
@@ -14,7 +14,7 @@ source ${SCRIPTS_DIR}/version-rancher
 source ${SCRIPTS_DIR}/version-harvester ${TOP_DIR}/../harvester
 source ${SCRIPTS_DIR}/version-monitoring
 
-BASE_OS_IMAGE="rancher/harvester-os:20220526"
+BASE_OS_IMAGE="rancher/harvester-os:20220705"
 HARVESTER_OS_IMAGE=rancher/harvester-os:$VERSION
 
 cd ${PACKAGE_HARVESTER_OS_DIR}


bk201:
@guangbochen @Vicente-Cheng This PR should switch the based OS to SLE Micro. Please help test it if possible. Thanks!

guangbochen (commented):
seems we need change the submission branch to v1.0 and update the script https://github.com/harvester/harvester-installer/blob/v1.0/scripts/build-bundle#L30 to clone Harvester v1.0 branch as well.

guangbochen (approved):
LGTM, the ISO build works and was able to install a new cluster successfully, the latest OS is now based on `SLE Micro 5.2`, the upgrade part also works but only runs with a simple cluster so additional cases can be verified separately, thanks.
```
NAME="SLE Micro"
VERSION="5.2"
VERSION_ID="5.2"
PRETTY_NAME="Harvester master"
ID="sle-micro-rancher"
ID_LIKE="suse"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sle-micro-rancher:5.2"
VARIANT="Harvester"
VARIANT_ID="Harvester-20220705"
GRUB_ENTRY_NAME="Harvester master"
```
image

Vicente-Cheng (approved):
I build ISO with harvester master. ISO build and install successfully.
Then I create a simple cluster (with 3 VMs) and do some testing like uploading image, creating volume, creating VM, and migrating VM. It all works normally. I thought the new base OS worked well on basic cases. Thanks!

bk201 (approved):
LGTM!

PR #297, opened 2022-07-06
[WIP] bump rke2 to v1.22.12-rc3+rke2r1

FrankYang0529:
ref: https://github.com/harvester/harvester/issues/2336

**Test Steps**:
* Create a cluster.
* Check whether `kube-system/rke2-coredns-rke2-coredns-autoscaler` resources are
```
limits:
  cpu: 100m
  memory: 64Mi
requests:
  cpu: 25m
  memory: 16Mi
```

### DIFF
--- a/package/harvester-os/files/usr/sbin/harv-install
+++ b/package/harvester-os/files/usr/sbin/harv-install
@@ -151,8 +151,11 @@ preload_rke2_images()
     fi
 
     RKE2_IMAGES_DIR="/var/lib/rancher/rke2/agent/images"
+    RKE2_IMAGES_LISTS_DIR="/tmp/images-lists"
     mkdir -p $TARGET/$RKE2_IMAGES_DIR
+    mkdir -p $TARGET/$RKE2_IMAGES_LISTS_DIR
     mount --bind ${ISOMNT}/bundle/harvester/images $TARGET/$RKE2_IMAGES_DIR
+    mount --bind ${ISOMNT}/bundle/harvester/images-lists $TARGET/$RKE2_IMAGES_LISTS_DIR
 
     cd $TARGET
     mount --bind /dev dev
@@ -192,7 +195,7 @@ preload_rke2_images()
       done
 
       # make sure all preloading images are ready
-      for i in /var/lib/rancher/rke2/agent/images/*.txt; do
+      for i in /tmp/images-lists/*.txt; do
         stdbuf -oL ctr-check-images.sh $i
       done
 
@@ -212,6 +215,7 @@ EOF
     umount proc
     cd - &> /dev/null
     umount ${TARGET}/${RKE2_IMAGES_DIR}
+    umount ${TARGET}/${RKE2_IMAGES_LISTS_DIR}
 }
 
 preload_rancherd_images()
--- a/scripts/build-bundle
+++ b/scripts/build-bundle
@@ -18,10 +18,12 @@ source ${SCRIPTS_DIR}/lib/image
 BUNDLE_DIR="${PACKAGE_HARVESTER_OS_DIR}/iso/bundle"
 CHARTS_DIR="${PACKAGE_HARVESTER_REPO_DIR}/charts"
 IMAGES_DIR="${BUNDLE_DIR}/harvester/images"
+IMAGES_LISTS_DIR="${BUNDLE_DIR}/harvester/images-lists"
 RANCHERD_IMAGES_DIR="${BUNDLE_DIR}/rancherd/images"
 
 mkdir -p ${CHARTS_DIR}
 mkdir -p ${IMAGES_DIR}
+mkdir -p ${IMAGES_LISTS_DIR}
 mkdir -p ${RANCHERD_IMAGES_DIR}
 
 # Prepare Harvester chart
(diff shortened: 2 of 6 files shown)


guangbochen (approved):
LGTM

bk201:
I saw two nodes fail when loading the image:
Screen Shot 2022-07-06 at 3 20 27 PM

[/var/log/console.log](https://github.com/harvester/harvester-installer/files/9052539/console.log)
[/run/cos/target/rke2.log](https://github.com/harvester/harvester-installer/files/9052540/rke2.log)

From the RKE2 log, it seems the etcd fail and the RKE2 process exit:
```
time="2022-07-06T07:16:02Z" level=info msg="Waiting for etcd manifest"
time="2022-07-06T07:16:02Z" level=fatal msg="Failed to wait for latest etcd manifest to be deployed: timed out waiting for the condition"
```

bk201:
I saw missing images in the air-gapped environment:
- rancher/hardened-cni-plugins:v1.0.1-build20220223
- rancher/mirrored-ingress-nginx-kube-webhook-certgen:v1.0

We can fix this by removing the previous hard-code workaround: https://github.com/bk201/harvester-installer/commit/2565d85021f04bbbcdd44b9ed6dd7977b102f3d5

--
I also found the node promotion fail:
Screen Shot 2022-07-11 at 11 09 02 PM

Looks like some behavior change and we need to tune the promote job script.

package/harvester-os/files/usr/sbin/harv-install
```diff
@@ -180,29 +180,49 @@ preload_rke2_images()
...
+      rm /rke2.log
+      rm /etc/rancher/rke2/rke2.yaml
+      rm -rf /var/lib/rancher/rke2/server
+      rm -rf /var/lib/rancher/rke2/agent/pod-manifests/*
+
       export PATH=/var/lib/rancher/rke2/bin:$PATH
       export CONTAINERD_ADDRESS=/run/k3s/containerd/containerd.sock
 
+      echo "Start containerd..."
+      containerd -c /var/lib/rancher/rke2/agent/etc/containerd/config.toml -a $CONTAINERD_ADDRESS --state /run/k3s/containerd --root /var/lib/rancher/rke2/agent/containerd &> /containerd.log &
```
guangbochen, line 202: can u please share a little bit more about the RKE2 changes and why we need to start the container independently for now?

package/harvester-os/files/usr/sbin/harv-install
```diff
@@ -180,29 +180,49 @@ preload_rke2_images()
...
         sleep 1
       done
 
+      # load images
+      for i in /var/lib/rancher/rke2/agent/images/*.tar.zst; do
+        echo "Load images from $i"
+        zstd -d $i -o /usr/local/image.tar
+        ctr -n=k8s.io images import /usr/local/image.tar
+        rm /usr/local/image.tar
+      done
```
guangbochen, line 216: why do we need to load the images independently again?

bk201 (approved):
LGTM.
- Test 4 node provision in air-gapped env.
- Test upgrade from v1.0.2 to the build ISO.
- Check rke2-coredns-rke2-coredns-autoscaler resources.

package/harvester-os/files/usr/sbin/harv-install
FrankYang0529: Hi @guangbochen, this is what Kiefer encountered when he tried to boot four nodes simultaneously on the same disk. RKE2 server may go down before images are all loaded. We changed to start the RKE2 server to load contained binary and don't rely on RKE2 to load images to avoid the timeout issue.

package/harvester-os/files/usr/sbin/harv-install
FrankYang0529: Same reason in an upper comment.

package/harvester-os/files/usr/sbin/harv-install
FrankYang0529: I tried to read RKE2 source code, but I can't find where may let the RKE2 server go down.

guangbochen: maybe @brandond can u share some insights here, thanks?

bk201: Previously, we bring up RKE2 server by calling `rke-server` and let RKE2 preload images in ZST tarballs. I'm not quite what changed, the rke2-server seems to timeout sooner if it can't see etcd ready.

brandond (changes requested):

package/harvester-os/files/usr/sbin/harv-install
dereknola: We had hoped that 8.5minutes would cover the time for the tarballs to be uncompressed. But I am happy to raise the timeout limit if you can give me a estimate of how much time.

package/harvester-os/files/usr/sbin/harv-install
```diff
@@ -180,29 +180,49 @@ preload_rke2_images()
...
       do
         sleep 1
       done
 
+      # load images
+      for i in /var/lib/rancher/rke2/agent/images/*.tar.zst; do
+        echo "Load images from $i"
+        zstd -d $i -o /usr/local/image.tar
+        ctr -n=k8s.io images import /usr/local/image.tar
+        rm /usr/local/image.tar
```
brandond, line 215: If we have to ship with this workaround, lets avoid decompressing the tarballs to disk, as they can be several GB.
```suggestion
        zstd -d $i | ctr -n=k8s.io images import -
```

brandond (changes requested):

Vicente-Cheng:
Test with the latest change.

Image/Volume/VM create works well. VM migration works well.
Looks good to me if there is nothing to change.

Vicente-Cheng (approved):

bk201:
Test with https://github.com/harvester/os2/pull/46/, lgtm

PR #299, opened 2022-07-06
Use correct harvester/harvester branch

bk201:
Make sure we use the correct backend code when building from the `v1.0` branch.


guangbochen (approved):
LGTM, thanks.

Vicente-Cheng (approved):
LGTM, thanks :)

PR #308, opened 2022-07-18
Introduce Logic To Remove Duplicate DNS Entries

irishgordo:
* util private function introduced to remove duplicate DNS entries that
  may get passed in during installation
* additional test case for helper function added

Resolves: feat/remove-duplicate-dns-entries

### DIFF
--- a/pkg/console/util.go
+++ b/pkg/console/util.go
@@ -197,8 +197,21 @@ func enableNTPServers(ntpServerList []string) error {
 	return nil
 }
 
+func _removeDuplicateDNSEntry(dnsServerList []string) []string {
+	allKeys := make(map[string]bool)
+	var list []string
+	for _, item := range dnsServerList {
+		if _, value := allKeys[item]; !value {
+			allKeys[item] = true
+			list = append(list, item)
+		}
+	}
+	return list
+}
+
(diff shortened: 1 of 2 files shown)


irishgordo:
Screenshot from 2022-07-18 14-01-38
Screenshot from 2022-07-18 14-00-30
Screenshot from 2022-07-18 13-26-30

Right now with `aa3d95d` Harvester, duplicate DNS Entries will get populated in the `/oem/custom_99.yaml` and also in the `/etc/sysconfig/network/config` - this PR will automatically remove duplicate DNS entries if they get passed in during config.

This might be un-needed but I'm curious if removing the duplicates will create a more stable platform?
Open to any suggestions, feedback, comments and I'm willing to iterate on this too :smile: !

bk201:
Thanks, @irishgordo, I did some experiments and found `netconfig` is smart enough to remove duplicated entries:
e.g,
```
# /etc/sysconfig/network/config
NETCONFIG_DNS_STATIC_SERVERS="1.1.1.1 1.1.1.1"
```

```
# /etc/resolv.conf
nameserver 1.1.1.1
```

Didn't dig into the source yet.

PR #309, opened 2022-07-20
build-bundle: reset chart workspace in bundle cache path

bk201:
We need to make sure the `harvester/deploy/charts` is reset before exiting
this script. Otherwise it will intervene version values in other scripts.

This PR is to make the speed up tip https://github.com/harvester/harvester/wiki/Speed-up-Harvester-ISO-build#skip-images-pulling-in-build-bundle-stage easier.


Vicente-Cheng (approved):
LGTM

futuretea (approved):
LGTM

PR #314, opened 2022-07-29
Disable rancher host port 8443

bk201:


Vicente-Cheng:
Also give the reference here
https://github.com/rancher/rancherd/blob/9ed47069632326399d7efdc4bf1c72143505171d/config-example.yaml#L21-

Vicente-Cheng (approved):
LGTM, thanks!

tjjh89017 (approved):
LGTM

PR #325, opened 2022-08-24
feat: bump rancher to v2.6.7

FrankYang0529:
ref https://github.com/harvester/harvester/issues/2011


FrankYang0529:
Close the PR. We will trace in https://github.com/harvester/harvester-installer/pull/344

PR #330, opened 2022-09-05
feat(upgrade): inject minUpgradableVersion into OS image

starbops:
A version string specifying the minimum version required to be able to upgrade is injected into `harvester-release.yaml` for further usage.

Related issue: harvester/harvester#2431
Related PR: harvester/harvester#2730

Signed-off-by: Zespre Chang <zespre.chang@suse.com>


starbops:
The branch name should not be that long due to the Kubernetes label length restriction (<63 chars). Close this PR.

PR #332, opened 2022-09-06
fix(install_panels): add a field for MTU config

starbops:
**Problem:**

Users are not able to configure `harvester-mgmt` bond with the desired MTU value via ISO installation. This could be a problem while the target machine being provisioned resides on a network whose path MTU is below 1500 bytes.

**Solution:**

Along with the ISO installation, users can configure the bond's MTU on the network configuration page when choosing the "Static" option for "IPv4 Method".

Configure MTU value:
Screen Shot 2022-09-24 at 00 20 48

Confirm that the MTU is shown on the page:
Screen Shot 2022-09-24 at 01 12 47

Press <Alt+F2> and execute `ip addr` after applied the network config to see if the MTU is being set:
Screen Shot 2022-09-24 at 00 21 59

**Related Issue:**

harvester/harvester#2437

**Test plan:**

1. Spin up a node with Harvester ISO installation
2. Go through installation steps
3. On the network configuration page, select "Static" in the network method field then the MTU field will show up
4. Configure IP address, gateway, and MTU with a value (should be in the range of 576 ~ 9000)
5. Go through the remaining installation steps
6. On the confirm page, check if the MTU value shown on the page
7. Proceed with the installation
8. After the installation succeeded, get into the node and check the MTU value on the uplink interface, management bridge and bond interface
Screen Shot 2022-09-24 at 00 55 20

### DIFF
--- a/pkg/console/install_panels.go
+++ b/pkg/console/install_panels.go
@@ -41,6 +41,7 @@ const (
 
 const (
 	ErrMsgVLANShouldBeANumberInRange string = "VLAN ID should be a number 1 ~ 4094."
+	ErrMsgMTUShouldBeANumber         string = "MTU should be a number."
 )
 
 var (
@@ -819,7 +820,7 @@ func showNetworkPage(c *Console) error {
 	if mgmtNetwork.Method != config.NetworkMethodStatic {
 		return showNext(c, askInterfacePanel, askVlanIDPanel, askBondModePanel, askNetworkMethodPanel, hostNamePanel)
 	}
-	return showNext(c, askInterfacePanel, askVlanIDPanel, askBondModePanel, askNetworkMethodPanel, addressPanel, gatewayPanel, hostNamePanel)
+	return showNext(c, askInterfacePanel, askVlanIDPanel, askBondModePanel, askNetworkMethodPanel, addressPanel, gatewayPanel, mtuPanel, hostNamePanel)
 }
 
 func addNetworkPanel(c *Console) error {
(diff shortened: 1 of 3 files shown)


pkg/console/install_panels.go
```diff
@@ -1162,28 +1170,72 @@ func addNetworkPanel(c *Console) error {
...
+		gocui.KeyEnter:     gatewayVConfirm,
+		gocui.KeyEsc:       gotoPrevPage,
+	}
+	setLocation(gatewayV.Panel, 3)
+	c.AddElement(gatewayPanel, gatewayV)
+
+	// mtuV
+	mtuV.PreShow = func() error {
+		c.Gui.Cursor = true
+		if mgmtNetwork.MTU == 0 {
```
bk201, line 1189: I'd suggest if the user doesn't specify any value, just leave it empty in the wicked configuration file and let wicked handle it.

starbops: Updated it, thanks!

bk201:
Two corner cases (note, the underlying wicked configs are correct, just need UI fixes).

**First**
- I can input a non-number string in the MTU field (like `abc`) and still go to the next page. It would be nice to detect it as well.

**Second**
- Configure 1300 MTU first and go to the next page.
- Go back network page again.
- Configure MTU with a blank value and go to the next page.
- Go back to the network page; the value is still 1300.

starbops:
> Two corner cases (note, the underlying wicked configs are correct, just need UI fixes).
> 
> **First**
> 
> * I can input a non-number string in the MTU field (like `abc`) and still go to the next page. It would be nice to detect it as well.
> 
> **Second**
> 
> * Configure 1300 MTU first and go to the next page.
> * Go back network page again.
> * Configure MTU with a blank value and go to the next page.
> * Go back to the network page; the value is still 1300.

Two corner cases are fixed. Please check it again, thanks!

PR #334, opened 2022-09-13
vlan: migrate harvester-mgmt to mgmt-br and mgmt-bo

tjjh89017:
Migrate harvester-mgmt to mgmt-br and mgmt-bo
It will be executed before cos-setup-initramfs.service

### DIFF
--- a/package/harvester-os/files/etc/systemd/system/cos-setup-initramfs.service.d/override.conf
+++ b/package/harvester-os/files/etc/systemd/system/cos-setup-initramfs.service.d/override.conf
@@ -1,0 +1,2 @@
+[Service]
+ExecStartPre=/usr/sbin/migration.sh
--- a/package/harvester-os/files/etc/systemd/system/cos-setup-initramfs.service.d/override.conf
+++ b/package/harvester-os/files/etc/systemd/system/cos-setup-initramfs.service.d/override.conf
@@ -1,2 +1,0 @@
-[Service]
-ExecStartPre=-/usr/sbin/migration.sh
--- a/package/harvester-os/files/etc/systemd/system/cos-setup-initramfs.service.d/override.conf
+++ b/package/harvester-os/files/etc/systemd/system/cos-setup-initramfs.service.d/override.conf
@@ -1,2 +1,2 @@
 [Service]
-ExecStartPre=/usr/sbin/migration.sh
+ExecStartPre=-/usr/sbin/migration.sh
--- a/package/harvester-os/files/usr/sbin/migration/migrate_harv_mgmt_to_mgmt_br.sh
+++ b/package/harvester-os/files/usr/sbin/migration/migrate_harv_mgmt_to_mgmt_br.sh
@@ -1,0 +1,161 @@
+#!/bin/bash
+
+HARV_CONFIG="${HARV_CONFIG:-/oem/99_custom.yaml}"
+HARV_MGMT="harvester-mgmt"
+IFCFG=
+IFROUTE=
+
+function migrate_mgmt_config () {
+	MODE=$1
+	echo "migrate $MODE config"
+
+	# search all files
+	for (( i=0; ; i++ ))
+	do
+		if ! yq -e ".stages.initramfs[0].files[$i]" "$HARV_CONFIG" > /dev/null 2>&1; then
+			echo "end files"
+			break
+		fi
+
+		if yq ".stages.initramfs[0].files[$i].path" "$HARV_CONFIG" | grep -q "/etc/sysconfig/network/ifcfg-${HARV_MGMT}"; then
+			# save config
+			IFCFG=$(yq ".stages.initramfs[0].files[$i].content" "$HARV_CONFIG")
+			echo "$IFCFG"
+		fi
+
+		if yq ".stages.initramfs[0].files[$i].path" "$HARV_CONFIG" | grep -q "/etc/sysconfig/network/ifroute-${HARV_MGMT}"; then
+			# save config
+			IFROUTE=$(yq ".stages.initramfs[0].files[$i].content" "$HARV_CONFIG")
+			echo "$IFROUTE"
+		fi
+	done
+
+	# start patch
+	# remove all ifcfg and ifroute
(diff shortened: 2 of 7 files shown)


bk201:
Will the system fail to boot if the exit status of `/usr/sbin/migration.sh` is not 0?

tjjh89017:
> Will the system fail to boot if the exit status of `/usr/sbin/migration.sh` is not 0?

It will fail boot if Pre command is fail.

tjjh89017:
Should I ignore the migration script error and force execute cos-setup-initramfs?

package/harvester-os/files/etc/systemd/system/cos-setup-initramfs.service.d/override.conf
```diff
@@ -0,0 +1,2 @@
+[Service]
+ExecStartPre=/usr/sbin/migration.sh
```
bk201, line 2: Prefix an `-` sign to ignore errors.
What's the consequence if the migration fails? I'd prefer still boot the system so users can fix it.

tjjh89017: Currently, we don't have a problem if the migration fails.
But if there is more script in the future.
We should consider dealing with it.

I will change it to ignore error in this stage

bk201:
It seems not to work. After upgrading the image and rebooting, the content in `/oem/` is still the same.
If I run the script `/usr/sbin/migration.sh` manually, the `/oem/99_custom.yaml` file does changes.

Please also back up the original file to a safe place.

bk201:
I also see a custom bonding device (`harvester-vlan`) not being migrated, is this expected?

The original configuration:
```
  networks:
    harvester-mgmt:       # (Mandatory) The management bond name.
      interfaces:
      - name: ens5
      default_route: true
      method: dhcp
      bond_options:
        mode: balance-tlb
        miimon: 100
    harvester-vlan:       # (Optional) The VLAN network bond name. If VLAN NIC names vary from
      interfaces:         # host to host, consider creating a bonding device. Users can then select
      - name: ens6        # `harvester-vlan` as the VLAN network NIC in the Harvester GUI.
      method: none
      bond_options:
        mode: balance-tlb
        miimon: 100
```

The original custom file:
```
    - path: /etc/sysconfig/network/ifcfg-harvester-vlan
      permissions: 384
      owner: 0
      group: 0
      content: |+
        STARTMODE='onboot'
        BONDING_MASTER='yes'
        BOOTPROTO='none'


        BONDING_SLAVE_0='ens6'

        BONDING_MODULE_OPTS='miimon=100 mode=balance-tlb '

        DHCLIENT_SET_DEFAULT_ROUTE='no'

      encoding: ""
      ownerstring: ""
```

package/harvester-os/files/usr/sbin/migration/migrate_harv_mgmt_to_mgmt_br.sh
```diff
@@ -0,0 +1,161 @@
...
+if ! grep -q "$HARV_MGMT" "$HARV_CONFIG"; then
+	echo "$HARV_MGMT not found. Skipping."
+	exit 0
+fi
+
+echo "$HARV_MGMT found."
+
+# check DHCP or static
+
+if grep -q "BOOTPROTO='static'" "$HARV_CONFIG"; then
```
bk201, line 156: The detection needs to be in `harvester-mgmt` file block only. Since the user might create more bonding devices and their `file` block also can contain the `BOOTPROTO=xxx` string.

tjjh89017:
> It seems not to work. After upgrading the image and rebooting, the content in `/oem/` is still the same. If I run the script `/usr/sbin/migration.sh` manually, the `/oem/99_custom.yaml` file does changes.
> 
> Please also back up the original file to a safe place.

Let me check it

> I also see a custom bonding device (harvester-vlan) not being migrated, is this expected?

This is expected.
This part should be new VLAN config.

tjjh89017:
- [x] backup script before patching
- [x] it will be executed before cos-setup-initramfs.service.

package/harvester-os/files/usr/sbin/migration/migrate_harv_mgmt_to_mgmt_br.sh
tjjh89017: fixed. only detect harvester-mgmt's type now.

yaocw2020:
Hope to be merged after 1.1.0 rc1 and before 1.1.0 rc2.

package/harvester-os/files/usr/sbin/migration/migrate_harv_mgmt_to_mgmt_br.sh
```diff
@@ -0,0 +1,166 @@
...
+	MODE=
+	# search all files
+	for (( i=0; ; i++ ))
+	do
+		if ! yq -e ".stages.initramfs[0].files[$i]" "$HARV_CONFIG" > /dev/null 2>&1; then
+			echo "end files"
+			break
+		fi
+
+		if yq ".stages.initramfs[0].files[$i].path" "$HARV_CONFIG" | grep -q "/etc/sysconfig/network/ifcfg-${HARV_MGMT}"; then
```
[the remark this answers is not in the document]
tjjh89017: Our /oem/99_custom.yaml will be generated from the installer result.
In installer config.yaml, yaml object "key" will be the interface name and file name also.
If we have several interface in installer config.yaml, it will not have the same entry.
So `/etc/sysconfig/network/ifcfg-${HARV_MGMT}` will be one and only.

If users add some files in the /oem/99_custom.yaml with the same filename, we could only use the first file to migrate the config.
And in our guide, we always ask users to modify the original files rather than adding a new file to replace

Vicente-Cheng (commented):
The installer part looked good to me.
But we still need to test with the VLAN part on the harvester side.

If the test passed (as I mentioned above), it would be good to merge.

starbops (approved):
LGTM, thanks!

Vicente-Cheng (approved):
LGTM! Also tested again with the latest VLAN upgrade script.

PR #341, opened 2022-09-21
Rewrite the scheme version in automatic/non-automatic mode

tjjh89017:
Rewrite the scheme version in automatic/non-automatic mode
Only assign Scheme Version in non-automatic mode 
and load scheme version from config.yaml directly in automatic mode.


tjjh89017:
Tested with ipxe-examples.
SchemeVersion!=1 will fail.
Currently, it works as expected behavior.

bk201 (approved):
LGTM!

PR #342, opened 2022-09-21
Bump RKE2 to v1.23.10+rke2r1

bk201:
Related issue: https://github.com/harvester/harvester/issues/2785


Vicente-Cheng (approved):
Tested with basic operations like creating VM/Volume/Image, attach/detach Volume, and VM migration.
Seems to work well, thanks!

guangbochen (approved):

PR #345, opened 2022-09-27
managedChart: add ignoring resource for modified longhorn component

Vicente-Cheng:
related to https://github.com/harvester/harvester/issues/2762

Longhorn add some new resources that would cause the fleet agent to label the modified flag after deploying bundle.

To resolve this problem, we need to ignore this newly resource by `jsonPointers`.
reference: https://fleet.rancher.io/troubleshooting/#gitrepo-or-bundle-stuck-in-modified-state


bk201:
If we can eliminate the modified state, then we can change the wait to "ready": https://github.com/harvester/harvester/blob/af5215bbd15a7002a0485016daf2a5ec868252a4/package/upgrade/upgrade_manifests.sh#L300

bk201:
@Vicente-Cheng We also need to patch the Harvester chart value from v1.0.3: https://github.com/harvester/harvester/blob/master/package/upgrade/migrations/managed_charts/v1.0.3.sh

For example https://github.com/harvester/harvester/blob/b019cd9010988e2bf1cd9822b13934b9807cf706/package/upgrade/migrations/managed_charts/v1.0.0.sh#L29-L31

pkg/config/templates/rancherd-10-harvester.yaml
```diff
@@ -76,6 +76,12 @@ resources:
     releaseName: harvester
     version: {{ .HarvesterChartVersion }}
     defaultNamespace: harvester-system
+    diff:
```
w13915984028, line 79: The `diff` in `rancherd-13-monitoring.yaml` add the specific leaf objects,  here it seems skips all `status` of all `objects` under `harvester` managedchart, is the range too big ?

image

bk201: yes, we should include a CRD name here. Good spot.

w13915984028: Is the new fix verified?

I took a further look, and updated https://github.com/harvester/harvester/issues/2762.

It looks, the difference is, CRD `engineimages.longhorn.io` (and another 2) has generation `2` after system installation, but normal CRD has generation 1, that's why fleet-agent complain.

I did not find a clear clue, why the CRD (installed by helm) itself is updated right at the installation stage.  

If the fix makes `fleet-agent` happy, we can go ahead now.

image

bk201 (approved):
LGTM

pkg/config/templates/rancherd-10-harvester.yaml
Vicente-Cheng: yap, this is a new fix.
I thought that the root cause is like what you found. Also explain on [here](https://github.com/rancher/fleet/blob/master/docs/bundle-diffs.md)

> The bundled charts may have some objects that are amended at runtime.
> This leads the status of the bundle and associated GitRepo to be reported as "Modified"

>Fleet bundles support the ability to specify a custom [jsonPointer patch](http://jsonpatch.com/).
>With the patch, users can instruct fleet to ignore object modifications.

So that is what we want, instruct the fleet agent to ignore the modification at runtime.

I test with the running cluster and a new cluster. Looks like works well :D

w13915984028 (approved):
LGTM, well done,  thanks.

PR #350, opened 2022-09-30
CI: Use main branch for ipxe-examples

bk201:
We've made the v1.0 branch as the default branch in ipxe-example, and it doesn't work with the installer's master branch.
Use the `main` branch instead.


Vicente-Cheng (approved):
thanks for fixing CI error!

PR #351, opened 2022-10-05
Bump OS to 20221005

bk201:
Include rancherd update: https://github.com/rancher/rancherd/pull/25.


Vicente-Cheng (approved):
verify with some basic operations like creating Image/Volume/VM, attach/detach volume, and VMs migrations.

LGTM

PR #352, opened 2022-10-10
bump up image for harvester-vm-import-controller

ibrokethecloud:
The PR introduces a new version of the harvester-vm-import-controller, which addresses a number of issues fixed by PR: https://github.com/harvester/vm-import-controller/pull/5

### DIFF
--- a/pkg/config/templates/rancherd-22-vm-import-controller.yaml
+++ b/pkg/config/templates/rancherd-22-vm-import-controller.yaml
@@ -12,4 +12,5 @@ resources:
       valuesContent: |
         image:
           tag: v0.1.1
+        fullnameOverride: harvester-vm-import-controller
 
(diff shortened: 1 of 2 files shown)


bk201 (approved):
LGTM!

pkg/config/templates/rancherd-22-vm-import-controller.yaml
```diff
@@ -11,5 +11,6 @@ resources:
       enabled: false
       valuesContent: |
         image:
-          tag: v0.1.0
+          tag: v0.1.1
+          fullnameOverride: harvester-vm-import-controller
```
bk201, line 15: Sorry, just found `fullnameOverride` should not be in `image`. It's on the top level.

ibrokethecloud: 👍

PR #353, opened 2022-10-12
package pcidevices addon

ibrokethecloud:
PR introduces changes to package up pcidevices-controller as an addon in the harvester-installer

### DIFF
--- a/pkg/config/cos.go
+++ b/pkg/config/cos.go
@@ -595,7 +595,7 @@ func genBootstrapResources(config *HarvesterConfig) (map[string]string, error) {
 
 	for _, templateName := range []string{
 		"12-monitoring-dashboard.yaml",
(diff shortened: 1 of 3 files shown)


ibrokethecloud:
Related to: https://github.com/harvester/harvester/issues/2913

bk201 (approved):
Screen Shot 2022-10-12 at 5 29 50 PM
LGTM!
Right now the image tag and chart version are coded in many places (including the upgrade path) and we might forget one of them when bumping versions.
We can factor later to see if these versions can be controlled in a single file.

guangbochen (approved):
LGTM, thanks for the changes.

Vicente-Cheng (approved):
Tested with the upgrade!
The AddOn works well.

LGTM!

PR #358, opened 2022-10-14
Workaround: patch rke2-canal chart in rke2-runtime image

bk201:


guangbochen (approved):
LGTM, thanks.

PR #364, opened 2022-10-18
Add iommu=pt for PCI passthrough

tlehman:
This adds a parameter to the grub bootloader which boosts performance of PCI devices enabled for passthrough by removing the need to do DMA translation.


tjjh89017 (approved):
LGTM

PR #373, opened 2022-10-25
ntp: rancher, rke2-server rke2-agent need to wait for time sync

tjjh89017:
use time-sync.target
wait for 3 min to sync time
if time sync took more 3 min, skip it and start rancherd, rke2-server and rke2-agent directly.

issue: https://github.com/harvester/harvester/issues/2860

### DIFF
--- a/package/harvester-os/files/etc/systemd/system/systemd-time-wait-sync.service.d/override.conf
+++ b/package/harvester-os/files/etc/systemd/system/systemd-time-wait-sync.service.d/override.conf
@@ -1,0 +1,2 @@
+[Service]
+TimeoutStartSec=180
--- a/package/harvester-os/files/system/oem/06_ntp.yaml
+++ b/package/harvester-os/files/system/oem/06_ntp.yaml
@@ -1,7 +1,0 @@
-name: "Enable Systemd Time Wait Sync"
-stages:
-  initramfs:
-    - name: "enable systemd-time-wait-sync.service"
-      if: 'grep -q root=LABEL=COS_ACTIVE /proc/cmdline && [ -n "$(blkid -L COS_ACTIVE)" ]'
-      commands:
-        - systemctl enable systemd-time-wait-sync.service
--- a/package/harvester-os/files/system/oem/06_ntp.yaml
+++ b/package/harvester-os/files/system/oem/06_ntp.yaml
@@ -1,0 +1,7 @@
+name: "Enable Systemd Time Wait Sync"
+stages:
+  initramfs:
+    - name: "enable systemd-time-wait-sync.service"
+      if: 'grep -q root=LABEL=COS_ACTIVE /proc/cmdline && [ -n "$(blkid -L COS_ACTIVE)" ]'
+      commands:
+        - systemctl enable systemd-time-wait-sync.service
--- a/package/harvester-os/files/etc/systemd/system/rancherd.service.d/override.conf
+++ b/package/harvester-os/files/etc/systemd/system/rancherd.service.d/override.conf
@@ -1,0 +1,3 @@
+[Unit]
+After=time-sync.target
+Wants=time-sync.target
--- a/package/harvester-os/files/etc/systemd/system/rke2-agent.service.d/override.conf
+++ b/package/harvester-os/files/etc/systemd/system/rke2-agent.service.d/override.conf
@@ -1,2 +1,6 @@
+[Unit]
+After=time-sync.target
+Wants=time-sync.target
+
 [Service]
 ExecStartPre=/usr/sbin/harv-update-rke2-server-url agent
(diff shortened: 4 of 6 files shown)


Vicente-Cheng:
Just curious, does the 180 seconds is enough for time sync?

Could we have some way to check the time is synced with the NTP server?

tjjh89017:
> Just curious, does the 180 seconds is enough for time sync?
> 
> Could we have some way to check the time is synced with the NTP server?

systemd-time-wait-sync.service will help us determine whether the time is synced.
But it cannot determine the NTP server is unreachable or time is still syncing.
So I setup 3 min timeout.
If NTP server is reachable, it will cost less than 1min to sync in my testing env.

package/harvester-os/files/etc/systemd/system/systemd-time-wait-sync.service.d/override.conf
```diff
@@ -0,0 +1,2 @@
+[Service]
+TimeoutStartSec=180
```
Vicente-Cheng, line 2: Maybe we should make it be `oneshot` service to prevent the service restart dependency?
Also, could we add `Description ` in the [Unit] section to clarify this service? Thanks!

Vicente-Cheng:
> > Just curious, does the 180 seconds is enough for time sync?
> > Could we have some way to check the time is synced with the NTP server?
> 
> systemd-time-wait-sync.service will help us determine whether the time is synced. But it cannot determine the NTP server is unreachable or time is still syncing. So I setup 3 min timeout. If NTP server is reachable, it will cost less than 1min to sync in my testing env.

OK, that makes sense to me. Just some nit on the following comments.

package/harvester-os/files/etc/systemd/system/systemd-time-wait-sync.service.d/override.conf
tjjh89017: We should not modify systemd service to `oneshot` to make sure the behavior.
and `/lib/systemd/system/systemd-time-wait-sync.service` has original `Description` that we should not override it.
I will consider to keep those as original systemd services as possible.
and users could check the systemd docs to know the behavior

tjjh89017: I just found `systemd-time-wait-sync.service` is already an oneshot service.

```
[Service]
Type=oneshot
ExecStart=/lib/systemd/systemd-time-wait-sync
TimeoutStartSec=infinity
RemainAfterExit=yes
```

Vicente-Cheng: Ah... I missing something. That is already a oneshot service. That's no question here.

tjjh89017: Thanks for this detail!

Vicente-Cheng: The dependency issues could happen because serval services depend on each other. Once you restart someone, another will also be restarted.

tjjh89017: > Once you restart someone, another will also be restarted
IIRC, systemd didn't do that until service has "PartOf" attribute.

tjjh89017: and oneshot service only said it is an one time command which is not a continuous process.
It's almost the same as "simple" service.

tjjh89017: And I used "Wants" to make sure the time sync is better to have but not requirement.
So after timeout, time sync service will be mark as failed.
and rke2-server/agent will ignore that service.

Vicente-Cheng (approved):
LGTM, test fresh installation with #375. Thanks!

package/harvester-os/files/system/oem/06_ntp.yaml
```diff
@@ -0,0 +1,7 @@
+name: "Enable Systemd Time Wait Sync"
+stages:
+  initramfs:
+    - name: "enable systemd-time-wait-sync.service"
```
bk201, line 4: If the user doesn't specify NTP servers, services still need to wait 3 minutes to start.
Could we detect if the user set NTP servers first and then overridden systemd units to avoid the wait?

tjjh89017: Let me check about this one.

tjjh89017: I enable time sync wait service if and only NTP server config is existing.

starbops (approved):
LGTM and tested with a fresh install in a virtualized environment. Checked with the following command to see if the services are actually waiting for the time synced.

```
journalctl -b -u systemd-timesyncd.service -u systemd-time-wait-sync.service -u time-sync.target -u rancherd.service -u rke2-server.service -u rke2-agent.service
```

Thanks!

bk201 (approved):
LGTM

bk201:
Please help backport the PR to the v1.1 branch, thanks.

PR #376, opened 2022-11-03
added some missing harvester settings

joeyloman:
This PR contains some missing harvester settings in the GetSystemSettingsAllowList since Harvester v1.1.0. We need the containerd-registry setting during installation otherwise pulling the rancher-agent will fail when registering the Harvester cluster in Rancher in an air gapped environment.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -64,10 +64,7 @@ func GetSystemSettingsAllowList() []string {
 		"overcommit-config",
 		"vip-pools",
 		"auto-disk-provision-paths",
-		"csi-driver-config",
 		"containerd-registry",
-		"storage-network",
-		"harvester-csi-ccm-versions",
 	}
 }
 
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -64,6 +64,10 @@ func GetSystemSettingsAllowList() []string {
 		"overcommit-config",
 		"vip-pools",
 		"auto-disk-provision-paths",
+		"csi-driver-config",
+		"containerd-registry",
+		"storage-network",
+		"harvester-csi-ccm-versions",
 	}
 }
 


bk201:
@guangbochen Should we could include this in v1.1.1?

pkg/config/config.go
```diff
@@ -64,6 +64,10 @@ func GetSystemSettingsAllowList() []string {
 		"overcommit-config",
 		"vip-pools",
 		"auto-disk-provision-paths",
+		"csi-driver-config",
+		"containerd-registry",
+		"storage-network",
```
bk201, line 69: @tjjh89017 Could you check if it's a good idea to allow the storage-network setting when bootstrapping a cluster? Thanks!

pkg/config/config.go
```diff
@@ -64,6 +64,10 @@ func GetSystemSettingsAllowList() []string {
...
+		"harvester-csi-ccm-versions",
```
guangbochen, line 70: don't include the `csi-driver-config` and `harvester-csi-ccm-versions` as they are preserved as system configuration.

pkg/config/config.go
tjjh89017: Please don't include storage-network.
That setting is day2 setting.

bk201:
>>otherwise pulling the rancher-agent will fail when registering the Harvester cluster in Rancher in an air gapped environment."

@joeyloman Just curious about this part. We should include all needed images in the ISO. Did you see any missing images in the air-gapped env?

pkg/config/config.go
joeyloman: Yes I understand and I will remove it. Shouldn't that also be the case for cluster_networks? Because I can still put that in the create-config and it's created after a deployment. However without any nics configured because the install.networks part is removed. Or is this still there for upgrade purposes?

pkg/config/config.go
joeyloman: Ok. I synced them because all the other options were there as well. But I understand that we don't need to add all of them. I will remove them.

joeyloman:
@bk201 I use Rancher 2.6.9. When Harvester downloads and applies the cluster-registration-url objects it wants to pull rancher/rancher-agent:v2.6.9. This results in a ImagePullBackOff which means that it's not included in the ISO right?

guangbochen (approved):
LGTM, please help to add the related issue for it so QA can help to validate it properly, thanks.

bk201 (approved):
LGTM!

joeyloman:
> LGTM, please help to add the related issue for it so QA can help to validate it properly, thanks.

See: https://github.com/harvester/harvester/issues/3116

PR #377, opened 2022-11-07
Move hostname to a dedicated page before network configuration to avoid crashes.

masteryyh:
**Related issues:**
[#2937](https://github.com/harvester/harvester/issues/2937)
[#3112](https://github.com/harvester/harvester/issues/3112)

**Test plan:**
1. Prepare a small screen/low resolution environment, use ISO to boot into installer;
2. Go to `Hostname` page and set a hostname;
3. Go to `Network configuration` page and use a static IPv4 config to configure network;
4. Leave `MTU` empty, when focusing on `MTU` input box, press up arrow key;
5. See if installer crashed.

**Pictures:**
截屏2022-11-08 10 40 33
截屏2022-11-08 10 41 01


pkg/console/install_panels.go
```diff
@@ -1293,7 +1345,13 @@ func addNetworkPanel(c *Console) error {
 			if err != nil {
 				return "", err
 			}
-			mgmtNetwork.MTU, err = strconv.Atoi(mtu)
+
+			if mtu != "" {
+				mgmtNetwork.MTU, err = strconv.Atoi(mtu)
+			} else {
+				mgmtNetwork.MTU = 0
```
guangbochen, line 1352: should leave it unset instead of `0`

yaocw2020: Please leave it unset and let it as the default value.

pkg/console/install_panels.go
```diff
@@ -1293,7 +1345,11 @@ func addNetworkPanel(c *Console) error {
 			if err != nil {
 				return "", err
 			}
-			mgmtNetwork.MTU, err = strconv.Atoi(mtu)
+
+			if mtu != "" {
+				mgmtNetwork.MTU, err = strconv.Atoi(mtu)
+			}
```
bk201, line 1351: Looks like if I type an "a" here and press up, it will still crash to the first panel.
Perhaps it's better to work like press enter (an error message is displayed).

guangbochen: Additionally, the default value will be gone, not sure if this is a problem in my env.

starbops (approved):
Tested with various inputs, and finally left with empty value then do the install. The default value, which is 1500, is respected.

LGTM, thanks!

bk201 (approved):
LGTM

guangbochen:
Got some input errors as attached below, can we PTAL, thanks.
https://user-images.githubusercontent.com/4569037/200749923-b293cce6-9f30-40a0-ad27-aaa5c49039e1.mp4

yaocw2020:
The text here is not accurate. 
image

yaocw2020 (approved):

guangbochen (approved):
LGTM, thanks.

bk201 (approved):

PR #383, opened 2022-11-23
addition of iommu arguments to kernel boot args

ibrokethecloud:
PR superseedes: https://github.com/harvester/harvester-installer/pull/380

It fixes issue: https://github.com/harvester/harvester/issues/2973 by moving the iommu argument definition out of harv-installer and into bootargs.cfg.

This will ensure args are applied during both install and upgrade.

### DIFF
--- a/package/harvester-os/files/etc/cos/bootargs.cfg
+++ b/package/harvester-os/files/etc/cos/bootargs.cfg
@@ -2,7 +2,7 @@ set console_params="console=tty1"
 set kernel=/boot/vmlinuz
 set crash_kernel_params="crashkernel=219M,high crashkernel=72M,low"
 if [ -n "$recoverylabel" ]; then
-    set kernelcmd="$console_params root=live:LABEL=$recoverylabel rd.live.dir=/ rd.live.squashimg=$img panic=0 net.ifnames=1 rd.cos.oemtimeout=120 intel_iommu=on amd_iommu=on iommu=pt"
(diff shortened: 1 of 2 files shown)


ttpcodes:
I believe Tobi tested this and determined that if one of the flags doesn't apply, it's simply a no-op.

bk201 (approved):
LGTM

guangbochen (approved):
LGTM and verified that the upgrade path from v1.0.3 can add the IOMMU config correctly.
```
dmesg | grep DMAR
[    0.151960] DMAR: IOMMU enabled
```
``` shell
cat /etc/cos/bootargs.cfg
set console_params="console=tty1"
set kernel=/boot/vmlinuz
set crash_kernel_params="crashkernel=219M,high crashkernel=72M,low"
if [ -n "$recoverylabel" ]; then
    set kernelcmd="$console_params root=live:LABEL=$recoverylabel rd.live.dir=/ rd.live.squashimg=$img panic=0 net.ifnames=1 rd.cos.oemtimeout=120"
else
    set kernelcmd="$console_params root=LABEL=$label cos-img/filename=$img panic=0 net.ifnames=1 rd.cos.oemtimeout=120 rd.cos.oemlabel=COS_OEM audit=1 audit_backlog_limit=8192 intel_iommu=on amd_iommu=on iommu=pt"
fi

set initramfs=/boot/initrd
```

ttpcodes (commented):
LGTM.

PR #388, opened 2022-12-02
Allow server_url without port and https prefix.

masteryyh:
**Related issue**
[#2737](https://github.com/harvester/harvester/issues/2737)

**Test steps**
1. Build an ISO with this branch;
2. Use the artifacts built with this branch to construct a PXE server, configure a `join` configuration and use a server URL without port, like `https://172.16.0.240`;
3. Use the PXE server to boot up a node and start automatic installation;
4. Complete the installation and see if node successfully joined cluster;
5. Get into console, run `yq '.serverurl' < /oem/harvester.config` to see if server URL is correctly configured.

Signed-off-by: Yanhong Yang <yanhong.yang@suse.com>

### DIFF
--- a/pkg/console/util.go
+++ b/pkg/console/util.go
@@ -43,6 +43,7 @@ You can see the full installation log by:
   - Login with user "rancher" (password is "rancher").
   - Run the command: less %s.
 `
+	https = "https://"
 )
 
 func newProxyClient() http.Client {
@@ -277,12 +278,37 @@ func getRemoteSSHKeys(url string) ([]string, error) {
 }
 
 func getFormattedServerURL(addr string) (string, error) {
-	ipErr := checkIP(addr)
-	domainErr := checkDomain(addr)
-	if ipErr != nil && domainErr != nil {
+	if addr == "" {
+		return "", errors.New("management address cannot be empty")
+	}
+	addr = strings.TrimSpace(addr)
+
+	realAddr := addr
+	if !strings.HasPrefix(addr, https) {
+		realAddr = https + addr
+	}
+	parsedUrl, err := url.ParseRequestURI(realAddr)
+	if err != nil {
+		return "", fmt.Errorf("%s is invalid", addr)
+	}
+
+	host := parsedUrl.Hostname()
+	if checkIP(host) != nil && checkDomain(host) != nil {
 		return "", fmt.Errorf("%s is not a valid ip/domain", addr)
 	}
-	return fmt.Sprintf("https://%s:%s", addr, rancherManagementPort), nil
+
+	if parsedUrl.Path != "" {
+		return "", fmt.Errorf("path is not allowed in management address: %s", parsedUrl.Path)
+	}
+
+	port := parsedUrl.Port()
+	if port == "" {
+		parsedUrl.Host += ":443"
+	} else if port != "443" {
+		return "", fmt.Errorf("currently non-443 port are not allowed")
+	}
+
+	return parsedUrl.String(), nil
 }
 
 func getServerURLFromRancherdConfig(data []byte) (string, error) {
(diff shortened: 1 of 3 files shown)


pkg/console/util.go
```diff
@@ -277,12 +278,31 @@ func getRemoteSSHKeys(url string) ([]string, error) {
 }
 
 func getFormattedServerURL(addr string) (string, error) {
-	ipErr := checkIP(addr)
-	domainErr := checkDomain(addr)
+	realAddr := addr
+	port := rancherManagementPort
+
```
bk201, line 283: Just wonder if we can leverage the package: https://pkg.go.dev/net/url

pkg/console/util.go
```diff
@@ -277,12 +278,31 @@ func getRemoteSSHKeys(url string) ([]string, error) {
...
+	if strings.HasPrefix(realAddr, https) {
+		realAddr = strings.TrimPrefix(realAddr, "https://")
+	}
+	if strings.Contains(realAddr, ":") {
+		splitAddr := strings.Split(realAddr, ":")
+		realAddr = splitAddr[0]
+		port = splitAddr[1]
```
futuretea, line 290: will panic when addr == ""

Vicente-Cheng (approved):
LGTM, thanks!

pkg/console/util_test.go
```diff
@@ -79,17 +79,11 @@ func TestGetFormattedServerURL(t *testing.T) {
 			output: "https://example.org:443",
 			err:    nil,
 		},
-		{
-			Name:   "invalid ip",
-			input:  "1.2.3.4/",
```
futuretea, line 84: Does RKE2 support this format？

masteryyh: It's a silly mistake for not checking path in url, added now.

pkg/console/util.go
```diff
@@ -277,12 +278,35 @@ func getRemoteSSHKeys(url string) ([]string, error) {
...
-	return fmt.Sprintf("https://%s:%s", addr, rancherManagementPort), nil
+
+	if parsedUrl.Path != "" {
+		return "", fmt.Errorf("path is not allowed in management address: %s", parsedUrl.Path)
+	}
+
+	port := parsedUrl.Port()
+	if port == "" {
+		parsedUrl.Host += ":443"
+	}
```
futuretea, line 307:
```go
else if port != "443" {

}
```

Does RKE2 support custom port?
Refer to the logs in https://github.com/harvester/harvester/issues/2737#issue-1362758670
It seems that RKE2 only support 8443 and 443.
And for Harvester, we only use 443.

masteryyh: So might be better to add a port verification if not using 443 or 8443?

futuretea: just need make sure the port is 443

futuretea: 8443 was close by Harvester for security

guangbochen (approved):
LGTM, thanks.

futuretea (approved):
LGTM
```yaml
scheme_version: 1
server_url: https://192.168.5.40
token: token             # Replace with a desired token
os:
  hostname: harv-dev-node2
  password: password     # Replace with a your password
  ntp_servers:
  - ntp.aliyun.com
  dns_nameservers:
  - 192.168.5.1
install:
  mode: join
  management_interface:       # The management bond name. This is mandatory.
    interfaces:
    - name: ens33
    default_route: true
    method: dhcp
    bond_options:
      mode: balance-tlb
      miimon: 100
  device: /dev/sda
  iso_url: http://192.168.5.79/_/harvester-pxe/artifacts/master/harvester-amd64.iso
```
image
image

image

PR #389, opened 2022-12-05
Set max-pods to 200

tlehman:
Fixes https://github.com/harvester/harvester/issues/2707


ibrokethecloud:
Should we open up the label process to look for a special label named `kubelet-args` which can then be passed to the actual kubelet? This should make it easier to customize nodes even more, not just the max pod arguments.

bk201:
Good work, We also need to take care of the upgrade path from previous versions. (change in harvester/harvester).

tlehman:
Closed, this HEP will supercede this: https://github.com/harvester/harvester/pull/3558

PR #397, opened 2022-12-29
Bump RKE2 to v1.24.9+rke2r1

bk201:
Related issue: https://github.com/harvester/harvester/issues/3321

Test bump. We need to use the 2023 Jan or Feb release.

### DIFF
--- a/scripts/version-rke2
+++ b/scripts/version-rke2
@@ -1,1 +1,1 @@
-RKE2_VERSION="v1.24.7+rke2r1"
+RKE2_VERSION="v1.24.9+rke2r1"


bk201:
Need to skip v1.24.9+rke2r1
https://github.com/rancher/rke2/releases/tag/v1.24.9%2Brke2r1
> ⚠️ WARNING
This release is affected by https://github.com/containerd/containerd/issues/7843, which causes the kubelet to restart all pods whenever RKE2 is restarted. For this reason, we have removed this RKE2 release from the channel server. Please use v1.24.9+rke2r2 instead.

PR #399, opened 2023-01-06
ci: add basic operations testing on CI

Vicente-Cheng:


Vicente-Cheng:
Hi @guangbochen, @bk201:
This installer CI is ready to review.
This change may do some simple operations by terraform, described as follow:
- create image (cirros)
- create new volume
- create VM from the above image and try to attach the above new volume
Then check the network after the above operation done.

We choose the cirros instead ubuntu because the ubuntu base image is too large. Sometimes download would be timeout.
We could improve this part as the next item.

ci/terraform/virtualmachines.tf
```diff
@@ -0,0 +1,41 @@
+resource "harvester_virtualmachine" "cirros-01" {
+  name                 = "cirros"
+  namespace            = "default"
+  restart_after_update = true
+
+  description = "test vm on cirros"
+
+  cpu    = 2
+  memory = "2Gi"
```
bk201, line 9: I'd suggest reducing it to 512 MB since it's a nested VM.

ci/run_vagrant_install_test.yml
```diff
@@ -68,6 +68,22 @@
...
+         type: rsa
+         size: 3072
+         state: present
+         force: no
+
+    - name: Set SSH Key
+      vars:
+        generated_ssh_key: "{{ lookup('file', '{{ WORKSPACE }}/harvester-installer/ci/terraform/tmp-ssh-key.pub') }}"
+        settings_file: "{{ WORKSPACE }}/ipxe-examples/vagrant-pxe-harvester/settings.yml"
+      shell: yq e -i ".harvester_config.ssh_authorized_keys = [\"{{ generated_ssh_key }}\"]" "{{ settings_file }}"
```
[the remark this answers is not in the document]
Vicente-Cheng: Sure, that makes sense.
Updated!

ci/terraform/virtualmachines.tf
Vicente-Cheng: Ah, I agree. Thanks for the reminder!!
updated!

bk201 (approved):
LGTM

ci/run_vagrant_install_test.yml
```diff
@@ -104,6 +120,38 @@
...
+        ./cleanup_test_files.sh
+      args:
+        chdir: "{{ WORKSPACE }}/harvester-installer/ci/terraform"
+
+    - name: Testing basic operations with terraform
+      register: terraform_test
+      shell: |
+        curl https://releases.hashicorp.com/terraform/1.3.7/terraform_1.3.7_linux_amd64.zip -o terraform_bin.zip
+        unzip -o terraform_bin.zip
+        ./get_kubeconfig.sh 192.168.0.30 192.168.0.131
```
guangbochen, line 134: Can we get it via the env variable? also need to doc where the node exists.

ci/terraform/volumes.tf
```diff
@@ -0,0 +1,6 @@
+resource "harvester_volume" "vol001" {
```
futuretea, line 1: It seems not be attached

Vicente-Cheng: Thanks for the reminder!
I update the VM.tf. Please help to recheck. Thanks!

ci/run_vagrant_install_test.yml
Vicente-Cheng: Sure. This IP/VIP is from the `settings.yml` on the `ipxe-example` repo.
I change the way to load it from the above file. WDUT?

ci/terraform/provider.tf
```diff
@@ -0,0 +1,13 @@
+terraform {
+  required_version = ">= 0.13"
+  required_providers {
+    harvester = {
+      source  = "harvester/harvester"
+      version = "0.5.3"
```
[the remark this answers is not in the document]
Vicente-Cheng: fixed!

guangbochen (approved):
LGTM, thanks.

PR #400, opened 2023-01-09
[WIP] Airgap Rancher Airgap Harvester Ansible Driver

irishgordo:


irishgordo:
Converted back to draft as:
https://github.com/harvester/harvester-installer/pull/618
Was merged in - this should be re-adjusted or reworked.

We're continuing to encounter some "occasional" issues with:
- resources on the node running the job is inadequate, running 4 Harvester Nodes Virtualized w/ Docker-Registry, Rancher K3s, and other elements all virtualized as well 'airgapped' on a single node, will occasionally just fail due to resources on the machine not being enough
- acquisition of Rancher dependent images occasionally just 'fail', most commonly `jetstack` based images, since docker-registry gets rebuilt and not persisted between runs
- Rancher will occasionally just time out waiting for rollout to finish, retry logic is not geared for exponential backoff currently, but failures seem to be associated again with resource issues on the node

PR #415, opened 2023-02-10
Introduce two images for upgrade o11y enhancement

starbops:
Signed-off-by: Zespre Chang <zespre.chang@suse.com>

This change will make the ISO image build process to include two new container images:

- Upgrade log downloader
- Upgrade log packager

Related issue: harvester/harvester#1926
Related PR: harvester/harvester#3379


starbops:
Close this PR because the two introduced images are not needed, use `harvester-upgrade` image instead.

PR #416, opened 2023-02-13
fix ci error

Vicente-Cheng:
Just rollback to 5.2 images to verify the status.


Vicente-Cheng:
close due to https://github.com/harvester/harvester-installer/pull/418

PR #418, opened 2023-02-14
CI: Enable software-emulation for nested VMs

bk201:
Enable soft-emulation, the new kernel seems not to play well with the Equinix EPYC instance.
- Set the configuration in KubeVirt CR. Do this in VM to avoid downloading kubectl.
```
spec:
  configuration:
    developerConfiguration:
      useEmulation: true
```
- Update pinging code to a checking loop (timeout 5 minutes).

Related issue: https://github.com/harvester/harvester/issues/3447

### DIFF
--- a/ci/run_vagrant_install_test.yml
+++ b/ci/run_vagrant_install_test.yml
@@ -120,6 +120,19 @@
       debug:
         msg: "{{ setup_harvester_result.stdout_lines }}"
 
+    - name: Enable software-emulation in KubeVirt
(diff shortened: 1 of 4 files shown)


Vicente-Cheng (approved):
LGTM, thanks for this fixes

guangbochen (approved):
LGTM, thanks.
Additionally, it would be worth mentioning that we may need to consider allowing to pass overwritten `values.yml` to the Harvester chart.

PR #428, opened 2023-02-21
installer: add parameter to disable installer

Vicente-Cheng:
**Problem:**
Add the parameter to disable the installer

**Solution:**
add `harvester.installer.disable` to control the installer setup

**Related Issue:**
https://github.com/harvester/harvester/issues/2663

**Test plan:**
Test the following three cases:
1. Do not change /proc/cmdline -> should go into installer
2. Add `harvester.installer.disable=false` to /proc/cmdline -> should go into installer
3. Add `harvester.installer.disable=true` to /proc/cmdline -> should not go into installer 

How to modify `/proc/cmdline`, you can refer the following steps:
1. when you enter the grub, press `esc` to stop on this console
esc_to_stop
2. then press `e` to modify `/proc/cmdline`, the following picture is original
original
3. modify the `/proc/cmdline` like the above test cases
after
4. finally, press `ctrl + x` to bootup, then you could verify the above test cases

### DIFF
--- a/package/harvester-os/files/system/oem/91_installer.yaml
+++ b/package/harvester-os/files/system/oem/91_installer.yaml
@@ -3,6 +3,7 @@ stages:
   boot:
     - commands:
       - setup-installer.sh
+      if: '! grep -q "harvester.installer.disable=true" /proc/cmdline'
   initramfs:
     - environment_file: "/etc/rancher/installer/env"
       environment:


bk201 (approved):
lgtm!

guangbochen (approved):
LGTM, and we need to add this to the doc https://docs.harvesterhci.io/v1.1/troubleshooting/installation

Vicente-Cheng:
> LGTM, and we need to add this to the doc https://docs.harvesterhci.io/v1.1/troubleshooting/installation

OK, I would do it. Do we need another issue for it (document improvement)?

package/harvester-os/files/system/oem/91_installer.yaml
```diff
@@ -3,6 +3,7 @@ stages:
   boot:
     - commands:
       - setup-installer.sh
+      if: '[ ! -n "$(grep "harvester.installer.disable" /proc/cmdline)" ] || ! grep -q "harvester.installer.disable=true" /proc/cmdline'
```
starbops, line 6: Seems it can achieve the same results with only the second part. Is there any concern to leave `[ ! -n "$(grep "harvester.installer.disable" /proc/cmdline)" ]` here? Thanks!

Vicente-Cheng: Nice catch! Because I changed the if condition, the first part could be dropped!

Vicente-Cheng: Updated! @starbops, please recheck it.

starbops (approved):
LGTM, thanks!

masteryyh (approved):
LGTM, thanks.

bk201:
@mergifyio backport v1.1

hernad:
As I can see,  my patch from #2663 is considered usefull, accepted and merged. I am aware this is trivial enhancement, but It should be registered somewhere in source code as my contribution @Vicente-Cheng, @yasker. I haven't notice that in this commit.

Vicente-Cheng:
Hi @hernad, 
I added the comment for this enhancement. Please check #437.
Thanks for your enhancement!

PR #431, opened 2023-02-23
Speedup image pulling process

bk201:
Add the support to skip image pulling if images are already on the system.
Note images with the repository or tag in the following exclude list are always pulled:

```
╰─$ cat scripts/images/cache.yaml
exclude:
  repos:
    - docker.io/rancher/harvester
    - docker.io/rancher/harvester-webhook
    - docker.io/rancher/harvester-upgrade
  tags: []
```
Images with `*-head` tags are always pulled.

Usage:
```
USE_LOCAL_IMAGES=true make
```
If env `USE_LOCAL_IMAGES` is not provided, we'll try pull images as usual.

### DIFF
--- a/scripts/images/cache.yaml
+++ b/scripts/images/cache.yaml
@@ -1,0 +1,6 @@
+exclude:
+  repos:
+    - docker.io/rancher/harvester
+    - docker.io/rancher/harvester-webhook
+    - docker.io/rancher/harvester-upgrade
+  tags: []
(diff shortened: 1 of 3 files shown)


Vicente-Cheng (approved):
LGTM, thanks!

scripts/images/cache.yaml
```diff
@@ -0,0 +1,9 @@
+exclude:
+  repos:
+    - docker.io/rancher/harvester
+    - docker.io/rancher/harvester-webhook
+    - docker.io/rancher/harvester-upgrade
+  tags:
+    - master-head
```
futuretea, line 7: Very useful enhancement!
Can we support `*-head` ? will be useful for custom branch build

starbops (approved):
LGTM, thanks

scripts/images/cache.yaml
bk201: Fixed! Images with`*-head` tags are always pulled now.
Users still can add more tags to exclude in the `cache.yaml` file.

PR #455, opened 2023-03-13
Add /var/lib/third-party to PERSISTENT_STATE_PATHS

futuretea:
**Problem:**

The default rook ceph cluster data path on the host is '/var/lib/rook'. However, this path is non-persistent on Harvester OS, which will cause the osd of rook ceph to crash after the node restarts.
We need a directory to store data for a third-party component like rook

**Solution:**

Add /var/lib/third-party to PERSISTENT_STATE_PATHS

**Related Issue:**
https://github.com/harvester/harvester/issues/2405

**Test plan:**

Refer to https://github.com/harvester/harvester/pull/3566

### DIFF
--- a/pkg/config/templates/cos-rootfs.yaml
+++ b/pkg/config/templates/cos-rootfs.yaml
@@ -22,6 +22,7 @@ environment:
     /var/lib/kubelet
     /var/lib/wicked
     /var/lib/cni
+    /var/lib/third-party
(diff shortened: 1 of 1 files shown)


bk201 (approved):
lgtm

Vicente-Cheng (approved):
LGTM, thanks!

PR #456, opened 2023-03-14
Use the v1.1 harvester branch

bk201:
Make sure we use the correct backend code when building from the v1.1 branch.
Fix the ISO using the wrong harvester/harvester code if its build is triggered from the `harvester/harvester-installer` repo.


starbops (approved):
LGTM, thanks for fixing this!

Vicente-Cheng (approved):
LGTM, thanks for the fix!

guangbochen (approved):

PR #458, opened 2023-03-21
fix(install): bring up containerd only

starbops:
**Problem:**

Originally, the installer places the tarballs under a specific location and brings up RKE2 server then import the container images by itself. This seems convenient, but when bad things happen it could be hard to diagnose/address.

**Solution:**

Now, the installer will bring up containerd only and actively import the images if it's an ISO installation (iPXE installation will be the same as usual). This prevents the hardcoded 15 mins timeout of RKE2 from happening on a slow-speed media installation such as IPMI and USB. And it also provides better observability for users to know the progress of an installation.

**Related Issue:**

harvester/harvester#2651

**Test plan:**

ISO mode (the installation could take much longer than expected, but it should success eventually)

1. Install Harvester on a bare-metal server with the ISO image mounted via its BMC's virtual media function
2. Wait for the installation to finish (observing the console, there should be some image copying and loading messages)
3. The installation should be successful

### DIFF
--- a/package/harvester-os/files/usr/sbin/harv-install
+++ b/package/harvester-os/files/usr/sbin/harv-install
@@ -151,14 +151,16 @@ preload_rke2_images()
     RKE2_IMAGES_LISTS_DIR="/tmp/images-lists"
     mkdir -p $TARGET/$RKE2_IMAGES_DIR
     mkdir -p $TARGET/$RKE2_IMAGES_LISTS_DIR
-    mount --bind ${ISOMNT}/bundle/harvester/images $TARGET/$RKE2_IMAGES_DIR
     mount --bind ${ISOMNT}/bundle/harvester/images-lists $TARGET/$RKE2_IMAGES_LISTS_DIR
+
+    echo "Copying RKE2 images to the target location..."
+    rsync -ah --progress ${ISOMNT}/bundle/harvester/images/rke2-images.*.tar.zst $TARGET/$RKE2_IMAGES_DIR
 
     cd $TARGET
     mount --bind /dev dev
     mount --bind /proc proc
 
-    echo "Loading images. This may take a few minutes..."
+    echo "Bringing up RKE2 server to extract essential binaries..."
     chroot . /bin/bash <<"EOF"
       set -e
       # update the nameserver
@@ -182,13 +184,43 @@ preload_rke2_images()
 
       $rke2_tmp/bin/rke2 server &> /rke2.log &
 
+      until [ -f "/var/lib/rancher/rke2/bin/containerd" ] && [ -f "/var/lib/rancher/rke2/agent/etc/containerd/config.toml" ]
+      do
+        sleep 1
+      done
+
+      echo "Stop rke2 and remove tmporary rke2 files..."
+      pkill rke2
+      rm /rke2.log
+      rm /etc/rancher/rke2/rke2.yaml
+      rm -rf /var/lib/rancher/rke2/server
+      rm -rf /var/lib/rancher/rke2/agent/pod-manifests/*
+EOF
+
+    echo "Copying remaining images to the target location..."
+    rsync -ah --progress ${ISOMNT}/bundle/harvester/images/ $TARGET/$RKE2_IMAGES_DIR
+
+    echo "Loading images. This may take a few minutes..."
+    chroot . /bin/bash <<"EOF"
+
       export PATH=/var/lib/rancher/rke2/bin:$PATH
       export CONTAINERD_ADDRESS=/run/k3s/containerd/containerd.sock
+
+      echo "Start containerd..."
+      containerd -c /var/lib/rancher/rke2/agent/etc/containerd/config.toml -a $CONTAINERD_ADDRESS --state /run/k3s/containerd --root /var/lib/rancher/rke2/agent/containerd &> /containerd.log &
 
       # wait for containerd to be ready
       until ctr --connect-timeout 1s version&>/dev/null
       do
         sleep 1
+      done
+
+      # load images
+      for i in /var/lib/rancher/rke2/agent/images/*.tar.zst; do
+        echo "Load images from $i"
+        zstd -d $i -o /usr/local/images.tar
+        ctr -n k8s.io images import /usr/local/images.tar
+        rm /usr/local/images.tar
       done
 
       # make sure all preloading images are ready
@@ -197,12 +229,9 @@ preload_rke2_images()
       done
 
       # stop containerd
-      pkill rke2
-      rm /rke2.log
-      rm /etc/rancher/rke2/rke2.yaml
-
-      rm -rf /var/lib/rancher/rke2/server
-      rm -rf /var/lib/rancher/rke2/agent/pod-manifests/*
+      echo "Stop containerd..."
+      pkill containerd
+      rm /containerd.log
 EOF
 
     until umount dev&>/dev/null
@@ -211,8 +240,9 @@ EOF
     done
     umount proc
     cd - &> /dev/null
-    umount ${TARGET}/${RKE2_IMAGES_DIR}
     umount ${TARGET}/${RKE2_IMAGES_LISTS_DIR}
+
+    rm -f ${TARGET}/${RKE2_IMAGES_DIR}/*
 }
 
 preload_rancherd_images()
--- a/package/harvester-os/files/usr/sbin/harv-install
+++ b/package/harvester-os/files/usr/sbin/harv-install
@@ -147,29 +147,57 @@ preload_rke2_images()
         return
     fi
 
-    RKE2_IMAGES_DIR="/var/lib/rancher/rke2/agent/images"
-    RKE2_IMAGES_LISTS_DIR="/tmp/images-lists"
-    mkdir -p $TARGET/$RKE2_IMAGES_DIR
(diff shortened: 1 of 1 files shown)


starbops:
With this fix applied, I observed that in an extreme environment, e.g. mounting ISO image via BMC's virtual media function over VPN, even the extracted `containerd` binary is missing, which means the `containerd` did not start. After 15 minutes, the `rke2 server` timed out for waiting for the `containerd` to be ready, and the whole installation process hung.

```
      # extract rke2 binary
      rke2_tmp="$inst_tmp/rke2"
      mkdir -p $rke2_tmp && tar xf $inst_tmp/rke2.linux-amd64.tar.gz -C $rke2_tmp

      $rke2_tmp/bin/rke2 server &> /rke2.log &

      until [ -f "/var/lib/rancher/rke2/bin/containerd" ] && [ -f "/var/lib/rancher/rke2/agent/etc/containerd/config.toml" ]
      do
        sleep 1
      done
```

Below is the `rke2.log`, starting from 03:24:24

Screen Shot 2023-03-21 at 11 43 17

And it indicates the `rke2 server` was stopped at 03:39:24

Screen Shot 2023-03-21 at 11 43 22

haimari:
Hello, 
When this fix is expected to be merged and released?
Thank you

starbops:
From what we observed and compared with the normal case, `rke2 server` seems to be stuck at an early stage checking local images to extract those binaries.

In [rke2.log](https://github.com/harvester/harvester-installer/files/11037079/rke2.log):

```
...
time="2023-03-22T02:19:00Z" level=info msg="Checking local image archives in /var/lib/rancher/rke2/agent/images for index.docker.io/rancher/rke2-runtime:v1.24.11-rke2r1"
...
```

Since the images are located in the directory which is bind mounted from the Live CD (ISO image), this could be problematic in an extreme installation environment. We're planning to simply copy the images to the target location instead of bind-mount them. This might slow down the installation speed of the normal cases but will add more stability to the cases with low transfer speeds like USB 2.0 and BMC virtual median mount.

starbops:
> Hello, When this fix is expected to be merged and released? Thank you

Hi @haimari, the fix is targeted in v1.2.0. Thank you.

starbops:
It's generally hard to hold up the fragile connection (across the Pacific over VPN) for 3~4 hours to finish the installation with that slow speed:

Screen Shot 2023-03-22 at 15 18 31

starbops:
Confirmed that the "containerd binary not extracted" issue is remedied if we copy the image to the local disk at first.

Screen Shot 2023-03-23 at 11 30 57

Screen Shot 2023-03-23 at 11 31 50

package/harvester-os/files/usr/sbin/harv-install
```diff
@@ -151,14 +151,16 @@ preload_rke2_images()
     RKE2_IMAGES_LISTS_DIR="/tmp/images-lists"
     mkdir -p $TARGET/$RKE2_IMAGES_DIR
     mkdir -p $TARGET/$RKE2_IMAGES_LISTS_DIR
-    mount --bind ${ISOMNT}/bundle/harvester/images $TARGET/$RKE2_IMAGES_DIR
     mount --bind ${ISOMNT}/bundle/harvester/images-lists $TARGET/$RKE2_IMAGES_LISTS_DIR
 
+    echo "Copying RKE2 images to the target location..."
+    rsync -ah --progress ${ISOMNT}/bundle/harvester/images/rke2-images.*.tar.zst $TARGET/$RKE2_IMAGES_DIR
```
guangbochen, line 157: would it be better to specify `*.tar.zst` only?

package/harvester-os/files/usr/sbin/harv-install
```diff
@@ -182,27 +184,54 @@ preload_rke2_images()
 
       $rke2_tmp/bin/rke2 server &> /rke2.log &
 
+      until [ -f "/var/lib/rancher/rke2/bin/containerd" ] && [ -f "/var/lib/rancher/rke2/agent/etc/containerd/config.toml" ]
```
bk201, line 187: The PR generally looks good, and I've finished installation multiple times.
By checking RKE2/k3s code, here the containerd configuration [is generated from a template.](https://github.com/k3s-io/k3s/blob/7af5b16788afe9ce1718d7b75b35eafac7454705/pkg/agent/containerd/containerd.go#L40)
So RKE2 may be killed immediately after this check and leave a non-complete configuration file.
Can we use the `ctr` command to wait until the containerd brought by RKE2 is ready? Somehow similar to the [wait](https://github.com/k3s-io/k3s/blob/7af5b16788afe9ce1718d7b75b35eafac7454705/pkg/agent/containerd/containerd.go#L96) here.

package/harvester-os/files/usr/sbin/harv-install
```diff
@@ -182,27 +184,54 @@ preload_rke2_images()
...
+      containerd -c /var/lib/rancher/rke2/agent/etc/containerd/config.toml -a $CONTAINERD_ADDRESS --state /run/k3s/containerd --root /var/lib/rancher/rke2/agent/containerd &> /containerd.log &
+
       # wait for containerd to be ready
       until ctr --connect-timeout 1s version&>/dev/null
       do
         sleep 1
       done
 
+      # load images
+      for i in /var/lib/rancher/rke2/agent/images/*.tar.zst; do
```
masteryyh, line 219: Can you put this path prefix to a variable?

masteryyh (commented):
Finished some installation tests in bad network environment and shows good stability, just some small details can be enhanced, appreciated.

package/harvester-os/files/usr/sbin/harv-install
starbops: There are actually no differences between "copy rke2-images first" and "copy all images." `rsync` will only copy files that are missing at the destination. The reason we only copy rke2-images here is to make containerd available quicker and to make the whole process looks more responsive; therefore, we have a two-stage copy.

starbops (commented):
Current implementation will slow down the iPXE installation speed. For the scenarios of iPXE install, such mechanism is not needed.

Thanks @bk201 for the input!

package/harvester-os/files/usr/sbin/harv-install
starbops: Updated, thanks

package/harvester-os/files/usr/sbin/harv-install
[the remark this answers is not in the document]
starbops: The `rke2*` images are still copied to the same path because `rke2 server` needs them. Other images are copied to a different location to prevent `rke2 server` from preloading them automatically.

package/harvester-os/files/usr/sbin/harv-install
starbops: Updated, thanks.

package/harvester-os/files/usr/sbin/harv-install
```diff
@@ -182,27 +184,54 @@ preload_rke2_images()
...
+    rsync -ah --progress ${ISOMNT}/bundle/harvester/images/ $TARGET/$RKE2_IMAGES_DIR
+
+    echo "Loading images. This may take a few minutes..."
+    chroot . /bin/bash <<"EOF"
+
       export PATH=/var/lib/rancher/rke2/bin:$PATH
       export CONTAINERD_ADDRESS=/run/k3s/containerd/containerd.sock
 
+      echo "Start containerd..."
+      containerd -c /var/lib/rancher/rke2/agent/etc/containerd/config.toml -a $CONTAINERD_ADDRESS --state /run/k3s/containerd --root /var/lib/rancher/rke2/agent/containerd &> /containerd.log &
```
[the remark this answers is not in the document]
starbops: Updated, and those constants are now showing up just once. Thanks.

bk201 (approved):
LGTM

package/harvester-os/files/usr/sbin/harv-install
```diff
@@ -185,34 +215,64 @@ preload_rke2_images()
...
-
-      rm -rf /var/lib/rancher/rke2/server
-      rm -rf /var/lib/rancher/rke2/agent/pod-manifests/*
+      # tearing down containerd/rke2
+      if [ "$install_mode" = "ISO" ]; then
+        echo "Stop containerd..."
+        pkill containerd
+        rm /containerd.log
+      else
+        echo "Stop rke2..."
```
bk201, line 253:
```suggestion
        echo "Stop RKE2..."
```

Just a small nit to upper case the word in the script.

starbops: Updated, thanks!

guangbochen (approved):
This PR is LGTM, as a future enhancement, can we add a proper integration test to verify this custom bring-up and image load mechanism(by a new issue)? So it will be able the spot and fail early on CI build if RKE2 or upstream contained has made any breaking changes.

futuretea (approved):
LGTM, Thanks!

masteryyh (approved):
LGTM, thanks

starbops:
@mergify backport v1.1

w13915984028:
A question:

Do we have a test about the time difference between `rsync` and `mount` for fast environment like KVM VM ?

If the difference is significant, we may consider add `file copy` or `mount` option  into `harvester-configuration`, as a further enhancement.

Thanks.

PR #402, opened 2023-01-17
change logging monitoring from managedchart to addon

w13915984028:
Signed-off-by: Jian Wang <w13915984028@gmail.com>

    change logging monitoring from managedchart to addon
    
      change logging managedchart to addon
      change monitoring managedchart to addon
      replace VIP in installer
      patch rancher-logging to add eventtailer feature
      expose addon to harvester config

    update logging chart version
    
      from 100.1.3+up3.17.7 to 102.0.0+up3.17.10
      to better match the coming Rancher v2.7.3 in Harvester v1.2.0

    extend harvester-configuration for edge scenarios
    
      expose LH manager and replica CPU resource reservation to harvester-configuration    
      expose default storage class (harvester-longhorn) replica count to harvester-configuration

Issue: 
https://github.com/harvester/harvester/issues/2992
https://github.com/harvester/harvester/issues/3721

Upgrade PR: https://github.com/harvester/harvester/pull/3380

Doc PR:
https://github.com/harvester/docs/pull/282
https://github.com/harvester/docs/pull/301

**Test plan:**
The monitoring, logging are disabled by default in new installation. User can enable/disable them in Harvester UI.

Meanwhile, a harvester-configuration file (https://docs.harvesterhci.io/v1.1/install/harvester-configuration/) like following can set those options, it is useful for ISO&PXE installation.

harvester-configuration.yaml
```
scheme_version: 1
install:
  mode: create
  addons:
    rancher_logging:
      enabled: false
    rancher_monitoring:
      enabled: false
  harvester:
    storageClass:
      replicaCount: 1
    longhorn:
      defaultSettings:
        guaranteedEngineManagerCPU: 4
        guaranteedReplicaManagerCPU: 4
```

About those options:

```
addons.enabled 
if not setting, use default value false
true will enable the addon during installation
false will disable the addon during installation.


replicaCount:  
if not setting, use default value 3
valid values: 1~3, other values are reverted to 3

guaranteedEngineManagerCPU: 
if not setting, use default value 12
valid values 0~12, other values are reverted to 12

guaranteedReplicaManagerCPU: 
if not setting, use default value 12
valid values 0~12, other values are reverted to 12
```

Check the effectiveness of those settings:
(1) `kubectl get addons.harvesterhci.io -A`, check the enable/disable of related addons

(2) `kubectl get settings.longhorn.io -A`, check the values of following:
```
longhorn-system   guaranteed-engine-manager-cpu                        12                                                 10m
[...]

PR #459, opened 2023-03-22
Support os.persistentStatePaths in Harvester configuration

futuretea:
**Problem:**

Currently, Harvester OS does not support custom paths for persistent storage or read-write paths. This means that any changes made to files not in default persistentStatePaths will be lost upon reboot.

**Solution:**

To address this issue, this pull request adds a  new options to the Harvester OS configuration in Harvester config. 

 The `os.persistentStatePaths` option allows users to configure custom paths where modifications made to files will persist across reboots. This means that any changes made to files in these paths will not be lost upon reboot.

an config example for install rook-ceph in Harvester
```yaml
os:
  persistentStatePaths:
    - /var/lib/rook
  modules:
    - rbd
    - nbd
``` 

Without this PR, users can achieve the same effect by modifying `/oem/90_custom.yaml` and then rebootting. This PR allows users to define the desired os layout at installation time.

**Related Issue:**
https://github.com/harvester/harvester/issues/3688

**Note**
Please ensure that any custom paths added using these options are valid and secure, and that any modifications made to the OS are done with caution and understanding of the potential risks.

**Test plan:**

Refer to https://github.com/harvester/harvester/pull/3566

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -148,6 +148,8 @@ type OS struct {
 	Password       string            `json:"password,omitempty"`
 	Environment    map[string]string `json:"environment,omitempty"`
 	Labels         map[string]string `json:"labels,omitempty"`
+
+	PersistentStatePaths []string `json:"persistentStatePaths,omitempty"`
 }
 
 type HarvesterConfig struct {
--- a/pkg/config/templates/cos-rootfs.yaml
+++ b/pkg/config/templates/cos-rootfs.yaml
@@ -25,6 +25,11 @@ environment:
     /var/lib/third-party
     /var/crash
     /var/lib/longhorn
+    {{- if .OS.PersistentStatePaths }}
+    {{- range $path := .OS.PersistentStatePaths }}
+    {{ $path }}
+    {{- end }}
+    {{- end }}
     {{- if not .ShouldMountDataPartition }}
     /var/lib/harvester/defaultdisk
     {{- end }}
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -276,6 +276,23 @@ func TestHarvesterRootfsRendering(t *testing.T) {
 				assert.NotContains(t, rootfs.Environment["PERSISTENT_STATE_PATHS"], "/var/lib/harvester/defaultdisk")
 			},
 		},
+		{
+			name: "Test additional persistent state paths",
+			harvConfig: HarvesterConfig{
+				OS: OS{
+					PersistentStatePaths: []string{
+						"/path1",
+						"/path2",
+					},
+				},
+			},
+			assertion: func(t *testing.T, rootfs *Rootfs) {
+				assert.Contains(t, rootfs.Environment["VOLUMES"], "LABEL=HARV_LH_DEFAULT:/var/lib/harvester/defaultdisk")
+				assert.NotContains(t, rootfs.Environment["PERSISTENT_STATE_PATHS"], "/var/lib/harvester/defaultdisk")
+				assert.Contains(t, rootfs.Environment["PERSISTENT_STATE_PATHS"], "/path1")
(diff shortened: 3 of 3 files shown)


bk201 (approved):

futuretea:
@bk201 I've added a commit, added an os.rwPaths option to allow changes to the RW_PATHS configuration, and the PR documentation has been updated

pkg/config/templates/cos-rootfs.yaml
```diff
@@ -5,7 +5,7 @@ environment_file: /run/cos/cos-layout.env
 environment:
   VOLUMES: "LABEL=COS_OEM:/oem LABEL=COS_PERSISTENT:/usr/local{{ if .ShouldMountDataPartition }} LABEL=HARV_LH_DEFAULT:/var/lib/harvester/defaultdisk{{ end }}"
   OVERLAY: "tmpfs:25%"
-  RW_PATHS: "/var /etc /srv /boot"
+  RW_PATHS: "/var /etc /srv /boot{{ if .OS.RWPaths }}{{- range $path := .OS.RWPaths }} {{ $path }}{{- end }}{{ end }}"
```
guangbochen, line 8: Would this also take effect with day2 configurations?

futuretea: can change it in /oem/90_custom.yaml

futuretea: Without this PR, users can achieve the same effect by modifying `/oem/90_custom.yaml` and then rebootting. This PR allows users to define the desired os layout at installation time.

futuretea:
convert this PR to draft, try to add [after-install-chroot](https://rancher.github.io/elemental-toolkit/docs/customizing/stages/#after-install-chroot) in https://github.com/harvester/harvester-installer/blob/12eeb114020e988b59bb46cba78b1bf0d3c0b48f/pkg/config/cos.go#L183

pkg/config/templates/cos-rootfs.yaml
futuretea: removed the rwpath option

bk201 (approved):
LGTM

Vicente-Cheng (approved):
LGTM!

guangbochen (approved):
LGTM, thanks.

pkg/config/config.go
```diff
@@ -148,6 +148,8 @@ type OS struct {
 	Password       string            `json:"password,omitempty"`
 	Environment    map[string]string `json:"environment,omitempty"`
 	Labels         map[string]string `json:"labels,omitempty"`
+
+	PersistentStatePaths []string `json:"persistentStatePaths,omitempty"`
```
w13915984028, line 152: do we consider to add some checks/limitations for those paths

e.g. /tmp should be excluded

and, if a path is invalid/not existing, what will happen ?

In this commit https://github.com/harvester/harvester-installer/pull/402/files#diff-c49da5a63fc83c7c4dda6020b19d2e217bf5a8bb1911c52b5c60d84793c51149,  some checks are done.

futuretea: **/tmp**: I'm not sure if there is such a user scenario, but I tested it and elemental supports persistence of this path
**not existing path**: this is where users typically use this feature, such as `/var/lib/rook`
**invalid path**: I think we do need to add a validator to verify that the user input is a linux path

futuretea: I did a quick look around and didn't find any existing functions to verify linux paths properly, so I decided to create a separate PR

guangbochen: @futuretea @w13915984028 Let's add the validation enhancement separately, I will merge this one first. cc @masteryyh PTAL and feedback are still welcome.

futuretea: added and merged https://github.com/harvester/harvester-installer/pull/487

PR #467, opened 2023-04-11
Support os.afterInstallChrootCommands in Harvester configuration

futuretea:
**Problem:**

Harvester OS does not support users adding additional software packages.

**Solution:**

To support this requirement, elemental-toolkit provides a stage called [after-install-chroot](https://rancher.github.io/elemental-toolkit/docs/customizing/stages/#after-install-chroot). Commands executed in this stage do not encounter file system write issues and the results of user commands are persistent and will not be lost due to a reboot. We can use this mechanism and expose it through Harvester Configuration.

**Related Issue:**
https://github.com/harvester/harvester/issues/3689

**Test plan:**

1.  build iso from this PR
2.  install node use the following Harvester configuration example:
```yaml
os:
  afterInstallChrootCommands:
    - mkdir /bin/justatestdir
    - cd /bin/justatestdir
    - touch /bin/justatestdir/a
```
3. ssh into node, check status
```bash
ls /bin/justatestdir/a
```
4. reboot
5. repeat step 3

### DIFF
--- a/pkg/config/templates/cos-after-install-chroot.yaml
+++ b/pkg/config/templates/cos-after-install-chroot.yaml
@@ -1,0 +1,9 @@
+# yip stage to overwrite after-install-chroot layout
+if: '[ ! -f "/run/cos/recovery_mode" ]'
+name: "Run after-install-chroot commands"
+commands:
+    {{- if .OS.AfterInstallChrootCommands }}
+    {{- range $cmd := .OS.AfterInstallChrootCommands }}
+    - {{ printf "%q" $cmd }}
+    {{- end }}
+    {{- end }}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -136,9 +136,10 @@ type File struct {
 }
 
 type OS struct {
-	SSHAuthorizedKeys []string `json:"sshAuthorizedKeys,omitempty"`
-	WriteFiles        []File   `json:"writeFiles,omitempty"`
-	Hostname          string   `json:"hostname,omitempty"`
+	AfterInstallChrootCommands []string `json:"afterInstallChrootCommands,omitempty"`
+	SSHAuthorizedKeys          []string `json:"sshAuthorizedKeys,omitempty"`
+	WriteFiles                 []File   `json:"writeFiles,omitempty"`
+	Hostname                   string   `json:"hostname,omitempty"`
 
 	Modules        []string          `json:"modules,omitempty"`
 	Sysctls        map[string]string `json:"sysctls,omitempty"`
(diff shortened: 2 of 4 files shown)


masteryyh (approved):
LGTM, tested with fresh installation
Before reboot
```
rancher@harv-test:~> ls -l /bin/justatestdir/a
-rw-r--r-- 1 root root 0 Apr 13 03:35 /bin/justatestdir/a
```
After reboot
```
rancher@harv-test:~> ls -l /bin/justatestdir/a
-rw-r--r-- 1 root root 0 Apr 13 03:35 /bin/justatestdir/a
```

bk201 (approved):
work well, thanks!

guangbochen (approved):
LGTM, thanks.

pkg/config/templates/cos-after-install-chroot.yaml
```diff
@@ -0,0 +1,9 @@
+# yip stage to overwrite after-install-chroot layout
+if: '[ ! -f "/run/cos/recovery_mode" ]'
+name: "Run after-install-chroot commands"
+commands:
+    {{- if .OS.AfterInstallChrootCommands }}
+    {{- range $cmd := .OS.AfterInstallChrootCommands }}
+    - {{ $cmd }}
```
Vicente-Cheng, line 7: Do we need to use `printf "%q"` here for better handling with the string `"`?
But I see the test code. It looks like we can already handle `"` well here.

Vicente-Cheng (approved):
Using the following test plan
```
afterInstallChrootCommands:
    - mkdir /bin/justatestdir
    - cd /bin/justatestdir
    - touch /bin/justatestdir/a
    - echo "hello world!" > /bin/justatestdir/hello.txt
    - echo "'aaabbcc'" >  /bin/justatestdir/hello2.txt
```

The result is following:
```
rancher@harvester-node-0:~> ls /bin/justatestdir/
a  hello.txt  hello2.txt
rancher@harvester-node-0:~> cat /bin/justatestdir/hello.txt
hello world!
rancher@harvester-node-0:~> cat /bin/justatestdir/hello2.txt
'aaabbcc'
```

LGTM, thanks!

w13915984028:
A general question is about security: when we allow user to run commands, how to make sure those commands are not breaking/touching harvester related path&files.

And, if the related command binaries are not in the system, what will happen?/ how should user do ? should we allow user to import and install binaries/packages ?

bk201:
@w13915984028 The target user of the feature is the people deploying the node. Usually, it's the admin. You can't forbid the user to do anything they want since the feature allows running arbitrary commands. The user use the feature should take care not to mess up the system.
If the binary/artifacts are not presented on the fs, the user must fetch them from an external server or add Zypper repo to install them.

w13915984028:
> @w13915984028 The target user of the feature is the people deploying the node. Usually, it's the admin. You can't forbid the user to do anything they want since the feature allows running arbitrary commands. The user use the feature should take care not to mess up the system. If the binary/artifacts are not presented on the fs, the user must fetch them from an external server or add Zypper repo to install them.

When we give all rights to users, we need an very detailed document to describe all those situations.  e.g download binary and install and execute them.

And it will be hard for us to define and test this feature, as the scope is un-defined/flexible.

w13915984028:
BTW: I am not doubting this PR, but the big-story itself.

PR #468, opened 2023-04-12
feat(rancherd): enable mcm (Warning: only works for new cluster)

FrankYang0529:
**Solution:**
Enable `multi-cluster-management` feature.

**Related Issue:**
https://github.com/harvester/harvester/issues/2679

**Test plan:**
1. Create a new cluster.
2. Check `mult-cluster-management` feature. It should be `true`.

### DIFF
--- a/package/harvester-os/files/usr/share/rancher/rancherd/config.yaml.d/50-defaults.yaml
+++ b/package/harvester-os/files/usr/share/rancher/rancherd/config.yaml.d/50-defaults.yaml
@@ -3,7 +3,10 @@ rancherValues:
   rancherImage: rancher/rancher
   rancherImageTag: v2.6.11
   noDefaultAdmin: false
-  features: multi-cluster-management=false,multi-cluster-management-agent=false
+  # multi-cluster-management-agent is to run downstream controllers.
+  # By default, a harvester cluster is not a downstream cluster, so we don't enable it.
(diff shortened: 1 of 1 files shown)


package/harvester-os/files/usr/share/rancher/rancherd/config.yaml.d/50-defaults.yaml
```diff
@@ -3,7 +3,7 @@ rancherValues:
   rancherImage: rancher/rancher
   rancherImageTag: v2.6.11
   noDefaultAdmin: false
-  features: multi-cluster-management=false,multi-cluster-management-agent=false
+  features: multi-cluster-management=true,multi-cluster-management-agent=false
```
guangbochen, line 6: can u please help to elaborate on what the `multi-cluster-management-agent` is used for and why we don't need it?

bk201 (approved):

package/harvester-os/files/usr/share/rancher/rancherd/config.yaml.d/50-defaults.yaml
FrankYang0529: Added it. Thanks.

guangbochen (approved):
LGTM, thanks.

PR #474, opened 2023-04-15
Avoid iptables rules affecting bridge forwarding

yaocw2020:
**Problem:**

In the Harvester cluster whose management network has a VLAN ID, the VM will be unable to access the host port or node port with the host IP where the VM is running. 

**Solution:**

Disable `net.bridge.bridge-nf-call-iptables` to avoid iptables rules affecting bridge forwarding.

**Related Issue:** https://github.com/harvester/harvester/issues/3414

**Test plan:**

- Spin up a Harvester whose management network has a VLAN ID
- Create a VM with VLAN network whose VLAN ID is same with the management network.
- Curl <nodeIP:443> in the VM.

### DIFF
--- a/package/harvester-os/files/etc/sysctl.d/net.conf
+++ b/package/harvester-os/files/etc/sysctl.d/net.conf
@@ -2,3 +2,6 @@
 net.ipv6.conf.all.disable_ipv6 = 1
 net.ipv6.conf.default.disable_ipv6 = 1
 net.ipv6.conf.lo.disable_ipv6 = 1
+
+# avoid iptables rules affecting bridge forwarding
+net.bridge.bridge-nf-call-iptables = 0


bk201:
@yaocw2020 I feel the PR is not for issue 3744. Could we create a new issue to track it? Thanks.

yaocw2020:
> @yaocw2020 I feel the PR is not for issue 3744. Could we create a new issue to track it? Thanks.

Sorry, it's issue 3414. I correct it.

guangbochen:
the test plan for this issue is not solid, we need to at least validate both the host network and guest k8s networks since this is a system-level configuration and its impact could be large.

yaocw2020:
Close because it could not solve the problem and I submit a new PR https://github.com/harvester/network-controller-harvester/pull/83

PR #485, opened 2023-04-21
Set max-pods default to 200 [Fixes harvester/harvester#2707]

tlehman:
Set the max-pods to 200 as per [issue 2707](https://github.com/harvester/harvester/issues/2707). 

For the upgrade path: https://github.com/harvester/harvester/pull/3815


bk201 (approved):
LGTM, the arguments take effect for both controller and agent nodes.

ibrokethecloud (approved):
LGTM. Tested and can see the new pod limit applied.

PR #497, opened 2023-05-31
config: update NTP to SystemSettings

Vicente-Cheng:
**Problem:**
If we can config the NTP server on UI, we need to make it consistent after the first node creation.

**Solution:**
write the related setting with the mechanism of updating the system settings

**Related Issue:**
https://github.com/harvester/harvester/issues/3133

**Test plan:**
Make sure we have `20-harvester-settings.yaml` and it should contain the ntp-servers information


bk201 (approved):
lgtm

guangbochen (approved):
LGTM, thanks.

starbops (approved):
LGTM, thanks!

PR #535, opened 2023-07-10
golangci: add gosec to linter

Vicente-Cheng:
**Problem:**
We prefer to add `gosec` to golangci linter

**Solution:**
Add `gosec` to linter and fix related issues.

**Related Issue:**
https://github.com/harvester/harvester/issues/4019

**Test plan:**
Make sure installation and upgrade should work as usual.


pkg/console/vip.go
```diff
@@ -25,8 +25,12 @@ func createMacvlan(name string) (netlink.Link, error) {
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to fetch %s", name)
 	}
-	rand.Seed(time.Now().UnixNano())
-	macvlanName := tempMacvlanPrefix + strconv.Itoa(rand.Intn(100))
+	randNum, err := rand.Int(rand.Reader, big.NewInt(100))
```
bk201, line 28: @yaocw2020 Can you help review this part? Thanks!

bk201 (approved):

pkg/console/vip.go
yaocw2020: What are the benefits to use `crypto/rand` instead of `math/rand`?

Vicente-Cheng: Hi @yaocw2020, thanks for the review.
Using `crypto/rand` instead `math/rand` due to the `G404: Use of weak random number generator (math/rand instead of crypto/rand)` (gosec) checking.

pkg/console/vip.go
```diff
@@ -25,8 +25,12 @@ func createMacvlan(name string) (netlink.Link, error) {
...
+	if err != nil {
+		logrus.Errorf("Failed to generate random number: %v", err)
```
yaocw2020, line 30: I suggest returning error directly.

.golangci.yaml
```diff
@@ -0,0 +1,31 @@
...
+    # https://github.com/securego/gosec#available-rules
+    excludes:
+      - G101 # Look for hard coded credentials
+      - G108 # Profiling endpoint is automatically exposed on /debug/pprof
+      - G204 # Audit the use of command execution
+      - G401 # Use of weak cryptographic primitive
+      - G402 # TLS InsecureSkipVerify set true
+      - G505 # Blocklisted import crypto/md5: weak cryptographic primitive
+    config:
+      G306: "0644" # Poor file permissions used when writing to a new file
```
guangbochen, line 31: Remove those additional configs if they are not required; thanks.

pkg/console/vip.go
Vicente-Cheng: nice suggestion! Updated!

.golangci.yaml
Vicente-Cheng: Thanks for the reminder. Updated!

yaocw2020 (approved):
LGTM, thanks

bk201 (approved):

guangbochen (approved):
LGTM, thanks.

guangbochen:
@mergify backport v1.2

PR #562, opened 2023-09-04
config: make sftp could be configured dynamically

Vicente-Cheng:
**Problem:**
Give the way to config `sftp` dynamically

**Solution:**
We could set up the `harvester.os.sshd.sftp` to true.
The harvester-installer would create the related config file by stage `AfterInstallChrootCommands`

We also need the https://github.com/harvester/os2/pull/78
That would include the customer config that we would like to add.

**Related Issue:**
https://github.com/harvester/harvester/issues/4480

**Test plan:**

- Perform an ipxe-example installation with this config
    ```
    os:
      sshd:
         sftp: true
    ```
- Try to access sftp
    ```
    sftp rancher@192.168.0.30
    (rancher@192.168.0.30) Password:
    Connected to 192.168.0.30.
    sftp> 
    ```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -186,8 +186,16 @@ type OS struct {
 	Password       string            `json:"password,omitempty"`
 	Environment    map[string]string `json:"environment,omitempty"`
 	Labels         map[string]string `json:"labels,omitempty"`
+	SSHD           SSHDConfig        `json:"sshd,omitempty"`
 
 	PersistentStatePaths []string `json:"persistentStatePaths,omitempty"`
+}
+
+// SSHDConfig is the SSHD configuration for the node
+//
+//   - SFTP: the switch to enable/disable SFTP
+type SSHDConfig struct {
+	SFTP bool `json:"sftp,omitempty"`
 }
 
 type HarvesterConfig struct {
--- a/pkg/config/cos.go
+++ b/pkg/config/cos.go
@@ -205,6 +205,9 @@ func ConvertToCOS(config *HarvesterConfig) (*yipSchema.YipConfig, error) {
 		},
 	}
 
+	// Handle the sshd components
+	overwriteSSHDComponent(config)
+
 	// Add after-install-chroot stage
 	if len(config.OS.AfterInstallChrootCommands) > 0 {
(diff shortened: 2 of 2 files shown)


Vicente-Cheng:
Hi @guangbochen, @bk201,

This is an easy way to make config `sftp` simple. 
For this implementation, we only support automatic installation (which means we need to load the harvester config).
We could also implement the related part if we would like to expose this to the interactive installation mode.

What do you think of this implementation?

bk201:
@ibrokethecloud Please help check the issue and give some feedback. Thanks!

ibrokethecloud:
the config for `AfterInstallChrootCommands` is going to be lost post an upgrade. 

would it be feasible to make this an addon? and run a daemonset serving sftp?  This way we can rollout updates and ensure service is maintained post upgrade?

Vicente-Cheng:
> the config for `AfterInstallChrootCommands` is going to be lost post an upgrade.
> 
> would it be feasible to make this an addon? and run a daemonset serving sftp? This way we can rollout updates and ensure service is maintained post upgrade?

After I checked, the `/etc/ssh` is the persistent path. Anything we added on it will be saved even we do a system upgrade.
So I think we could simply apply this file.

pkg/config/config.go
```diff
@@ -147,6 +147,7 @@ type Install struct {
 	TTY       string `json:"tty,omitempty"`
 	ForceGPT  bool   `json:"forceGpt,omitempty"`
 	Role      string `json:"role,omitempty"`
+	EnableSFTP bool   `json:"enableSftp,omitempty"`
```
bk201, line 150: I suggest we design the flag like this:
```yaml
os:
  sshd:
     sftp: true
```

This provides the opportunity to add more sshd configuration in the future. The reason to put it under `os` is the configuration is supposed to take effect after the installation (in real OS).

Vicente-Cheng: Sure, I move it into the sshd components.
Updated. Please help to recheck!

ibrokethecloud (approved):
lgtm. thanks.

Vicente-Cheng:
@mergifyio backport v1.2

PR #566, opened 2023-09-11
Fix partition table type

jaevans:
**Problem:**
Invalid setup when setting partition table type to MBR in the installer. The string in the elemental `config.yaml` should be `msdos`, not `mbr`

**Solution:**
Change the string in `PartTable` to the correct value

**Related Issue:**

harvester/harvester#4515

**Test plan:**


masteryyh (approved):
LGTM, tested installation with legacy mode without issue, thanks.

bk201 (approved):
lgtm!

Vicente-Cheng (approved):
LGTM, tested with the installation. Thanks for fixing it.

bk201:
@mergifyio backport v1.2

PR #567, opened 2023-09-12
build: keep the boot partition with label COS_GRUB

Vicente-Cheng:
**Problem:**
The boot partition changed after we repacked to support the legacy BIOS.
That would make the installation hang because of the wrong partition. We depend on the COS_LIVE to mount the data partition on installation.

**Solution:**
Keep the boot partition has the `COS_GRUB` label

**Related Issue:**
https://github.com/harvester/harvester/issues/4510

**Test plan:**
Make sure the whole installation works as usual.

### DIFF
--- a/scripts/package-harvester-os
+++ b/scripts/package-harvester-os
@@ -96,7 +96,7 @@ cp "${extract_dir}/rootfs.squashfs" "${ARTIFACTS_DIR}/${PROJECT_PREFIX}-rootfs-$
 # and copy EFI files to it to make it UEFI bootable
 uefi_img="${extract_dir}/boot/uefi.img"
 dd if=/dev/zero of="${uefi_img}" bs=1k count=4096 status=progress
-mkfs.vfat "${uefi_img}"
+mkfs.vfat "${uefi_img}" -n COS_GRUB
 mcopy -s -i "${uefi_img}" "${extract_dir}/EFI" ::
 
 # Remove original ISO, and repack it using xorriso


masteryyh (approved):
LGTM, tested installation without any issue.
I'm wondering how to _fully_ test this issue, as I can't even reproduce the problem this issue brought up 🤔

Vicente-Cheng:
> LGTM, tested installation without any issue. I'm wondering how to _fully_ test this issue, as I can't even reproduce the problem this issue brought up 🤔

Thanks for the review. For me, I reproduce in the following ways
1. Simulate the USB drive with ISO on the KVM (@bk201 has some trick for it)
2. Use dd or similar behavior ISO editor to flash ISO to USB to reproduce on my i7 laptop

BTW, `rufus` would not work because it will break the original ISO layout.

Vicente-Cheng:
Investigate the lab Jenkins failure, tested with interactive/iPXE mode on my local environment.

Vicente-Cheng:
> Investigate the lab Jenkins failure, it works fine with interactive/iPXE mode on my local environment.

Somehow installation was stuck before `installation`, try to install it on vnc and looks fine.
Re-trigger test again and looks normal.

bk201 (approved):
lgtm!

guangbochen (approved):
LGTM, thanks.

futuretea (approved):
LGTM, thanks. validated by balenaEtcher + KVM

bk201:
@mergifyio backport v1.2

PR #570, opened 2023-09-19
common: always set log level to debug when we install

Vicente-Cheng:
**Problem:**
Try to get more logs when we use the elemental command.

**Solution:**
Add debug parameter

**Related Issue:**
https://github.com/harvester/harvester/issues/4541

**Test plan:**
make sure the installation as usual

### DIFF
--- a/package/harvester-os/files/usr/sbin/harv-install
+++ b/package/harvester-os/files/usr/sbin/harv-install
@@ -5,10 +5,6 @@ ISOMNT=/run/initramfs/live
 TARGET=/run/cos/target
 DATA_DISK_FSLABEL="HARV_LH_DEFAULT"
 
-# Update environment variables
-export QUIET=${HARVESTER_SILENT:+"--quiet"}
-export DEBUG=${HARVESTER_DEBUG:+"--debug"}
-
 clear_disk_label()
 {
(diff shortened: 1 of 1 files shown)


masteryyh (approved):
LGTM, tested installation successful with a lot of logs.
BTW, why don't we add another GRUB menu entry for debug mode? It's somewhat confusing when seeing a lot of message outputs compares to earlier versions.

guangbochen (approved):
LGTM, thanks.

Vicente-Cheng:
> LGTM, tested installation successful with a lot of logs. BTW, why don't we add another GRUB menu entry for debug mode? It's somewhat confusing when seeing a lot of message outputs compares to earlier versions.

Thanks @masteryyh for this question!
Did you mean the `rd.debug` for the grub parameter?
That will not only affect the elemental output but will append more like dracut ... etc.

For the installation, we might not set the debug flag as default because we could try again and add `rd.debug` or `--debug` for it.
But we need to add a debug flag as default for the upgrade because we could not try again easily.

If you all thought we could add a debug flag once an error happened on installation, I will close this PR.
Feel free to give some thoughts on it! Thanks!! cc @bk201, @guangbochen, @masteryyh

bk201 (approved):

bk201:
@mergifyio backport v1.2

PR #596, opened 2023-11-21
console: validate disk size earlier, update error wording

tserong:
**Problem:**
The error messages are sometimes unclear when one has an installation disk that's too small.

**Solution:**
This commit adds disk size validation in the calls to `diskConfirm()` and `dataDiskConfirm()`, so that as you walk through the fields you'll see potential error messages as you go, rather than having to wait until leaving the last field on the page.  The error message clears if you press the up arrow to go to previous fields.
    
The persistent partition size check remains where it is, i.e. it's only done when you try to move to the next page.  I couldn't figure out a neat way to incorporate this into the `validateAllDiskSizes()` function, which gets called before the default persistent size is set in `persistentSizeV.PreShow()`.
    
This commit also changes the wording of the "disk size is too small" error messages to explicitly state *which* disk is too small, and makes a small tweak to the "partition size" error message.

**Related Issue:**

https://github.com/harvester/harvester/issues/4734

**Test plan:**

- Boot a system with one disk < 250Gi. Verify that when you hit ENTER or DOWN on the Installation disk field that the "Installation disk size is too small. Minimum 250Gi is required" message appears.
- Boot a system with two disks, one 120Gi, one 200Gi and verify:
  - With the 120Gi disk selected for both installation and data, you see "Installation disk size is too small. Minimum 250Gi is required"
  - With the 200Gi disk selected for both installation and data, verify the same error as above appears immediately.
  - Set the 120Gi disk for data and verify the error goes away.
- In all cases, verify that an appropriate error appears immediately after you leave each field going down (down arrow or enter key) and that the error disappears when pressing the up arrow to return to the previous field.

### DIFF
--- a/pkg/util/disk.go
+++ b/pkg/util/disk.go
@@ -17,12 +17,12 @@ const (
 
 func ParsePartitionSize(diskSizeBytes uint64, partitionSize string) (uint64, error) {
 	if diskSizeBytes < MinDiskSize {
-		return 0, fmt.Errorf("Disk size is too small. Minimum %dGi is required", ByteToGi(MinDiskSize))
+		return 0, fmt.Errorf("Installation disk size is too small. Minimum %dGi is required", ByteToGi(MinDiskSize))
 	}
 	actualDiskSizeBytes := diskSizeBytes - fixedOccupiedSize
 
 	if !sizeRegexp.MatchString(partitionSize) {
-		return 0, fmt.Errorf("Partition size should be ended with 'Mi', 'Gi', and no dot and negative is allowed")
+		return 0, fmt.Errorf("Partition size must end with 'Mi' or 'Gi'. Decimals and negatives are not allowed.")
 	}
 
 	size, err := strconv.ParseUint(partitionSize[:len(partitionSize)-2], 10, 64)
--- a/pkg/config/cos_test.go
+++ b/pkg/config/cos_test.go
@@ -35,22 +35,22 @@ func TestCalcCosPersistentPartSize(t *testing.T) {
 		{
 			diskSize:      150,
 			partitionSize: "100Gi",
-			err:           "Disk size is too small. Minimum 250Gi is required",
+			err:           "Installation disk size is too small. Minimum 250Gi is required",
 		},
 		{
 			diskSize:      300,
 			partitionSize: "153600Ki",
-			err:           "Partition size should be ended with 'Mi', 'Gi', and no dot and negative is allowed",
+			err:           "Partition size must end with 'Mi' or 'Gi'. Decimals and negatives are not allowed.",
 		},
 		{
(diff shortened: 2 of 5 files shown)


tserong:
Note to reviewers: I've put "Fixes: https://github.com/harvester/harvester/issues/4734" in the commit message, which will presumably automatically close that issue if this PR is merged. I'm not sure if that's usually done with harvester PRs. If _not_, please let me know and I'll remove that from the commit message.

bk201:
@tserong Thanks for the great PR! No, we don't like GitHub automatically closing an issue unless QA verifies it. So prefer not to use the fix/close keywords in the commits or PR.

tserong:
Thanks @bk201, I'll fix the commit message.

I've also just done some more testing and have discovered that by setting `c.config.Install.PersistentPartitionSize` in `persistentSizeV.PreShow()`, the default partition size ends up always being 150Gi, even if you change the install disk to something bigger, i.e. the calculation based on `DefaultPersistentPercentageNum` doesn't happen after changing to a different, larger installation disk, so I need to re-work that slightly.

tserong:
OK, that's better now.

Vicente-Cheng:
Hi @tserong, please force push again to trigger CI.
We fixed the issue https://github.com/harvester/harvester/issues/4754, which caused the CI to fail.

tserong:
Thanks @Vicente-Cheng that seems to have worked :-)

pkg/util/disk.go
```diff
@@ -17,12 +17,12 @@ const (
 
 func ParsePartitionSize(diskSizeBytes uint64, partitionSize string) (uint64, error) {
 	if diskSizeBytes < MinDiskSize {
-		return 0, fmt.Errorf("Disk size is too small. Minimum %dGi is required", ByteToGi(MinDiskSize))
+		return 0, fmt.Errorf("Installation disk size is too small. Minimum %dGi is required", ByteToGi(MinDiskSize))
```
bk201, line 20: nit: this makes the function only useful for the installation disk. Although we don't use the function in the data disk path, can we make the function more general like supporting passing a disk type? 

The other parts of the PR work great!

tserong: > Although we don't use the function in the data disk path, can we make the function more general like supporting passing a disk type?

I'm not sure. I mean, yeah, we could add a parameter to specify disk type :-) but looking at the rest of the function, it's specifically testing sizes against MinDiskSize and MinPersistentSize, so ISTM the way it's implemented now is really geared towards looking at the installation disk only. Maybe this is something we can look at refactoring separately?

bk201: Yeah, no need to be in this PR!

bk201 (approved):

bk201:
@mergifyio backport v1.2

PR #604, opened 2023-11-28
installer: add role to specify the capability of node

Vicente-Cheng:
Now support `default` or `etcd` role.
  - default means the master node (control-plane, etcd, master)
  - etcd means the etcd node (etcd)

  NOTE: this only supports with join mode because the first node needs to
  be the master node.

**Problem:**

Support specify the node role

**Solution:**
Add `install.role` to choose the wanted role.

**Related Issue:**
https://github.com/harvester/harvester/issues/3266

**Test plan:**


ibrokethecloud (approved):
LGTM. thanks.

pkg/console/util.go
```diff
@@ -456,6 +456,14 @@ func doInstall(g *gocui.Gui, hvstConfig *config.HarvesterConfig, webhooks Render
 		return err
 	}
 
+	// specific the node label for the specific node role
+	if hvstConfig.Role == config.RoleEtcd {
+		if hvstConfig.Labels == nil {
+			hvstConfig.Labels = make(map[string]string)
+		}
+		hvstConfig.Labels["node-role.harvesterhci.io/etcd"] = "true"
```
bk201, line 464: `node-role.harvesterhci.io/etcd` is used around the code, it would be nice to make it a constant.

bk201 (approved):
LGTM!

PR #608, opened 2023-12-05
install: manually create custom grub file

Vicente-Cheng:
**Problem:**
The boot stuck on the grub might be because of the search for a non-existent file.
(With specific Hardware, like BOSS-S1 raid card)

**Solution:**
Manually create these files with the installation

**Related Issue:**
https://github.com/harvester/harvester/issues/4784

**Test plan:**
Make sure the boot is not stuck on the bare metal with the BOSS-S1 raid card.

### DIFF
--- a/package/harvester-os/files/usr/sbin/harv-install
+++ b/package/harvester-os/files/usr/sbin/harv-install
@@ -366,11 +366,20 @@ update_grub_settings()
     # stuck on the grub file search for about 30 minutes, this can be
     # mitigated by adding the `grubenv` file.
     #
-    # PATCH: add /oem/grubenv if it does not exist
+    # We need to patch grubenv, grubcustom
     oem_dir=${TARGET}/oem
-    GRUBENV_FILE="${oem_dir}/grubenv"
-    if ! [ -f ${GRUBENV_FILE} ]; then
-        grub2-editenv ${GRUBENV_FILE} create
+
+    # PATCH1: add /oem/grubenv if it does not exist
+    # grubenv use load_env to load, so we use grub2-editenv
+    TARGET_FILE="${oem_dir}/grubenv"
+    if ! [ -f ${TARGET_FILE} ]; then
+        grub2-editenv ${TARGET_FILE} create
+    fi
+    # PATCH2: add /oem/grubcustom if it does not exist
+    # grubcustom use source to load, so we can use touch directly
+    TARGET_FILE="${oem_dir}/grubcustom"
+    if ! [ -f ${TARGET_FILE} ]; then
+        touch ${TARGET_FILE}
     fi
 
     add_debug_grub_entry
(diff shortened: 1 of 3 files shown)


package/harvester-os/files/usr/sbin/harv-install
```diff
@@ -366,16 +366,24 @@ update_grub_settings()
...
-    # PATCH: add /oem/grubenv if it does not exist
-    oem_dir=${TARGET}/oem
-    GRUBENV_FILE="${oem_dir}/grubenv"
-    if ! [ -f ${GRUBENV_FILE} ]; then
-        grub2-editenv ${GRUBENV_FILE} create
-    fi
+    # PATCH1: add /oem/grubenv if it does not exist
+    # PATCH2: add /oem/grubcustom if it does not exist
+    patch_grub_files "grubenv"
+    patch_grub_files "grubcustom"
```
tserong, line 372: `grubenv` is a grub environment block, so that's OK to create with `grub2-editenv`, but `grubcustom` holds the debug grub entry, i.e. it's a menuentry thing, not an environment bock. Would it be better to just `touch grubcustom`?

Vicente-Cheng: Hi @tserong, Thanks for the review!
I thought that is different location, add_debug_grub_entry will create `grubcustom` on `COS_STATE` partition.
Here, we would like to create it on `/oem`.

tserong: My mistake, I'd missed the different directory paths. So I'll go back to my earlier comment that I crossed out :-) and suggest to maybe just `touch grubcustom`, rather than using `grub2-editenv` to create it.

Vicente-Cheng: `touch` could work in this case.
But we would like to make this file fit the grub file format, use `grub2-editenv` to make whole these files keep the same format.

I thought that would not cost too much in the installation stage, so we prefer this.

bk201 (approved):
LGTM!

package/harvester-os/files/usr/sbin/harv-install
tserong: But is grubcustom acutally meant to be in that grub environment format?

I'm just looking at https://github.com/rancher/elemental-toolkit/blob/main/pkg/features/embedded/grub-config/etc/cos/grub.cfg, and it uses `load_env -f ...` to load `grubenv`, but uses `source ...` to load `grubcustom`.

According to https://www.gnu.org/software/grub/manual/grub/html_node/Command_002dline-and-menu-entry-commands.html#Command_002dline-and-menu-entry-commands `load_env` is for loading environment files, and `source` is for loading grub configuration (e.g. the debug menu entry that we have in `/run/initramfs/cos-state/grubcustom`).

Vicente-Cheng: Yes, you are correct.
The `grubcustom` is loaded by `source` command.
We could use touch directly. Update!

Thanks for the detailed checking!

PR #623, opened 2024-01-02
Fix grafana login issue

w13915984028:
**Problem:**

The `grafana` login page showed 404.

A main reason is, when accesing grafana from the local cluster, the general URL is `https://1.2.3.4/k8s/clusters/local/api/v1/....`, the Rancher will trim `/k8s/clusters/local/`, this works fine for normal link. But for `login`, it is another story, the `grafana` will return a `302` code with a new `Location`, unfortunately, the `Rancher` will NOT add `/k8s/clusters/local` back to the `Location`. When grafana's front page code runs, it will get a different `appSubUrl` and assume error, then showed 404 page.

For rancher managed cluster, the regex processing in appSubUrl is not perfect to deal with `302`.

**Solution:**

(1) For local cluster, add a fallback item in the nginx sub-filter
(2) For rancher-managed cluster, use long match in regex to avoid duplicate URL elements.

**Related Issue:**
https://github.com/harvester/harvester/issues/4756

**Test plan:**

1. Install a new Harvester v1.3.0 cluster, enable `rancher-monitoring` addon.   Or use this workaround https://github.com/harvester/harvester/issues/4756#issuecomment-1874242139 in existing v1.2.1/v1.3.0 cluster.
2. Click `grafana` from the `Dashboard` page, which will open a new page.
3. Click `login`, try `login`; or return to previous page, all should run smoothly.

In Rancher-integration scenario, repeat the same processes.

### DIFF
--- a/pkg/config/templates/patch/rancher-monitoring/102.0.0+up40.1.2/nginx-config.yaml
+++ b/pkg/config/templates/patch/rancher-monitoring/102.0.0+up40.1.2/nginx-config.yaml
@@ -91,9 +91,10 @@ data:
       }
 
       map $http_referer $final_appSubUrl {
-        ~.*/k8s/clusters/(c-m-.+)/.*      '"appSubUrl":"/k8s/clusters/$1/api/v1/namespaces/cattle-monitoring-system/services/http:rancher-monitoring-grafana:80/proxy"';
+        ~.*/k8s/clusters/(c-m-.+)/api/v1/namespaces/cattle-monitoring-system/.*      '"appSubUrl":"/k8s/clusters/$1/api/v1/namespaces/cattle-monitoring-system/services/http:rancher-monitoring-grafana:80/proxy"';
         ~.*/dashboard/harvester/c/(c-m-.+)/kubevirt.io.virtualmachine/.*/.* '"appSubUrl":"/k8s/clusters/$1/api/v1/namespaces/cattle-monitoring-system/services/http:rancher-monitoring-grafana:80/proxy"';
         ~.*/dashboard/harvester/c/(c-m-.+)/.*      '"appSubUrl":"/k8s/clusters/$1/api/v1/namespaces/cattle-monitoring-system/services/http:rancher-monitoring-grafana:80/proxy"';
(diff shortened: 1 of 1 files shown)


bk201:
@w13915984028 If the user turns on the add-on in v1.2.1, does this fix work if he upgrades to the newer version?

w13915984028:
@bk201

For v1.2.1 -> v1.3.0,
After https://github.com/harvester/harvester/issues/4940, the `rancher-logging-crd` will have a newer version, the upgrade process will make sure the `rancher-logging-crd managedchart` is upgraded with new value, for all PODs in `cattle-logging-system`, they will be replaced when NODE is rebooted/addon is upgraded, the bug will be fixed after upgrade.

For v1.2.1 -> potential v1.2.2, we will not bump a newer version of `rancher-logging` chart, the `fleet-agent` may not be able to upgrade the `rancher-logging-crd managedchart` as the chart version is not changed, then the configmap is not updated. I will file another issue to track this.

bk201:
> @bk201
> 
> For v1.2.1 -> v1.3.0, After [harvester/harvester#4940](https://github.com/harvester/harvester/issues/4940), the `rancher-logging-crd` will have a newer version, the upgrade process will make sure the `rancher-logging-crd managedchart` is upgraded with new value, for all PODs in `cattle-logging-system`, they will be replaced when NODE is rebooted/addon is upgraded, the bug will be fixed after upgrade.
> 
> For v1.2.1 -> potential v1.2.2, we will not bump a newer version of `rancher-logging` chart, the `fleet-agent` may not be able to upgrade the `rancher-logging-crd managedchart` as the chart version is not changed, then the configmap is not updated. Tracked in [harvester/harvester#4756](https://github.com/harvester/harvester/issues/4756).

Thanks for the explanation, but did you mean the `rancher-monitoring` chart, not the logging one?

w13915984028:
@bk201 You are right :) , it is `rancher-monitoring` not `rancher-logging` related, the description is updated.

bk201 (approved):
LGTM, thanks!

Vicente-Cheng (approved):
LGTM!

bk201:
@mergifyio backport v1.2

PR #630, opened 2024-01-10
installer: rename the role name witness instead etcd

Vicente-Cheng:
**Problem:**
We can use `witness` instead `etcd`. The `witness` seems more suitable for the configuration.

**Solution:**
Change the console and label content. 

**Related Issue:**
https://github.com/harvester/harvester/issues/4840

**Test plan:**
Make sure the interactive console show the `Witness Role`.

And node label should be `node-role.harvesterhci.io/witness: true`

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -277,6 +277,10 @@ func (c *HarvesterConfig) GetKubeletArgs() ([]string, error) {
 		)
 	}
 
+	if c.Role == RoleWitness {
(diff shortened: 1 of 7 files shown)


Vicente-Cheng:
Tested with basic operations and guest cluster.
Please check this PR with the installer part https://github.com/harvester/harvester/pull/4967

ibrokethecloud (approved):
lgtm. thanks.

bk201 (approved):

PR #631, opened 2024-01-12
README.md: Add details of the ISO build and system install process

tserong:
I wrote this up while learning the codebase, so thought I'd better add it to the README file as an aid to future contributors.  This PR also adds a Vagrantfile for faster dev/test of the interactive part of the installer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,28 @@ On the newly installed system, `harvester-installer` remains active
 on the console in order to show the cluster management URL along with
 the current node's hostname and IP address.
 
+## Hacking the Interactive Part of `harvester-installer`
+
+Ordinarily `harvester-installer` needs to be run from a booted ISO
+so it can actually install a system.  But, if you're only working
+on changes to the interactive part of the installer (e.g. adding
+or changing fields, or altering the workflow) and don't actually
+need to perform final installation, the binary can be quickly tested
+using vagrant with vagrant-libvirt.  Here's how:
+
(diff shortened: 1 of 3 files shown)


bk201 (approved):
Thanks, nice enhancement!

README.md
```diff
@@ -2,13 +2,84 @@ harvester-installer
...
+the current node's hostname and IP address.
+
+## Hacking the Interactive Part of `harvester-installer`
+
+Ordinarily `harvester-installer` needs to be run from a booted ISO
+so it can actually install a system.  But, if you're only working
+on changes to the interactive part of the installer (e.g. adding
+or changing fields, or altering the workflow) and don't actually
+need to perform final installation, the binary can be quickly tested
+using vagrant with vagrant-libvirt.  Here's how:
```
bk201, line 69: just have the idea in my mind today, will it be easier to test the executable in a container?

tserong: There's two issues I can think of when running it in a container:
1. It lists disks to install on in the second screen. With a VM we can easily give it a 500G disk to show in that list. I'm not sure how to do the same thing in a container (although I assume there's a way).
2. Part way through the installation, after you select a NIC, it actually applies the network configuration to the system, which I don't think is going to work in a container (so you wouldn't be able to proceed past that screen).

bk201: Thanks for the detailed info. A VM is a better isolate environment here.

PR #634, opened 2024-01-18
installer/config: add the role setup

Vicente-Cheng:
**Problem:**
We tried to support various roles, so the installer need the related changes.

**Solution:**
Add an interactive part for role selection. Also, we need to add the related label for it.

**Related Issue:**
https://github.com/harvester/harvester/issues/4786

**Test plan:**
We can check this feature with two parts as below.
1. 3 node cluster with one worker node. (Should only have one management(master) in the cluster)
  - create a cluster with three nodes
  - choose `worker` role on the second or third node
  - make sure the final cluster only has one management(master)

2. 5 node cluster (to test promote will promote the management role first)
  - create a cluster with five nodes
  - choose `management` role on the fifth node
  - remove the second or third management (master) node
  - it should promote the fifth node instead fourth node

### DIFF
--- a/pkg/config/constants.go
+++ b/pkg/config/constants.go
@@ -8,6 +8,8 @@ const (
 
 	RoleDefault = "default"
 	RoleWitness = "witness"
+	RoleMgmt    = "management"
+	RoleWorker  = "worker"
 
 	NetworkMethodDHCP   = "dhcp"
 	NetworkMethodStatic = "static"
--- a/pkg/console/util.go
+++ b/pkg/console/util.go
@@ -447,6 +447,28 @@ func saveTemp(obj interface{}, prefix string) (string, error) {
 	return tempFile.Name(), nil
 }
 
+func roleSetup(c *config.HarvesterConfig) error {
+	if c.Role == "" {
+		return nil
+	}
+	if c.Labels == nil {
+		c.Labels = make(map[string]string)
+	}
+	switch c.Role {
+	case config.RoleMgmt:
+		c.Labels[util.HarvesterMgmtNodeLabelKey] = "true"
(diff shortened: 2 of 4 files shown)


ibrokethecloud (approved):
lgtm. thanks.

pkg/config/constants.go
```diff
@@ -8,6 +8,8 @@ const (
 
 	RoleDefault = "default"
 	RoleWitness = "witness"
+	RoleMgmt    = "mgmt"
```
bk201, line 11: how about `management` to be consistent with the label `HarvesterMgmtNodeLabelKey`.

Vicente-Cheng: updated!

pkg/console/util.go
```diff
@@ -447,6 +447,25 @@ func saveTemp(obj interface{}, prefix string) (string, error) {
...
+	}
+	switch c.Role {
+	case config.RoleMgmt:
+		c.Labels[util.HarvesterMgmtNodeLabelKey] = "true"
+	case config.RoleWorker:
+		c.Labels[util.HarvesterWorkerNodeLabelKey] = "true"
+	case config.RoleWitness:
+		c.Labels[util.HarvesterWitnessNodeLabelKey] = "true"
+	default:
+		logrus.Warnf("unknown role %s", c.Role)
```
bk201, line 465: Can we add a check in validato.go if an inputted role is unsupported.

Vicente-Cheng: Updated!
Now supported role: `default`, `management`, `witness` and `worker`

If the user does not set up a role, it will be default.

bk201 (approved):

PR #635, opened 2024-01-18
Don't cut off last line of text in panel above select widget

tserong:
**Problem:**
There's an off-by-one (or possibly off-by-two) error in the current implementation of the select widget, when there's content in the panel above it.  This means the last line of text in the panel is overwritten by the select widget (i.e. the select widget is one line too high). See the below screenshot where the text finishes with `Your disk will be formatted and Harvester will be installed with`:

installer-1 2 1-confirm-cutoff

**Solution:**
This commit changes the logic slightly to offset the select widget by the number of lines of panel content plus one, allowing all the text in the panel to be shown. Now we can see the full confirmation message plus a blank line above the select widget:

installer-offset-confirm-fixed

**Related Issue:**
None (should I open one, or is it OK to just randomly fix things?)

**Test plan:**
- Run through the installer.
- Verify that select widgets with _no_ content above still look sensible (i.e. no extra blank lines above the select widget on the "Choose installation mode" screen, and the NIC select widget appears exactly on top of the "Management NIC" field)
- Verify that the text isn't cut off on the final "Confirm installation options" screen.

### DIFF
--- a/pkg/widgets/select.go
+++ b/pkg/widgets/select.go
@@ -47,9 +47,12 @@ func (s *Select) Show() error {
 		return err
 	}
 	optionViewName := s.Name + "-options"
-	offset := len(strings.Split(s.Content, "\n"))
-	y0 := s.Y0 + offset - 1
-	y1 := s.Y0 + offset + len(s.options)
+	offset := 0
+	if len(s.Content) > 0 {
+		offset = len(strings.Split(s.Content, "\n")) + 1
+	}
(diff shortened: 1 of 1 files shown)


Vicente-Cheng (approved):
good catch! LGTM!

ibrokethecloud (approved):
lgtm. thanks.

bk201:
@mergifyio backport v1.2

PR #636, opened 2024-01-18
Add hardware spec preflight checks

tserong:
**Problem:**
We need to run some checks to ensure harvester is being installed on minimum supported hardware.

**Solution:**
This adds three preflight checks:
    
- Number of CPU cores
- Available memory
- Whether the system is running virtualized or not
    
Each check will result in a warning message if it doesn't pass. If there are any warning messages, a new page is inserted at the start of the installer listing the warnings and asking the user if they wish to proceed.  If the user selects "no", the system is rebooted. If there are no warnings, this page does not appear.
    
Outstanding issues:
    
- There are no unit tests.
- The warning messages are quite long, and wrap oddly on very small screens/terminal windows.
- The preflight checks have no effect for PXE/automated installs. For this path we should fail the install if there's any warnings, but allow overriding with a flag (see https://github.com/harvester/harvester/issues/1154#issuecomment-1888723063)
- More checks?

I'm opening this PR now even though there's still outstanding work in case anyone would like to critique the approach I've taken here.

Note that this PR also includes the Vagrantfile from #631 and the select widget fix from #635. I'll rebase once those are merged.

**Related Issue:**

**Test plan:**

Requires three systems:

1. One meeting production requirements (>=16 CPU cores, >=64GiB RAM, bare metal, i.e. not running under virtualization)
2. One meeting testing requirements (8-15 CPU cores, 32-60GiB RAM, can be virtualized)
3. One that doesn't meet any of those requirements (<8 CPU cores, <32GiB RAM, can be virtualized)

Boot the Harvester ISO on each system:

1. For the system that meets production requirements, verify that the installer starts as normal on the console with no hardware check warnings, i.e. first screen is "Choose installation mode".
2. For the system that meets testing requirements, verify that the first screen of the installer shows the Hardware Checks, similar to the following:
```
8 CPU cores detected. Harvester requires at least 16 cores for production use.
3 GiB RAM detected. Harvester requires at least 64GiBfor production use.
System is virtualized (kvm) which is not supported.

Do you wish to proceeed?
```
3. For the system that doesn't meet any requirements, it should behave the same as point 2, but with warning messages calling out the testing limits, e.g.:
```
Only 4 CPU cores detected. Harvester requires at least 8 cores for testing and 16 for production use.
Only 7GiB RAM detected. Harvester requires at least 32GiB for testing and 64GiB for production use.
System is virtualized (kvm) which is not supported.
```
4. For both the above cases, selecting "Yes" to proceed should take you to the normal first page of the installer ("Choose installation mode"). Selecting "No" should reboot the system.

To verify PXE boot:

1. Deploy a system that meets the production requirements via PXE - it should deploy just fine.
2. Deploy a system that only meets the testing requirements (or doesn't meet any requirements). Ensure the `harvester.install.skipchecks=true` kernel command line parameter is _not_ specified. Installation should fail, with the above hardware check warning messages printed to the console, and also in `/var/log/console.log` in the installation environment.
3. Deploy a system that only meets the testing requirements, but use the `harvester.install.skipchecks=true` kernel command line parameter. Installation should succeed, but once the system is installed, the hardware check warnings should be present in `/oem/install/console.log` (try `grep warning /oem/install/console.log`)

**Sample Screenshots:**

Preflight warnings on a VM with 4 CPU cores and 32GiB RAM:

installer-preflight-warnings

Preflight warnings on a ridiculously small VM (512MB / 1 CPU), in a small terminal window, to demonstrate the poor wrapping of the long messages:

installer-preflight-tiny-vm

I guess the easy way to fix the above is to just put explicit `'\n'` characters in the messages to cap them at about 60 columns, but then they'd look weird on large screens - it'd be nicer to maybe tweak `Panel.SetContent()` to automatically insert a linebreak in place of a space character in lines that exceed the actual width of the panel.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -129,6 +129,7 @@ type HarvesterChartValues struct {
 
 type Install struct {
 	Automatic           bool    `json:"automatic,omitempty"`
+	SkipChecks          bool    `json:"skipchecks,omitempty"`
 	Mode                string  `json:"mode,omitempty"`
 	ManagementInterface Network `json:"managementInterface,omitempty"`
 
--- a/pkg/preflight/checks.go
+++ b/pkg/preflight/checks.go
@@ -19,11 +19,11 @@ const (
 	MinMemoryProd = 64
 )
 
-// The Run() method of a PreflightCheck returns a string.  If the string
+// The Run() method of a preflight.Check returns a string.  If the string
 // is empty, it means the check passed.  Otherwise, the string contains
 // some text explaining why the check failed.  The error value will be set
 // if the check itself failed to run at all for some reason.
-type PreflightCheck interface {
+type Check interface {
 	Run() (string, error)
 }
 
--- a/pkg/preflight/checks.go
+++ b/pkg/preflight/checks.go
@@ -19,6 +19,12 @@ const (
 	MinMemoryProd = 64
 )
 
+var (
+	// So that we can fake this stuff up for unit tests
+	execCommand = exec.Command
+	procMemInfo = "/proc/meminfo"
+)
+
 // The Run() method of a preflight.Check returns a string.  If the string
 // is empty, it means the check passed.  Otherwise, the string contains
 // some text explaining why the check failed.  The error value will be set
@@ -32,7 +38,7 @@ type MemoryCheck struct{}
 type VirtCheck struct{}
 
 func (c CPUCheck) Run() (msg string, err error) {
-	out, err := exec.Command("/usr/bin/nproc", "--all").CombinedOutput()
+	out, err := execCommand("/usr/bin/nproc", "--all").Output()
 	if err != nil {
 		return
 	}
@@ -48,7 +54,7 @@ func (c CPUCheck) Run() (msg string, err error) {
 }
 
 func (c MemoryCheck) Run() (msg string, err error) {
-	meminfo, err := os.Open("/proc/meminfo")
+	meminfo, err := os.Open(procMemInfo)
 	if err != nil {
 		return
 	}
@@ -95,7 +101,7 @@ func (c MemoryCheck) Run() (msg string, err error) {
 }
 
 func (c VirtCheck) Run() (msg string, err error) {
-	out, err := exec.Command("/usr/bin/systemd-detect-virt", "--vm").CombinedOutput()
+	out, err := execCommand("/usr/bin/systemd-detect-virt", "--vm").Output()
 	virt := strings.TrimSpace(string(out))
 	if err != nil {
 		// systemd-detect-virt will return a non-zero exit code
--- a/pkg/preflight/checks.go
+++ b/pkg/preflight/checks.go
@@ -1,0 +1,114 @@
+package preflight
+
+import (
+	"bufio"
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"strconv"
+	"strings"
+)
+
+const (
+	// Constants here from Hardware Requirements in the documentaiton
+	// https://docs.harvesterhci.io/v1.3/install/requirements/#hardware-requirements
+	MinCPUTest    = 8
+	MinCPUProd    = 16
+	MinMemoryTest = 32
+	MinMemoryProd = 64
+)
+
+// The Run() method of a PreflightCheck returns a string.  If the string
+// is empty, it means the check passed.  Otherwise, the string contains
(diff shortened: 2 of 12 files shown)


tserong:
I figured out how to automatically insert linebreaks in long lines of text :-)

image

This also allows removing the hard linebreaks from the long lines on the confirmation page:

image

image

tserong:
Some comments on possible other checks based on our hardware requirements (https://docs.harvesterhci.io/v1.3/install/requirements):

1. Disk capacity is already handled on the installation disk screen (although I'm not sure what happens if you do a PXE install on disks that are to small).
2. I'm not sure how to test disk performance in a nondestructive way prior to installation.
3. Network card speed check could maybe be inserted after/on the Configure network screen, but that's only looking at the Management NIC, so I'm not sure if it's worthwhile.
4. I have no idea how I'd test anything related to an external Network Switch
5. The requirement for each node to have a unique `/sys/class/dmi/id/product_uuid` would have to be done on nodes _joining_ the cluster, and would require querying all the other nodes somehow.

tserong:
...and at long last, some unit tests.

Vicente-Cheng (approved):
just a discussion here.
overall lgtm, thanks for improving so much panel-related and preflight checks!

pkg/config/config.go
```diff
@@ -129,6 +129,7 @@ type HarvesterChartValues struct {
 
 type Install struct {
 	Automatic           bool    `json:"automatic,omitempty"`
+	SkipChecks          bool    `json:"skipchecks,omitempty"`
```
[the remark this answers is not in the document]
tserong: > IIRC, the default bool variable will be false. This means if we do not set `harvester.install.skipchecks=true` it will run check by default.

Correct.

> I know we introduced this parameter, but some older install scripts (maybe an ansible yaml file) would not notice that change. So, they will face some trouble if their hardware does not fit our expectations.

Yes, older install scripts would need to be updated to add this parameter.

> Could we default skip check instead? (pxe-only) This means only automatic=true -> then the default check is false. That will not break the older automation deployment. WDYT?

A couple of thoughts:

1. If we default to skipping for PXE, my concern is that nobody will actually bother to enable the checks :-) as PXE users would have to explicitly set `harvester.install.skipchecks=false` if they wanted that behaviour. If I'm right that few PXE users will turn the checks on, we could just get rid of the flag altogether, and always skip for PXE (with failed checks still logged to `/var/log/console.log`).
2. Alternately, we could keep the flag, set it to default skip for automatic installs for the 1.3.0 release, then later (say, the 1.4.0 release?), change the default back to false. This means we can ensure it's well documented in 1.3.0 that these new checks exist, and that you'll need to prepare any deployment scripts for a behaviour change in the _next_ release (1.4.0).

I don't really have a strong preference either way. WDYT?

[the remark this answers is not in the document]
bk201: The concern makes sense. We need to make sure the change is well-mentioned in the release note. Record in the issue https://github.com/harvester/harvester/issues/1154#issuecomment-1909194169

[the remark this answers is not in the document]
Vicente-Cheng: Thanks for the explanation. That makes sense to me! Let's keep the current design.

pkg/preflight/checks.go
```diff
@@ -0,0 +1,120 @@
...
+			memReported, MinMemoryTest, MinMemoryProd)
+	} else if float32(memTotalGiB) < (MinMemoryProd * 0.95) {
+		msg = fmt.Sprintf("%s RAM detected. Harvester requires at least %dGiB for production use.",
+			memReported, MinMemoryProd)
+	}
+	return
+}
+
+func (c VirtCheck) Run() (msg string, err error) {
+	out, err := execCommand("/usr/bin/systemd-detect-virt", "--vm").Output()
```
bk201, line 104: nit: Besides checking if a system is in a virtualized environment, do we also want to check if the host is capable of running hardware-accelerated VMs? (like the kvm-ok command does).

tserong: Good idea, I'll see what I can do. Do you want me to add that to this PR, or handle it separately?

bk201: Both work, it's OK to handle this in another PR.

bk201 (approved):

pkg/preflight/checks.go
tserong: Cool, let's do that bit in a separate PR.

PR #637, opened 2024-01-18
feat: add precedent image lists into iso image

starbops:
**Problem:**

There's no way for Harvester itself to know the image list of each release. Such information is useful for the image cleanup mechanism introduced in harvester/harvester#4995.

**Solution:**

This will add all the previous versions' published image lists into the ISO image according to the upgrade matrix file in harvester repo

Based on [upgrade-helpers/bin/harv-purge-images.sh at d490e51c76b107c8d5df2df71d0e9b7000b392d6 · harvester/upgrade-helpers](https://github.com/harvester/upgrade-helpers/blob/d490e51c76b107c8d5df2df71d0e9b7000b392d6/bin/harv-purge-images.sh), downloading and adding all the previous versions' published image lists into the ISO image that is going to be built. These image lists will be used later for the upgrade controller to purge the unneeded container images.

**Related Issue:**

harvester/harvester#4425

**Test plan:**

Please reference https://github.com/harvester/harvester/issues/4425#issuecomment-1900566020

### DIFF
--- a/scripts/archive-images-lists.sh
+++ b/scripts/archive-images-lists.sh
@@ -1,0 +1,45 @@
+#!/usr/bin/env sh
+set -e
+
+UPGRADE_MATRIX_FILE=$1
+IMAGES_LISTS_DIR=$2
+RANCHERD_IMAGES_DIR=$3
+IMAGES_LISTS_ARCHIVE_DIR=$4
+
+WORKING_DIR=$(mktemp -d)
+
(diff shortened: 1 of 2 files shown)


Vicente-Cheng (approved):
LGTM, tested with https://github.com/harvester/harvester/pull/4995
After upgrade, the old images were deleted.

bk201 (approved):

starbops:
@mergify backport v1.2

PR #639, opened 2024-01-19
Update rancherd-12-monitoring-dashboard.yaml

ulikl:
**Problem:**
The panel titles "Network Receive Bytes" and "Network Transmit Bytes" are a bit misleading as the values are in Bits/s. And Storage Write Traffic Bytes" is really in Bytes/s.
 See expr : "metrics* 8!"
The units where already correctly shown in "MB/s".

**Solution:**
 Rename panel titles to "Bits"

**Related Issue:**

### DIFF
--- a/pkg/config/templates/rancherd-12-monitoring-dashboard.yaml
+++ b/pkg/config/templates/rancherd-12-monitoring-dashboard.yaml
@@ -497,7 +497,7 @@ resources:
             "timeFrom": null,
             "timeRegions": [],
             "timeShift": null,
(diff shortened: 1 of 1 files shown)


bk201:
@w13915984028 I'm not entirely sure, but this perhaps better fix in upstream

w13915984028 (approved):
LGTM,  @ulikl  thanks very much for this PR.

ibrokethecloud (approved):
lgtm. thanks.

PR #642, opened 2024-01-24
Feat: create a net-install ISO

bk201:
**Problem:**

To create a net-install ISO.
- Save time for downloading the installation medium.
- (hopefully) Prevent issues on buggy BMC ISO redirection feature: https://github.com/harvester/harvester/issues/2651

**Solution:**

We can ship a net-install ISO without image tarballs since we already https://github.com/harvester/harvester/issues/2670.

**Related Issue:**
https://github.com/harvester/harvester/issues/4295

**Test plan:**

- Build the code
- Boot the net-install ISO (Its name is similar to `harvester-74e9450-dirty-amd64-net-install.iso`) and verify the installation. Both interactive mode and ipxe mode should work.

**Additional context**

- Add a kernel parameter `harvester.install.with_net_images=true` to GRUB menuentry in the net-install ISO. The installer only preloads the `harvester-cluste-repo` image if it sees the parameter. We still preload this image to ease development.
- The kernel parameter `harvester.install.with_net_images=true` also works for the ordinary ISO.

### DIFF
--- a/package/harvester-os/files/usr/sbin/harv-install
+++ b/package/harvester-os/files/usr/sbin/harv-install
@@ -282,6 +282,13 @@ preload_rancherd_images()
     cp ${ISOMNT}/bundle/rancherd/images/* $TARGET/var/lib/rancher/agent/images
 }
 
+defer_preload_images()
+{
+    echo "Save image tarball(s) to $TARGET/var/lib/rancher/rke2/agent/images"
+    mkdir -p $TARGET/var/lib/rancher/rke2/agent/images
+    cp ${ISOMNT}/bundle/harvester/images/*.tar.zst $TARGET/var/lib/rancher/rke2/agent/images
+}
+
 do_preload()
 {
     # Bind mount persistent folder to preload images
@@ -294,8 +301,14 @@ do_preload()
         mount -o bind $TARGET/$state_dir $TARGET/$i
     done
 
-    preload_rancherd_images
-    preload_rke2_images
+    if [ -n "$HARVESTER_WITH_NET_IMAGES" ]; then
+        # Preload images after the installing node reboots and RKE2 starts
+        defer_preload_images
+    else
+        # Preload all images now
+        preload_rancherd_images
+        preload_rke2_images
+    fi
 
     for i in ${BIND_MOUNTS[@]}; do
         umount $TARGET/$i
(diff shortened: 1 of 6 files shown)


tserong (approved):
Neat!

I did a test build here and got an 839MB net-install ISO, vs. the 5.6GB regular ISO, and it runs through the interactive install quite quickly. After reboot it took a bit over 30 minutes to complete setup, although this is of course presumably due to downloading all those images over my Tasmanian internet connection :-)

A couple of questions:
- Do we need to document the net-install ISO somewhere?
- Can this work in an airgap environment, e.g.: via a proxy server?
- Is there a good way to get better deployment progress than just looking at the console and waiting for it to show "Ready"? (I know I can ssh in and do things like `journalctl -f` and `kubectl get pods -A ; kubectl get deployments -A`, or run `k9s`... but I don't know how to tell exactly how far along we are...)

bk201:
Thanks for the reviewing!

> 
> A couple of questions:
> 
> * Do we need to document the net-install ISO somewhere?

Good idea, I'll mention the change in the release note (there is a section **ISO**, e.g.: https://github.com/harvester/harvester/releases/tag/v1.2.1) that we have such an ISO.

> * Can this work in an airgap environment, e.g.: via a proxy server?

The user can mirror all images to a private registry and:
- Create `/etc/rancher/rke2/registries.yaml` for [RKE2](https://docs.rke2.io/install/containerd_registry_configuration)
- Create `/etc/rancher/common/registries.yaml` for wharfie

I guess this might work, but need to verify.

> * Is there a good way to get better deployment progress than just looking at the console and waiting for it to show "Ready"? (I know I can ssh in and do things like `journalctl -f` and `kubectl get pods -A ; kubectl get deployments -A`, or run `k9s`... but I don't know how to tell exactly how far along we are...)

That's one way except there are also daemon sets. Another way is to check the fleet bundles:
```
kubectl get bundles -A
NAMESPACE     NAME                         BUNDLEDEPLOYMENTS-READY   STATUS
fleet-local   fleet-agent-local            1/1
fleet-local   local-managed-system-agent   1/1
fleet-local   mcc-harvester                1/1
fleet-local   mcc-harvester-crd            1/1
fleet-local   mcc-rancher-logging-crd      1/1
fleet-local   mcc-rancher-monitoring-crd   1/1
```

Fleet keeps monitoring all resources in a helm are ready.

Vicente-Cheng (approved):
LGTM, tested with installation and basic operations work well.

PR #649, opened 2024-02-02
Immediately validate the persistent size

votdev:
... and clear the error message when necessary.

Peek 2024-02-02 13-43

**Problem:**
Error message is not cleared and validation is only done before page change is triggered.

**Solution:**
The error message is cleared when necessary and the validation of the `Persistent size` is done immediately when the input field loses focus (enter or key-down).

**Related Issue:**
https://github.com/harvester/harvester/issues/5096

**Test plan:**
- Go to the disk setup page
- Choose `Use the installation disk` for `Data disk`
- Go to the `Persistent size` field and enter a value < 150GiB
- Press `Enter` or `Key-Down`
- The error message `Partition size is too small. Minimum 150Gi is required` must be displayed
- Changing value to >= 150GiB the error message must disappear.

### DIFF
--- a/pkg/console/install_panels.go
+++ b/pkg/console/install_panels.go
@@ -433,6 +433,20 @@ func addDiskPanel(c *Console) error {
 		}
 		return true, nil
 	}
+	validatePersistentPartitionSize := func(persistentSize string) (bool, error) {
+		installDisk := c.config.Install.Device
+		dataDisk := c.config.Install.DataDisk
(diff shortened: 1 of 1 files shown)


tserong (approved):
LGTM!

FrankYang0529 (approved):
LGTM. Thanks for the PR.

bk201 (approved):

PR #653, opened 2024-02-08
Add preflight check for network card speed

tserong:
**Problem:**
We need to run some checks to ensure harvester is being installed on minimum supported hardware.

**Solution:**
This check ensures that the network interfaces configured during installation are all 10Gbps.  If they're less than this, during interactive install, a warning message will be displayed on the network screen.  For automated installs, as with other preflight checks, failure will abort the install unless
`harvester.install.skipchecks=true` is set.

nic-speed-test

A few oddities worth mentioning:

* We can't do the network speed check upfront with the other checks because we don't know which interfaces to test until we're on the network screen (or have them from the remote config).

* I've moved `c.config.SkipChecks = true` from the `KeyEnter` handler in `addPreflightCheckPanel()` to `askCreateV.PreShow()`.  Imagine the case where the initial preflight checks all pass (so that panel isn't shown), but later the network speed check fails.  In this case, `SkipChecks` wouldn't be set, and the installation would abort in the `installV.PreShow()` function.

* virtio NICs when testing under libvirt don't report speed via `/sys/class/net/$IFACE/speed` (or, rather, the speed is set to -1) so the test is skipped in this case.  Other NIC types under libvirt (e.g. e1000) all seem to be 1Gbps only.

**Related Issue:**


ibrokethecloud (approved):
lgtm. thanks.

Vicente-Cheng (approved):
overall lgtm, thanks!

Just would like to bring some points like the 2.5G NICs.
If we have 2 x 2.5G NICs, maybe it could be set up as a production environment in a specific configuration.

tserong:
Thanks @Vicente-Cheng. I've opened a separate issue to clarify the network requirements in more detail.

bk201:
@votdev Sorry I didn't notice you are on the reviewer list. Please let us know if you want something fixed.

PR #654, opened 2024-02-09
Clarify wording of VirtCheck failure, fix weird line wrapping

tserong:
**Problem:**
When running under virtualization the current warning message says "System is virtualized (kvm) which is not supported", but should say "not supported in production". Also, in some cases, there's extra linebreaks in the warning messages, e.g.:

image (6)

**Solution:**
This PR

**Related Issue:**

**Test plan:**
N/A

### DIFF
--- a/pkg/widgets/panel.go
+++ b/pkg/widgets/panel.go
@@ -139,7 +139,9 @@ func (p *Panel) SetLocation(x0, y0, x1, y1 int) {
 }
 
 func (p *Panel) SetContent(content string) {
-	panelWidth := p.X1 - p.X0
+	// Need to subtract 1 from panelWidth here to match the view.Size()
+	// calculation done in gocui's view.Draw() function
+	panelWidth := p.X1 - p.X0 - 1
 	if panelWidth > 0 {
 		p.Content = ""
 		lines := strings.Split(content, "\n")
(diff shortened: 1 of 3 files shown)


w13915984028 (approved):
LGTM, thanks.

pkg/widgets/panel.go
```diff
@@ -139,7 +139,9 @@ func (p *Panel) SetLocation(x0, y0, x1, y1 int) {
 }
 
 func (p *Panel) SetContent(content string) {
-	panelWidth := p.X1 - p.X0
+	// Need to subtract 1 from panelWidth here to match the view.Size()
+	// calculation done in gocui's view.Draw() function
+	panelWidth := p.X1 - p.X0 - 1
```
connorkuehl, line 144: Is there any chance at all of this becoming negative? Would that be bad or does gocui do the right thing and just occlude anything that's not on the page?

connorkuehl (approved):

pkg/widgets/panel.go
tserong: I don't believe it's possible for that to ever end up negative, but just in case I'm wrong there's an `if panelWidth > 0` to ensure our wrapping doesn't kick in if the dimensions are bogus :-)

w13915984028:
FYI: With the fix #654 
image

PR #656, opened 2024-02-13
Arm build support

ibrokethecloud:
**Problem:**

PR introduces experimental support for arm hosts

**Solution:**

**Related Issue:**
https://github.com/harvester/harvester/issues/371
**Test plan:**

### DIFF
--- a/scripts/build-bundle
+++ b/scripts/build-bundle
@@ -17,6 +17,7 @@ source ${SCRIPTS_DIR}/lib/image
 source ${SCRIPTS_DIR}/hack/patch-rancher-monitoring
 source ${SCRIPTS_DIR}/hack/patch-rancher-logging
 source ${SCRIPTS_DIR}/hack/patch-rancher-monitoring-crd
+source ${SCRIPTS_DIR}/hack/patch-harvester-chart
 
 BUNDLE_DIR="${PACKAGE_HARVESTER_OS_DIR}/iso/bundle"
 CHARTS_DIR="${PACKAGE_HARVESTER_REPO_DIR}/charts"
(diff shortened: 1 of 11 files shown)


ibrokethecloud:
Additional work is still needed to identify correct arm based kubevirt images from registry.opensuse.org

The build currently patches harvester chart to download images from `quay.io/kubevirt` and packages the same with harvester.

scripts/build-bundle
```diff
@@ -58,6 +59,15 @@ ${SCRIPTS_DIR}/patch-harvester ${TOP_DIR}/../harvester
 # Package harvester chart
 harvester_chart_path=${harvester_path}/deploy/charts/harvester
 harvester_crd_chart_path=${harvester_path}/deploy/charts/harvester-crd
+
+## for arm based builds we need to patch registry to registry.opensuse.com
```
tserong, line 63: Comment needs updating (`patch_harvester_chart` is using quay.io ATM)

scripts/package-harvester-os
```diff
@@ -18,6 +18,7 @@ source ${SCRIPTS_DIR}/version-monitoring
 source ${SCRIPTS_DIR}/version-logging
 source ${SCRIPTS_DIR}/lib/iso
 
+## Revert back to sle-micro-head
```
tserong, line 21: Spurious comment?

tserong (approved):
LGTM - all I have is comments about comments ;-)

bk201:
@ibrokethecloud `set -ex` is turn on in multiple scripts, do we need to be verbose?

Vicente-Cheng (approved):
LGTM.

There is just one point I would like to raise.

There are too many customized operations with different arch.
We might refine the iso building flow in the next version.

Thanks!

Vicente-Cheng:
BTW, could we squash some commits to make the tracking easier?

bk201 (approved):

PR #657, opened 2024-02-14
Fix CVE and bump monitoring to 103.0.3+up45.31.1

w13915984028:
**Problem:**

Per https://github.com/harvester/security/issues/44, we need to bump rancher-monitoring version, use latest shell image

**Solution:**

1. Bump to `rancher-monitoring version 103.0.3+up45.31.1`
2. Patch `rancher/shell` to v0.1.22 in Harvester (note: the yq can also replace the image tag, but it will remove all black line from the original yaml file, the `patch` is used instead)
3. Set related version in addon

**Related Issue:**
https://github.com/harvester/security/issues/44

**Test plan:**

1. Install new cluster, enable `rancher-monitoring`
2. Upgrade from v1.2.1

**Local test:**
1. Cluster is installed as expected
2. Rancher-monitoring add is successfully enabled/disabled
4. `With crctrl image ls`, before/after the first time eanbling rancher-monitoring, no additional contaier images are imported
5. Grafana metrics, login ... all work


ibrokethecloud (approved):
lgtm. thanks.

scripts/hack/patch-rancher-monitoring
```diff
@@ -33,6 +33,20 @@ patch_rancher_monitoring_chart()
   echo "patched file"
   ls -alth "$origfile"
 
+  local valuesfile="./rancher-monitoring/values.yaml"
+  local difffile="${pkg_monitoring_path}/${monitoring_version}/patch-values-shell-version.diff"
+  echo "patch rancher/shell image version"
+  echo "the current values.yaml has following shell image version"
+  grep "rancher/shell" $valuesfile -1 || true
+  if [ -f "$difffile" ]; then
+    ls -alth "$difffile"
```
Vicente-Cheng, line 42: redundant?

w13915984028: Good catch :), updated.

Vicente-Cheng (approved):
LGTM!

PR #658, opened 2024-02-14
Fix CVE and bump logging to 103.0.0+up3.17.10

w13915984028:
**Problem:**

Per https://github.com/harvester/security/issues/44, we need to bump rancher-logging version, use latest fluentbit image.

**Solution:**

1. Bump to rancher-logging version 103.0.0+up3.17.10
1. Patch fluentbit image to v2.2.0 in Harvester instead of v1.9.5
1. Set related version in addon

**Related Issue:**
https://github.com/harvester/security/issues/44

**CVE Fix**:
https://security-tracker.debian.org/tracker/source-package/glibc
bullseye (security)	2.31-13+deb11u7	fixed

[CVE-2023-4911](https://security-tracker.debian.org/tracker/CVE-2023-4911)	A buffer overflow was discovered in the GNU C Library's dynamic loader ...

In fluent-bit v2.2.0, we have ` (Debian GLIBC 2.31-13+deb11u7) `, which is with above fix.
(taken from a fluent-bit v2.2.0-debug image, which has shell available)
```
$ ldd --version
ldd (Debian GLIBC 2.31-13+deb11u7) 2.31
Copyright (C) 2020 Free Software Foundation, Inc.
```

**Test plan:**

- Install new cluster, enable rancher-logging addon
- Upgrade from v1.2.1

Local test:

1. Cluster is installed as expected
2. Rancher-logging addon is successfully enabled/disabled
3. With `crictl image ls`, before/after the first time enabling rancher-logging, no additional container images are imported

**Node:**

Due to this chart https://charts.rancher.io/assets/rancher-logging/rancher-logging-103.0.1+up3.17.10.tgz is not available as of 15.02.2024

We fallback to 103.0.0+up3.17.10 and patch the only difference: use fluentbit image v2.2.0 instead of v1.9.5


markhillgit (approved):
we do what we must! looks reasonable.

ibrokethecloud (approved):
lgtm. thanks.

Vicente-Cheng (approved):
lgtm, thanks!

PR #660, opened 2024-02-16
additional installer option to wipe all disks before install

ibrokethecloud:
**Problem:**

During out internal automation testing it is possible different disks were used to install of harvester.
When the disk is changed the older installation is left intact which can cause unexpected results during reboots.
To ensure this does not happen we need to be able to allow wiping of disks via the installer.

**Solution:**

PR adds an additional option in harvester config `wipeDisks` which can be passed via a config url or kernel arguments `harvester.install.wipe_disks=true` and will result in the harv-install script wiping all disks on the node.

**Related Issue:**
https://github.com/harvester/harvester/issues/2066
https://github.com/harvester/harvester/issues/2781
https://github.com/harvester/harvester/issues/4527

**Test plan:**

### DIFF
--- a/package/harvester-os/files/usr/sbin/harv-install
+++ b/package/harvester-os/files/usr/sbin/harv-install
@@ -7,17 +7,46 @@ DATA_DISK_FSLABEL="HARV_LH_DEFAULT"
 
 clear_disk_label()
 {
-    # Clear the label of partitions that has $DATA_DISK_FSLABEL to prevent misidentification
-    # Also, while yip is partitioning the disk, if it sees the LABEL to be used exists,
-    # it won't create the partition. So it's necessary to clear the label
-    for part in $(blkid -t LABEL="$DATA_DISK_FSLABEL" -o device); do
-        echo "Remove filesystem label from $part"
-        # Run this tune2fs twice because sometimes the first run would show "Recovering journal"
-        # and label is not modified
-        tune2fs -L "" $part > /dev/null
-        tune2fs -L "" $part > /dev/null
-    done
-    udevadm settle
+    # For internal automation environments we need to be able to wipe all disks on the node
+    # to ensure that host does not boot off partitions from another disk in the host
+    # lsblk produces the following json output, which we filter using jq to identify disks
+    # which we will be wiping
+    # lsblk -d -n -J -o NAME,TYPE
+    # {
+    #  "blockdevices": [
+    #  {
+    #     "name": "loop0",
+    #     "type": "loop"
+    #  },{
+    #     "name": "sr0",
+    #     "type": "rom"
+    #  },{
+    #     "name": "vda",
+    #     "type": "disk"
+    #     }
+    #   ]
+    # }
+    if [ "$HARVESTER_WIPE_DISKS" == "true" ]; then
+        echo "wiping all disks"
+        for disk in $(lsblk -d -n -J -o NAME,TYPE | jq -r '.blockdevices[] | select(.type == "disk") | .name')
+        do
+            sgdisk -Z /dev/$disk
+            partprobe -s /dev/$disk
+        done
+    else
+        # Clear the label of partitions that has $DATA_DISK_FSLABEL to prevent misidentification
+        # Also, while yip is partitioning the disk, if it sees the LABEL to be used exists,
+        # it won't create the partition. So it's necessary to clear the label
+        for part in $(blkid -t LABEL="$DATA_DISK_FSLABEL" -o device); do
+            echo "Remove filesystem label from $part"
+            # Run this tune2fs twice because sometimes the first run would show "Recovering journal"
+            # and label is not modified
+            tune2fs -L "" $part > /dev/null
+            tune2fs -L "" $part > /dev/null
+        done
+        udevadm settle
+    fi
+   
 }
 
 umount_target() {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -149,6 +149,7 @@ type Install struct {
 	ForceGPT      bool   `json:"forceGpt,omitempty"`
 	Role          string `json:"role,omitempty"`
 	WithNetImages bool   `json:"withNetImages,omitempty"`
+	WipeDisks     bool   `json:"wipeDisks,omitempty"`
 
 	// Following options are not cOS installer flag
 	ForceMBR bool   `json:"forceMbr,omitempty"`


Vicente-Cheng (approved):
LGTM!

package/harvester-os/files/usr/sbin/harv-install
```diff
@@ -7,6 +7,32 @@ DATA_DISK_FSLABEL="HARV_LH_DEFAULT"
...
+    #     "type": "disk"
+    #     }
+    #   ]
+    # }
+    if [ "$HARVESTER_WIPE_DISKS" == "true" ]; then
+       echo "wiping all disks"
+       for disk in $(lsblk -d -n -J -o NAME,TYPE | jq -r '.blockdevices[] | select(.type == "disk") | .name')
+       do
+         sgdisk -Z /dev/$disk
+       done
```
Vicente-Cheng, line 34: just a nit: we can do `partprobe` here. For rescanning the whole partition.

ibrokethecloud: done

innobead:
cc @m-ildefons

package/harvester-os/files/usr/sbin/harv-install
```diff
@@ -7,6 +7,32 @@ DATA_DISK_FSLABEL="HARV_LH_DEFAULT"
...
+    else
```
tserong, line 36: nit: block below wants indenting

tserong (approved):
Just doing a build to give this a quick test, but LGTM.

package/harvester-os/files/usr/sbin/harv-install
```diff
@@ -7,6 +7,33 @@ DATA_DISK_FSLABEL="HARV_LH_DEFAULT"
 
 clear_disk_label()
 {
+    # For internal automation environments we need to be able to wipe all disks on the node
+    # to ensure that host does not boot off partitions from another disk in the host
+    # lsblk produces the following json output, which we filter using jq to identify disks
+    # which we will be wiping
+    # lsblk -d -n -J -o NAME,TYPE
+    #{
+    #"blockdevices": [
```
starbops, line 16: nit(style): missing spaces after comment mark

starbops (approved):
LGTM, thank you!

tserong (changes requested):
Hang on... I just tried this in a previously used VM, and hit an error. It was a manual install, not PXE, but I edited the kernel args in the grub menu to add `harvester.install.wipe_disks=true` to get this PR into play.  Here's what I ended up with:

image

Note that the wipe _did_ succeed - previously /dev/vdb and /dev/vdc had a bunch of partitions on them, and now they're gone:

```
# cat /proc/partitions 
major minor  #blocks  name

   7        0     581632 loop0
 253        0  125829120 vda
 253       16 4294967296 vdb
 253       32  209715200 vdc
 253       48   20971520 vdd
  11        0    5863104 sr0
```

I'll re-run and see what I can figure out....

tserong:
It's the `partprobe`:

```
# cat /proc/partitions
major minor  #blocks  name

   7        0     581632 loop0
 253        0  125829120 vda
 253       16 4294967296 vdb
 253       17       1024 vdb1
 253       18      51200 vdb2
 253       19    8388608 vdb3
 253       20   15728640 vdb4
 253       21 1287651328 vdb5
 253       22 2983144448 vdb6
 253       32  209715200 vdc
 253       33       1024 vdc1
 253       34      65536 vdc2
 253       35    4194304 vdc3
 253       36    8388608 vdc4
 253       37  197063680 vdc5
 253       48   20971520 vdd
  11        0    5863104 sr0

# lsblk -d -n -J -o NAME,TYPE | jq -r '.blockdevices[] | select(.type == "disk") | .name'
vda
vdb
vdc
vdd

# sgdisk -Z /dev/vda
Creating new GPT entries in memory.
GPT data structures destroyed! You may now partition the disk using fdisk or
other utilities.
# sgdisk -Z /dev/vdb
GPT data structures destroyed! You may now partition the disk using fdisk or
other utilities.
# sgdisk -Z /dev/vdc
GPT data structures destroyed! You may now partition the disk using fdisk or
other utilities.
# sgdisk -Z /dev/vdd
Creating new GPT entries in memory.
GPT data structures destroyed! You may now partition the disk using fdisk or
other utilities.

# partprobe -s
/dev/vdd: msdos partitions
/dev/vdb: msdos partitions
Error: Can't have a partition outside the disk!
/dev/vdc: msdos partitions
/dev/vda: msdos partitions

# echo $?
1
```

Vicente-Cheng:
Hi @tserong,
Could you check the vdb partitions?
It seems to be the root cause of the issue.

tserong:
In this case `/dev/vdb` is a 4TiB disk, and I don't know why it purports to have msdos partitions what with having just been zapped... But that'll be where the error is coming from.

tserong:
```
# fdisk -l /dev/vdb
Disk /dev/vdb: 4 TiB, 4398046511104 bytes, 8589934592 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes

# parted /dev/vdb print
Error: /dev/vdb: unrecognised disk label
Model: Virtio Block Device (virtblk)                                      
Disk /dev/vdb: 4398GB
Sector size (logical/physical): 512B/512B
Partition Table: unknown
Disk Flags: 
```

There's nothing there. What am I missing?

Vicente-Cheng:
> ```
> # fdisk -l /dev/vdb
> Disk /dev/vdb: 4 TiB, 4398046511104 bytes, 8589934592 sectors
> Units: sectors of 1 * 512 = 512 bytes
> Sector size (logical/physical): 512 bytes / 512 bytes
> I/O size (minimum/optimal): 512 bytes / 512 bytes
> 
> # parted /dev/vdb print
> Error: /dev/vdb: unrecognised disk label
> Model: Virtio Block Device (virtblk)                                      
> Disk /dev/vdb: 4398GB
> Sector size (logical/physical): 512B/512B
> Partition Table: unknown
> Disk Flags: 
> 
> # cat /proc/partitions 
> major minor  #blocks  name
> 
>    7        0     581632 loop0
>  253        0  125829120 vda
>  253       16 4294967296 vdb
>  253       32  209715200 vdc
>  253       48   20971520 vdd
>   11        0    5863104 sr0
> ```
> 
> There's nothing there. What am I missing?

How about vdc?

tserong:
vdc is the same (no partitions). I did a little more digging, turns out it's picking up the CD-ROM:

```
# for dev in vda vdb vdc vdd sr0 ; do echo $dev ; partprobe -s /dev/$dev ; done
vda
/dev/vda: msdos partitions
vdb
/dev/vdb: msdos partitions
vdc
/dev/vdc: msdos partitions
vdd
/dev/vdd: msdos partitions
sr0
Error: Can't have a partition outside the disk!
```

So, I've got a couple of ideas:

1. Move the `partprobe` inside the above loop (so we call `partprobe -s /dev/$disk` for each disk wiped, or,
2. Keep the existing partprobe but ignore the return value (so change it to `partprobe -s || :`)

Vicente-Cheng:
> vdc is the same (no partitions). I did a little more digging, turns out the partprobe of all devices is picking up the CD-ROM:
> 
> ```
> # for dev in vda vdb vdc vdd sr0 ; do echo $dev ; partprobe -s /dev/$dev ; done
> vda
> /dev/vda: msdos partitions
> vdb
> /dev/vdb: msdos partitions
> vdc
> /dev/vdc: msdos partitions
> vdd
> /dev/vdd: msdos partitions
> sr0
> Error: Can't have a partition outside the disk!
> ```
> 
> So, I've got a couple of ideas:
> 
> 1. Move the `partprobe` inside the above loop (so we only call `partprobe -s /dev/$disk` for each disk wiped), or,
> 2. Keep the existing partprobe but ignore the return value (so change it to `partprobe -s || :`)

nice catch, sr0 can not show the partition table.

If you just run `partprobe`, does it still hit error?

tserong:
> If you just run `partprobe` (w/o display partition), does it still hit error?

Yup:

```
# partprobe
Error: Can't have a partition outside the disk!
# echo $?
1
```

tserong:
One other thought - as I mentioned in https://github.com/harvester/harvester/issues/4527#issuecomment-1948077812:

> Thinking about this further though, some care must be taken before wiping everything that looks like a disk that's attached to a given host. Imagine you had a system that could see a bunch of random disks on a SAN. Wiping all of these might not be a good idea.

I still think this PR is fine, but when we document the `harvester.install.wipe_disks=true` option, we're going to need to say something like "DO NOT USE THIS UNLESS YOU REALLY REALLY MEAN IT"

tserong (approved):
I've just re-tested with the `partprobe -s /dev/$disk` change - works like a charm.

bk201 (approved):

bk201:
@mergifyio backport v1.2

PR #667, opened 2024-03-05
suffix os arch to harvester-images-list.txt and image-lists.tar.gz

ibrokethecloud:
**Problem:**

Currently CI for arm64 and amd64 images overwrites the harvester-images-list.text and image-lists.tar.gz file.

**Solution:**

The PR adds an arch suffix to both these files.

**Related Issue:**
https://github.com/harvester/harvester/issues/371
**Test plan:**


bk201 (approved):

Vicente-Cheng (approved):
lgtm!

PR #670, opened 2024-03-07
config/templates: update nvidia-driver-toolkit tag

Vicente-Cheng:
**Problem:**
we should use the same tag as the baseos image

**Solution:**
Use the same tag as the baseos image

**Related Issue:**
None

**Test plan:**
None


bk201 (approved):

ibrokethecloud (approved):
lgtm. thanks.

PR #671, opened 2024-03-08
generate squashfs for arm builds

ibrokethecloud:
**Problem:**

currently arm builds are not generating squashfs artifacts.
This is caused due to incorrect gating in the original arm support PR which only generates squashfs for amd64 archs.

**Solution:**

**Related Issue:**

**Test plan:**


bk201 (approved):

Vicente-Cheng (approved):
lgtm, thanks for fixing that.

Vicente-Cheng:
@mergifyio backport v1.3

PR #675, opened 2024-03-11
Issue 5178

mingshuoqiu:
**Problem:**
Please refer [harvester/harvester#5178](https://github.com/harvester/harvester/issues/5178). The additional mgmt-br.VLAN_ID interface would cause confusion and not helpful when debugging packet flow on network problem. 

**Solution:**
Do not create the additional bridge interface while applying network settings

**Test plan:**
1. install harvester
2. input vlan id for the management interface in network settings
3. Verify correct IP address is assigned for the specified VLAN
4. Press Esc to back to step 2 and input different VLAN ID


pkg/config/cos.go
```diff
@@ -606,7 +594,7 @@ func updateBridge(stage *yipSchema.Stage, name string, mgmtNetwork *Network) err
 		Group:       0,
 	})
 
-	preUpScript, err := render("wicked-setup-bridge.sh", MgmtBondInterfaceName)
+	preUpScript, err := render("wicked-setup-bridge.sh", mgmtNetwork.VlanID)
```
w13915984028, line 597: If user does not input mgmt vlan id, namely untagged network, is mgmtNetwork.VlanID == 0 ? what will hapen? thanks.

mingshuoqiu: The `BRIDGE_PRIORITY=0` will show in `ifcfg-mgmt-br` file which will not be effective. Since it's 0, which is not a valid vlan id, the `bridge add vid` command will not be in the setup_bridge.sh file. So the `pre-up` script will just set back to the default vlan setting, 1 for PVID, and 2 - 4094 for vid.

w13915984028 (approved):
LGTM, thanks.

pkg/config/templates/wicked-setup-bridge.sh
```diff
@@ -7,11 +7,14 @@ case $ACTION in
 	pre-up)
 		# enable vlan-aware
 		ip link set $INTERFACE type bridge vlan_filtering 1
-		;;
-
-	post-up)
+		# reset to default vlan state
 		# accept all vlan, PVID=1 by default
 		bridge vlan add vid 2-4094 dev $INTERFACE self
```
starbops, line 12: A general question: If we're going to drop the VLAN sub-interface for the node's management interface, can we get rid of this all-range-vlan-addition command?

```
bridge vlan add vid 2-4094 dev $INTERFACE self
```

pkg/config/templates/wicked-setup-bridge.sh
```diff
@@ -7,11 +7,14 @@ case $ACTION in
...
-		bridge vlan add vid 2-4094 dev {{ . }}
+		bridge vlan add vid 1 dev $INTERFACE pvid untagged self
```
starbops, line 13: I saw the following runtime configuration when I select 100 as the management interface's VLAN ID:

```
port              vlan-id  
mgmt-br           1 Egress Untagged
...
                  100 PVID Egress Untagged
...
```

Do we need `1 Egress Untagged`?

starbops (commented):
The functionality looks good, but I have some questions. Please see the comments below. Thanks.

pkg/config/templates/wicked-setup-bridge.sh
mingshuoqiu: On bridges, VID 0x001 (the default VLAN ID) is often reserved for a management VLAN. Cisco uses VLAN 1 as a default VLAN, and management protocols (STP, CDP, DTP, etc) are sent on VLAN 1. It's just a conventional usage.

pkg/config/templates/wicked-setup-bridge.sh
starbops: Thanks for providing the context. Now I know it is for the declarative outcome.

I was thinking if we could change this line into the exact opposite:

```
bridge vlan del vid 2-4094 dev $INTERFACE self
```

Will that achieve the same goal (resetting VLANs for the `mgmt-br` interface) and eliminate those unnecessary VIDs at the same time?

pkg/config/templates/wicked-setup-bridge.sh
starbops: I'm a bit curious. If there are indeed **VLAN ID 1** packets, e.g. CDP, entering the `mgmt-br` interface, which is currently configured as `1 Egress Untagged` + `100 PVID Egress Untagged`, won't they be tagged with **VLAN ID 100**? Is that a desired outcome? In other words, such a configuration will make both **VLAN ID 1 and 100** packets leave the `mgmt-br` interface untagged.

IIRC, those management protocols, like CDP, LLDP, etc., work on the native VLAN if the port is configured as a trunk port (in Cisco's terms). But it seems we're not configuring the `mgmt-br` interface as a trunk port here. Please correct me if I misunderstood the design. Thank you.

pkg/config/templates/wicked-setup-bridge.sh
mingshuoqiu: Yup. It means the same thing. Actually enable vlan_filtering can allow all VLAN packet to pass through.

pkg/config/templates/wicked-setup-bridge.sh
mingshuoqiu: Unless the packet is `xmit` or `generated` by the `mgmt-br` interface, the packet won't be tagged as VLAN100. So the CDP which pass through the `mgmt-br` will be untagged when the either the Host recv it from the `mgmt-br` interface or forward it to other `interface` (not port). 
If we enable `mgmt-br` with `vlan_filtering`. It's actually a trunk. If a port configured like our `mgmt-bo` to accept vid `2-4094`. The `mgmt-br` is the trunk port here.

starbops: I did some experiments regarding the dangling `1 Egress Untagged` VLAN filter entry. Given the following VLAN filter list:

```
port              vlan-id  
mgmt-br           1 Egress Untagged
                  100 PVID Egress Untagged
...
veth46b47c78      1 PVID Egress Untagged
```

The `veth46b47c78` port was added when I created a VM attached to a VM Network (`10.172.0.0/24`) with VLAN ID `1`. When the VM broadcasts ARP requests, I can observe the broadcast packets on the `mgmt-br` interface by `tcpdump`ing on it. That means the host can receive ARP broadcasts from the VM (though it will not reply), which is not our intention. I put the management interface in a different VLAN than the VM because I wanted to segregate the traffic.

```
node-0:~ # ip -d l show mgmt-br
4: mgmt-br: <BROADCAST,MULTICAST,PROMISC,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DEFAULT group default qlen 1000
    link/ether 52:54:01:45:00:01 brd ff:ff:ff:ff:ff:ff promiscuity 1 minmtu 68 maxmtu 65535 
    bridge forward_delay 0 hello_time 200 max_age 2000 ageing_time 30000 stp_state 0 priority 100 vlan_filtering 1 vlan_protocol 802.1Q bridge_id 0064.52:54:1:45:0:1 designated_root 0064.52:54:1:45:0:1 root_port 0 root_path_cost 0 topology_change 0 topology_change_detected 0 hello_timer    0.00 tcn_timer    0.00 topology_change_timer    0.00 gc_timer  140.24 vlan_default_pvid 1 vlan_stats_enabled 0 vlan_stats_per_port 0 group_fwd_mask 0 group_address 01:80:c2:00:00:00 mcast_snooping 1 mcast_router 1 mcast_query_use_ifaddr 0 mcast_querier 0 mcast_hash_elasticity 16 mcast_hash_max 4096 mcast_last_member_count 2 mcast_startup_query_count 2 mcast_last_member_interval 100 mcast_membership_interval 26000 mcast_querier_interval 25500 mcast_query_interval 12500 mcast_query_response_interval 1000 mcast_startup_query_interval 3124 mcast_stats_enabled 0 mcast_igmp_version 2 mcast_mld_version 1 nf_call_iptables 0 nf_call_ip6tables 0 nf_call_arptables 0 addrgenmode eui64 numtxqueues 1 numrxqueues 1 gso_max_size 65536 gso_max_segs 65535 
node-0:~ # ip link set mgmt-br promisc off
node-0:~ # ip -d l show mgmt-br
4: mgmt-br: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DEFAULT group default qlen 1000
    link/ether 52:54:01:45:00:01 brd ff:ff:ff:ff:ff:ff promiscuity 0 minmtu 68 maxmtu 65535 
    bridge forward_delay 0 hello_time 200 max_age 2000 ageing_time 30000 stp_state 0 priority 100 vlan_filtering 1 vlan_protocol 802.1Q bridge_id 0064.52:54:1:45:0:1 designated_root 0064.52:54:1:45:0:1 root_port 0 root_path_cost 0 topology_change 0 topology_change_detected 0 hello_timer    0.00 tcn_timer    0.00 topology_change_timer    0.00 gc_timer   83.56 vlan_default_pvid 1 vlan_stats_enabled 0 vlan_stats_per_port 0 group_fwd_mask 0 group_address 01:80:c2:00:00:00 mcast_snooping 1 mcast_router 1 mcast_query_use_ifaddr 0 mcast_querier 0 mcast_hash_elasticity 16 mcast_hash_max 4096 mcast_last_member_count 2 mcast_startup_query_count 2 mcast_last_member_interval 100 mcast_membership_interval 26000 mcast_querier_interval 25500 mcast_query_interval 12500 mcast_query_response_interval 1000 mcast_startup_query_interval 3124 mcast_stats_enabled 0 mcast_igmp_version 2 mcast_mld_version 1 nf_call_iptables 0 nf_call_ip6tables 0 nf_call_arptables 0 addrgenmode eui64 numtxqueues 1 numrxqueues 1 gso_max_size 65536 gso_max_segs 65535

node-0:~ # tcpdump -ennvi mgmt-br arp
...
tcpdump: listening on mgmt-br, link-type EN10MB (Ethernet), snapshot length 262144 bytes
09:42:12.021342 52:54:01:45:00:01 > ff:ff:ff:ff:ff:ff, ethertype ARP (0x0806), length 42: Ethernet (len 6), IPv4 (len 4), Request who-has 192.168.100.240 (ff:ff:ff:ff:ff:ff) tell 192.168.100.240, length 28
09:42:15.069452 52:54:01:45:00:01 > ff:ff:ff:ff:ff:ff, ethertype ARP (0x0806), length 42: Ethernet (len 6), IPv4 (len 4), Reply 192.168.100.240 is-at 52:54:01:45:00:01, length 28
09:42:17.681715 02:dc:70:54:4b:b8 > ff:ff:ff:ff:ff:ff, ethertype ARP (0x0806), length 42: Ethernet (len 6), IPv4 (len 4), Request who-has 10.172.0.1 tell 10.172.0.13, length 28
09:42:17.681913 a2:80:c4:44:da:66 > 02:dc:70:54:4b:b8, ethertype ARP (0x0806), length 42: Ethernet (len 6), IPv4 (len 4), Reply 10.172.0.1 is-at a2:80:c4:44:da:66, length 28
...
```

This is not a desired behavior since the management interface is bound to VLAN `100`, and we should consider removing `1 Egress Untagged` entirely. So ideally, we will have only 1 VLAN pass through the `mgmt-br` port, either `1 PVID Egress Untagged` by default or `X PVID Egress Untagged` if the user specifies other VLAN ID for the management interface. There won't be packets more than one VLAN leaving the port untagged.

mingshuoqiu:
```
That means the host can receive ARP broadcasts from the VM (though it will not reply), which is not our intention. I put the management interface in a different VLAN than the VM because I wanted to segregate the traffic.
```
What's the intention? The broadcast packets toward non-PVID VLAN should not be observed?
The `fdb` of the software bridge can't achieve the result...It only determines the target by dst mac address(VxLAN is another story). The broadcast packet would be flooded, the target port can only decide to either drop or receive it according to the VLAN tag.  

In my environment, I have 3 VID configured. The `tcpdump` on my computer which plugged on the VLAN 200 port of my management swich can still sniffer the broadcast packet from non-VLAN.

We need an example of the unicast (multicast) packet which could confuse the `fdb`. ex. 2 mac entries but on different vlans.

tjjh89017: > In my environment, I have 3 VID configured. The `tcpdump` on my computer which plugged on the VLAN 200 port of my management swich can still sniffer the broadcast packet from non-VLAN.

Could you provide your switch config for that switch port?

mingshuoqiu: I think the point here is `can we really have the packet which is tagged with vlan id 1?` VLAN ID 1 in most cases is only tagged internally by switch. It's mostly treated as untagged when enter/exit bridge. 
We need the example to show how the `tagged VLAN 1` packet could be generated and how it cause impact.

tjjh89017: FYI, If you used cisco, they could send tagged VLAN 1 out and deny all untagged packet.

tjjh89017: image
If you are talking about the term here, it means you cannot remove the VLAN1 from the VLAN database.
But it didn't mean you cannot remove native vlan 1 from a trunk port.
You could still deny any VLAN1 traffic to go out to this port.

Another thing, in CISCO switch, they didn't always send the Control-plane protocol through the VLAN1, just because VLAN1 is default, and not been disabled.
If you disable VLAN1 interface, CISCO will choose the lowest VLAN id to do the control plane stuffs, and it will depend on the native or trunk setting of the ports.

https://www.cisco.com/c/en/us/support/docs/network-management/discovery-protocol-cdp/118736-technote-cdp-00.html
image

tjjh89017: > In our cases, if I create a VM network as either UntaggedNetwork or L2VlanNetwork VLAN ID 1, the created veth are the same 1 PVID Egress Untagged. But I can't observe any packet tagged with VLAN 1 from these veth. So I don't know how the br_forward treat these untagged packet to the ports with the same VLAN.


Bridge will not push the VLAN header to the original packet.
It will only marks the `vlan_tci` and `vlan_proto` in the `skb`.
Refer to [this function](https://elixir.bootlin.com/linux/latest/source/net/bridge/br_vlan.c#L534)
So the `skb` packet part will not have any vlan header on it.

If bridge should push the tag for the egress packet, it will push it in `__dev_queue_xmit`.
Refer [__dev_queue_xmit](https://elixir.bootlin.com/linux/latest/C/ident/__dev_queue_xmit)
...omit...
Refer [__vlan_insert_inner_tag](https://elixir.bootlin.com/linux/latest/C/ident/__vlan_insert_inner_tag)

starbops (approved):
LGTM! Thank you for the enhancement.

PS. I can still sniff tagged frames originating from VMs in networks other than VLAN 100 (assume that the host interface is configured to VLAN 100 during installation) on the `mgmt-br` interface, but that's because we turned on the promiscuous mode on the `mgmt-br` interface. I'm unsure what the impact would be if we turned that off. Anyway, we can investigate/discuss this topic in a new thread.

PR #679, opened 2024-03-12
added additional check for first console when booting on arm hosts

ibrokethecloud:
**Problem:**

For iso based installs, we added logic to skip ttyAMA0 as the serial console runs into resolution issues, and this causes the installer to crash. 

Automatic install fails on arm hosts since the installer checks first console which is generally always returned as ttyAMA0. As a result of which automatic install is never triggered.

**Solution:**

The PR attempts to fix this by skipping ttyAMA0 if
* there are multiple consoles specified
* ttyAMA0 is the first console
* underlying host is an arm64 machine
* 
**Related Issue:**

**Test plan:**

### DIFF
--- a/pkg/console/tty.go
+++ b/pkg/console/tty.go
@@ -3,6 +3,7 @@ package console
 import (
 	"io/ioutil"
 	"os"
+	goruntime "runtime"
 	"strings"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


bk201 (approved):
Thanks

bk201:
@mergifyio backport v1.3

Vicente-Cheng (approved):
lgtm!

PR #682, opened 2024-03-25
Make image list packing architecture-awared

brandboat:
**Problem:**
The script archive-image-lists.sh lacks architecture awareness and could potentially lead to failures within our CI pipeline, particularly if newer versions (> 1.3.0) have been released.

**Solution:**
Compare the version in `upgrade-matrix.yaml` with v1.3.0, if version >= v1.3.0, which means that version already support different architecture images, and we need to download the image lists tarball from  `https://releases.rancher.com/harvester/{VERSION}/image-lists-{ARCH}.tar.gz`.

**Related Issue:**
https://github.com/harvester/harvester/issues/5313

**Test plan:**
1. Download harvester repo
2. replace [this line](https://github.com/harvester/harvester/blob/master/scripts/build-iso#L12) with `git clone --branch HARV-5313 --single-branch --depth 1 https://github.com/brandboat/harvester-installer.git ../harvester-installer`
3. add below yaml code snippet to [upgrade-matrix.yaml](https://github.com/harvester/harvester/blob/master/package/upgrade-matrix.yaml)
    ```yaml
    - name: v1.3.0
      minUpgradableVersion: v1.2.1
    ```
4. execute `make build-iso` without error

As you can see the console output shows that the script download v1.3.0 image list tarball with a specific architecture amd64, while others still use generic URL without architecture specification.
image

### DIFF
--- a/scripts/archive-images-lists.sh
+++ b/scripts/archive-images-lists.sh
@@ -41,7 +41,7 @@ for prev_ver in $(echo "$previous_versions"); do
 
   image_lists_url=https://releases.rancher.com/harvester/"$prev_ver"/image-lists.tar.gz
   # arch-aware image is available after v1.3.0
-  if [ "$(version_compare "$prev_ver" "v1.3.0")"  -eq 1 ]; then
+  if [ "$(version_compare "$prev_ver" "v1.3.0")" -eq 1 ]; then
     image_lists_url=https://releases.rancher.com/harvester/"$prev_ver"/image-lists-"$ARCH".tar.gz
   fi
 
--- a/scripts/archive-images-lists.sh
+++ b/scripts/archive-images-lists.sh
@@ -13,9 +13,9 @@ version_compare() {
     local v1=$1
     local v2=$2
     if [[ "$(printf "%s\n" "$v1" "$v2" | sort -V | head -n1)" == "$v2" ]]; then
-        echo "1"  # v1 is greater than or equal to v2
+        echo 1  # v1 is greater than or equal to v2
     else
-        echo "0"  # v1 is less than v2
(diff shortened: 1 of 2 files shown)


bk201 (approved):
lgtm, thanks!

bk201:
I created PRs to sync the matrix in the master and v1.3 branch:
- https://github.com/harvester/harvester/pull/5471
- https://github.com/harvester/harvester/pull/5472

starbops (approved):
LGTM, just one nit. Please see the comment below.

scripts/archive-images-lists.sh
```diff
@@ -28,7 +39,14 @@ for prev_ver in $(echo "$previous_versions"); do
 
   mkdir "$WORKING_DIR"/"$prev_ver"
 
-  curl -fL https://releases.rancher.com/harvester/"$prev_ver"/image-lists.tar.gz -o "$WORKING_DIR"/image-lists.tar.gz || ret=$?
+  image_lists_url=https://releases.rancher.com/harvester/"$prev_ver"/image-lists.tar.gz
+  # arch-aware image is available after v1.3.0
+  if [ "$(version_compare "$prev_ver" "v1.3.0")"  -eq "1" ]; then
```
starbops, line 44: Though it works, we usually use `-eq` to compare integers. It's better to follow the rules. Thank you.

starbops (approved):

scripts/archive-images-lists.sh
brandboat: oops, yes it should be integer, my bad :(

brandboat:
@mergify backport v1.3

PR #686, opened 2024-03-27
bump rke2 to v1.27.13+rke2r1 and rancher to v2.8.3

FrankYang0529:
**Problem:**

**Solution:**

**Related Issue:**
https://github.com/harvester/harvester/issues/4657, https://github.com/harvester/harvester/issues/5792

**Test plan:**
Create a cluster and operate VM without error.

PR #688, opened 2024-03-28
fix: CATTLE_FLEET_MIN_VERSION or CATTLE_RANCHER_WEBHOOK_MIN_VERSION may be empty

FrankYang0529:
**Problem:**
In some Rancher version like `v2.8.3-rc6`, these environment variables may be empty, so `min_version` in `update_chart_app_versions` will get input for `output_file`.

**Solution:**
Make sure `min_version` can get value from `CATTLE_FLEET_MIN_VERSION` or `CATTLE_RANCHER_WEBHOOK_MIN_VERSION`.

**Related Issue:**
https://github.com/harvester/harvester/issues/4657

**Test plan:**
CI should pass without error.

### DIFF
--- a/scripts/collect-deps.sh
+++ b/scripts/collect-deps.sh
@@ -72,9 +72,9 @@ update_rancher_deps()
   docker rm $(<$cid_file)
 
   # Get the latest version >= min_version and update it to yaml file
-  update_chart_app_versions $repo_index fleet $CATTLE_FLEET_MIN_VERSION $output_file
-  update_chart_app_versions $repo_index fleet-crd $CATTLE_FLEET_MIN_VERSION $output_file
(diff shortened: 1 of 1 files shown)


brandboat (approved):
LGTM ! :+1:

FrankYang0529:
Close this because we still need to get `CATTLE_FLEET_MIN_VERSION` and `CATTLE_RANCHER_WEBHOOK_MIN_VERSION`. These values are changed to [build.yaml](https://github.com/rancher/rancher/blob/release/v2.9/build.yaml). We fix it in another PR https://github.com/harvester/harvester-installer/pull/686/commits/9e08569d437991002ad9cbcc5dba97834208c543.

PR #690, opened 2024-03-28
Fix if condition in archive-images-lists.sh

brandboat:
related to https://github.com/harvester/harvester/issues/5313
this is a followup in https://github.com/harvester/harvester-installer/pull/682


starbops (approved):
LGTM, thank you!

bk201 (approved):

PR #697, opened 2024-04-10
ci: migrate .drone.yml to github build action

tserong:
**Problem:**
Need to migrate from Drone to Github Actions

**Solution:**
This gives us amd64 and arm64 ISO builds on every pull request and also on branch pushes of master and v*.  ISO images are saved with the workflow as artifact ZIP files in case they are useful for manual testing.  Unlike with the Drone build, we are _not_ publishing images to releases.rancher.com to avoid potential collisions with images published from the main harvester repo.

This PR also makes a small change to `scripts/lib/image` to make it easier to debug image pull failures during build.

**Related Issue:**
https://github.com/harvester/harvester/issues/5327


FrankYang0529 (approved):
LGTM. Thank you.

bk201 (approved):
Thanks!

tserong:
@mergifyio backport v1.2 v1.3

PR #708, opened 2024-04-18
change how addons are packaged

ibrokethecloud:
**Problem:**

PR changes how addons are packaged in the installer.

They have now been moved to `harvester/addons` and a cli is used to generate the templates.

**Solution:**

**Related Issue:**

**Test plan:**


pkg/config/cos.go
```diff
@@ -732,6 +731,23 @@ func genBootstrapResources(config *HarvesterConfig) (map[string]string, error) {
...
 
+	// render addons
+	for _, templateName := range []string{
+		"harvester-seeder.yaml",
+		"nvidia-driver-toolkit.yaml",
+		"pcidevices-controller.yaml",
+		"rancher-logging.yaml",
+		"rancher-monitoring.yaml",
+		"vm-import-controller.yaml",
+	} {
```
tserong, line 742: Just a thought as I'm looking through this - is it possible to somehow grab all the template names from the addons repo, so if a new addon is created in future, we don't need to explicitly specify its name here?

ibrokethecloud: i had made some changes to the addons side, so we are back to using just one addons file.

scripts/build
```diff
@@ -27,6 +27,12 @@ else
     git config --global --add safe.directory /go/src/github.com/harvester/harvester
 fi
 
+addons_path=../addons
+if [ ! -d ${addons_path} ];then
+    echo "No existed harvester source. Pulling..."
```
tserong, line 32: Probably should mention this is "addons":

```suggestion
    echo "No existing harvester addons source. Pulling..."
```

scripts/build-bundle
```diff
@@ -9,6 +9,12 @@ PACKAGE_HARVESTER_REPO_DIR="${TOP_DIR}/package/harvester-repo"
 
 cd ${TOP_DIR}
 
+addons_path=../addons
+if [ ! -d ${addons_path} ];then
+    echo "No existed harvester source. Pulling..."
+    git clone --branch main --single-branch --depth 1 https://github.com/harvester/addons.git ../addons
+fi
+
```
tserong, line 17: Do we really need this, given it's already done in `scripts/build`? (I assume the answer is "yes", given we already do the same with the harvester source, but I've just noticed the duplication, so I had to ask...)

scripts/build-bundle
```diff
@@ -17,20 +23,13 @@ source ${SCRIPTS_DIR}/lib/image
 source ${SCRIPTS_DIR}/hack/patch-rancher-monitoring
 source ${SCRIPTS_DIR}/hack/patch-rancher-logging
 source ${SCRIPTS_DIR}/hack/patch-rancher-monitoring-crd
-source ${SCRIPTS_DIR}/hack/patch-harvester-chart
```
tserong, line 20: Should the above line be removed here? It looks arm-related to me.

tserong (commented):
LGTM, just a couple more questions :-)

scripts/build-bundle
ibrokethecloud: I believe both are around in case someone directly calls `make build` and `make build-bundle` targets manually

scripts/build-bundle
ibrokethecloud: good spot. i think i deleted it. fixed it up now.

tserong (approved):

starbops (approved):
LGTM, thanks!

PR #717, opened 2024-04-24
Add ability to set hostname via DHCPv4 client

tserong:
**Problem:**
We need to provide the option of allowing system hostnames to be set via DHCP.

**Solution:**
This adds a boolean config option, os.dhclient_set_hostname, which maps to DHCLIENT_SET_HOSTNAME in /etc/sysconfig/network/dhcp.

When running the installer interactively, there's a new "Set via DHCP" drop down field on the hostname page.  The default is "no".

image

If the user selects "yes", this will later be shown on the confirmation screen as "hostname: XXXX (will be set via DHCP)", to indicate that the provided hostname will be overridden by DHCP:

image

For automated installs, either set `os.dhclient_set_hostname: true` in the config yaml, or pass `harvester.os.dhclient_set_hostname=true` as a kernel command line argument.

In order for this setting to take effect, your DHCP server needs to be configured to send the host name option to the client.  For more information see `man dhcpd.conf` (look for "option host-name" and "use-host-decl-names on").

Note that even if os.dhclient_set_hostname is set to true, we still require the user to provide a hostname during installation, because that will be written to /etc/hostname and be used by the system briefly early in the boot process, before DHCP does its thing.

**Related Issue:**
https://github.com/harvester/harvester/issues/1444

**Test plan:**
- Configure a DHCP server to set the host name option (TBD: add more details here, possibly including tweaks to ipxe-examples)
- Boot the installer, but leave "Set via DHCP" set to "no"
- Verify that whatever hostname the DHCP server wants to set, doesn't get set.
- Repeat the above, but set "Set via DHCP" to "yes"
- Verify that whatever hostname the DHCP server wants to set, does get set, both during installation (after network config is applied), and then subsequently once the system comes up for real.

### DIFF
--- a/pkg/console/install_panels.go
+++ b/pkg/console/install_panels.go
@@ -1176,14 +1176,25 @@ func showHostnamePage(c *Console) error {
 		return err
 	}
 
+	if err := setLocation(askDhclientSetHostnamePanel, 3); err != nil {
+		return err
+	}
+
 	if err := setLocation(hostnameValidatorPanel, 0); err != nil {
 		return err
 	}
-	return showNext(c, hostnamePanel)
+	return showNext(c, askDhclientSetHostnamePanel, hostnamePanel)
 }
 
 func addHostnamePanel(c *Console) error {
 	hostnameV, err := widgets.NewInput(c.Gui, hostnamePanel, hostNameLabel, false)
+	if err != nil {
+		return err
+	}
+
+	askDhclientSetHostnameV, err := widgets.NewDropDown(c.Gui, askDhclientSetHostnamePanel, "Set via DHCP", func() ([]widgets.Option, error) {
+		return []widgets.Option{{Value: "no", Text: "No"}, {Value: "yes", Text: "Yes"}}, nil
+	})
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


tserong:
> What will happen if the dhcp server doesn't set the hostname for the IP?
> Will it fallback to manual hostname setting and give a warning or simply bypass?

It just falls back to the manual hostname setting. AFAICT there's no warnings anywhere if the DHCP server doesn't send a hostname. That would be up to wickedd and/or wickedd-dhcp4 to do, but from looking at the wicked code and experimenting with extra debugging information turned on, wicked doesn't seem to really care whether or not the DHCP server sends a hostname or not. If the server does send a hostname and DHCLIENT_SET_HOSTNAME is set to "yes", it will be used, but it won't complain at all if either of those things aren't true.

pkg/console/network.go
```diff
@@ -16,14 +16,26 @@ import (
...
 		return nil, err
 	}
 
+	yesOrNo := "no"
+	if dhclientSetHostname {
+		yesOrNo = "yes"
+	}
+	output, err := exec.Command("sed", "-i",
+		fmt.Sprintf(`s/^DHCLIENT_SET_HOSTNAME=.*/DHCLIENT_SET_HOSTNAME="%s"/`, yesOrNo),
+		"/etc/sysconfig/network/dhcp").CombinedOutput()
```
bk201, line 33: The file control global settings. Should we set the value on the management interface only? Although we don't request IPs on VLAN bridges (those generated from harvester-network-controller), setting this per interface can lower the chance of interfering with other interfaces.

From the [man page](https://manpages.opensuse.org/Tumbleweed/wicked/ifcfg-dhcp.5.en.html) it seems we can override the setting in the `ifcfg-mgmt-br` file. And if we make the change, we can consider moving the configuration field `dhclientSetHostname` into `install.management_interface`.

pkg/console/install_panels.go
```diff
@@ -1188,6 +1192,13 @@ func addHostnamePanel(c *Console) error {
 		return err
 	}
 
+	askDhclientSetHostnameV, err := widgets.NewDropDown(c.Gui, askDhclientSetHostnamePanel, "Set via DHCP", func() ([]widgets.Option, error) {
+		return []widgets.Option{{Value: "no", Text: "No"}, {Value: "yes", Text: "Yes"}}, nil
```
bk201, line 1196: I'll also suggest adding a warning message if the user selects `yes` to get the hostname from the DHCP server. The idea is that the hostname should be **always** assigned persistently by the DHCP server. hostname changes can lead to unexpected consequences. etcd, certificates, and many services rely on the hostname.

pkg/console/network.go
tserong: Yeah, I agree, setting this on the management interface only rather than globally makes good sense.

pkg/console/install_panels.go
tserong: Yep. If we do take this change (rather than the alternative in #726) adding a warning here seems like a very good idea.

tserong:
Closing in favour of #726

PR #725, opened 2024-05-02
enhancement: check default route during installation

mingshuoqiu:
**Problem:**
RKE2 requires a default route to start, some installations are stuck with no obvious message if there is no default route on the system.

**Solution:**
Check the existence of the default route after applying network settings and stop the installation if no default route found. The check will be applied in both ISO and iPXE installation.

**Related Issue:**
https://github.com/harvester/harvester/issues/5675

**Test plan:**
1. Set up a DHCP server and enable the iPXE capability which runs w/o `option routers x.x.x.x`
2. Proceed the Harvester ISO installation with the management network in DHCP mode.
3. Make sure the installer prompt the `No default route found` message and stop the installation
4. Proceed the iPXE automatic installation with pre-configured Harvester configuration file.
5. Make sure the iPXE automatic installation stops and prompt the `No default route found` message.

### DIFF
--- a/pkg/console/install_panels.go
+++ b/pkg/console/install_panels.go
@@ -42,6 +42,7 @@ const (
 	ErrMsgVLANShouldBeANumberInRange string = "VLAN ID should be a number 1 ~ 4094."
 	ErrMsgMTUShouldBeANumber         string = "MTU should be a number."
 	NtpSettingName                   string = "ntp-servers"
+        ErrMsgNoDefaultRoute		 string = "No default route found. Please check the router setting on the DHCP server."
 )
 
 var (
@@ -1388,6 +1389,15 @@ func addNetworkPanel(c *Console) error {
 			mgmtNetwork.Gateway = ""
 			mgmtNetwork.MTU = 0
 		}
+
+		isDefaultRouteExist, err := checkDefaultRoute()
+		if err != nil {
+			return fmt.Sprintf("Failed to check default route: %s.", err.Error()), nil
+		}
+		if !isDefaultRouteExist {
+			return ErrMsgNoDefaultRoute, nil
+		}
+
 		return "", nil
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/console/install_panels.go
```diff
@@ -1388,6 +1388,11 @@ func addNetworkPanel(c *Console) error {
 			mgmtNetwork.Gateway = ""
 			mgmtNetwork.MTU = 0
 		}
+
+		if !isDefaultRouteExist() {
+			return fmt.Sprint("No default route found. Please check the router setting on the DHCP server"), nil
```
tserong, line 1393: This can just be `return "No default route found. Please check the router setting on the DHCP server", nil` (there's no need for the `fmt.Sprint`, because there's no other variables being interpolated into the string).

pkg/console/install_panels.go
```diff
@@ -2103,6 +2108,11 @@ func addInstallPanel(c *Console) error {
 				}
 			}
 
+			if !isDefaultRouteExist() {
+				printToPanel(c.Gui, fmt.Sprint("No default route found. Please check the router setting on the DHCP server"), installPanel)
```
tserong, line 2112: Same comment as above about not needing `fmt.Sprint`.

Also, I'd suggest adding `logrus.Error("No default route [...]")` so the same message goes to both the panel, and to `/var/log/console.log`

tserong (changes requested):
LGTM in general, just a couple of small suggestions.

tserong (approved):

pkg/console/network.go
```diff
@@ -16,6 +17,24 @@ import (
 	"github.com/harvester/harvester-installer/pkg/config"
 )
 
+func isDefaultRouteExist() bool {
+	routes, err := netlink.RouteList(nil, syscall.AF_INET)
+	if err != nil {
```
w13915984028, line 22: This is also a possible return path, the error message can only be checked on log instead of installation panel.

Please consider to return `bool, error` and capture the error on the caller to show it. thanks.

pkg/console/install_panels.go
```diff
@@ -2103,6 +2112,18 @@ func addInstallPanel(c *Console) error {
...
 
+			isDefaultRouteExist, err := checkDefaultRoute()
+			if err != nil {
+				logrus.Error(err)
+				printToPanel(c.Gui, "Failed to check default route.", installPanel)
+				return
+			}
+			if !isDefaultRouteExist {
+				logrus.Error("No default route found. Please check the router setting on the DHCP server.")
+				printToPanel(c.Gui, "No default route found. Please check the router setting on the DHCP server.", installPanel)
```
starbops, line 2123: There are three places that use this very same sentence. It would be nice if we made it a const.

starbops (approved):
LGTM, just one nit.

mingshuoqiu:
@w13915984028  @starbops revised per your suggestion. Thanks

w13915984028 (approved):
LGTM, thanks.

PR #726, opened 2024-05-08
Add ability to set hostname via DHCP during install

tserong:
**Problem:**
We need to provide the option of allowing system hostnames to be set via DHCP.

**Solution:**
When installing Harvester, if the management interface is configured via DHCP, we will default the system hostname to the client hostname specified by the DHCP server.
    
For interactive installation, the order of the "Configure hostname" and "Configure network" pages is swapped, so that "Configure network" comes first.  This allows us to set DHCLIENT_SET_HOSTNAME="yes" when applying the management NIC config.  If DHCP is used, and the DHCP server is configured to specify a hostname, the live environment will then automatically have its hostname set to whatever it got from the DHCP server.  Then, on the hostname page, we can default the hostname to that value, which the user is still free to change should they wish.
    
For automated/PXE installation, it works like this:
- If os.hostname is specified in the harvester config, we use that.
- If os.hostname is not specified, we use the DHCP provided hostname if set, otherwise we fall back to a randomly generated hostname.

(This is an alternative to https://github.com/harvester/harvester-installer/pull/717 - the advantage with this approach is that we set the hostname once, during installation, whereas #717 potentially allows the DHCP server to change the hostname of an installed system later, which would likely cause all sorts of trouble)

**Related Issue:**
https://github.com/harvester/harvester/issues/1444

**Test plan:**
- Configure a DHCP server, but don't have it set to provide client hostnames (i.e. _don't_ use `option host-name XXX` or `use-host-decl-names on` in `dhcpd.conf`)
- Boot the installer ISO and walk through the installation process, selecting DHCP for the management interface.
- Verify that the network page now comes before the hostname page, but that the hostname by default remains blank.
- Verify that installation completes successfully
- Do a PXE install, set to use DHCP, and verify that it completes as usual.
- Reconfigure the DHCP server to provide client hostnames.
- Boot the installer ISO and walk through the installation process, selecting DHCP for the management interface.
- Verify that when you get to the hostname page, the hostname defaults to whatever was specified in the DHCP server config for this host, but that you can also still change the hostname if you want.
- Verify that the install completes successfully
- Do a PXE install, but make sure `os.hostname` is _not_ set in the harvester config yaml.
- Verify that the installation completes successfully, and that the system picks up the DHCP-provided hostname.

Example dhcpd.conf snippet which should set the hostname to "harv-0-dhcp" for the MAC address 02:00:00:0D:62:E2:

```
host harvest_node_0 {
    hardware ethernet 02:00:00:0D:62:E2;
    fixed-address 192.168.0.30;
    option host-name "harv-0-dhcp";
}
```

### DIFF
--- a/pkg/console/constant.go
+++ b/pkg/console/constant.go
@@ -83,4 +83,6 @@ const (
 	persistentSizeNote     = "Note: persistent partition stores data like system package and container images, not the VM data. \nYou can specify a size like 200Gi or 153600Mi. \nLeave it blank to use the default value."
 
 	authorizedFile = "/home/rancher/.ssh/authorized_keys"
+
+	defaultHostname = "rancher"
 )
--- a/pkg/console/install_panels.go
+++ b/pkg/console/install_panels.go
@@ -506,7 +506,7 @@ func addDiskPanel(c *Console) error {
 		if installModeOnly {
 			return showNext(c, passwordConfirmPanel, passwordPanel)
 		}
-		return showHostnamePage(c)
+		return showNetworkPage(c)
 	}
 
 	diskConfirm := func(_ *gocui.Gui, _ *gocui.View) error {
@@ -840,7 +840,7 @@ func addAskCreatePanel(c *Console) error {
 				return showRolePage(c)
 			}
 			if alreadyInstalled {
-				return showHostnamePage(c)
+				return showNetworkPage(c)
 			}
 			return showDiskPage(c)
 		},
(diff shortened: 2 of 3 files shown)


ibrokethecloud (approved):
lgtm. thanks.

tserong:
re: those last two commits, printing the `yip` output means we get some sort of error message if we can't apply the network config, e.g.:

cant-apply-networks

...and re-ordering the config merge means we can correctly handle remote config that specifies MAC addresses but not interface names.

bk201 (approved):
LGTM!

PR #730, opened 2024-05-15
Use dmidecode to check available RAM in preflight checks

tserong:
**Problem:**
The current implementation of the preflight memory check uses `MemTotal` from `/proc/meminfo` to try to figure out how much RAM there is.  This number will always be a little bit low, which is confusing ("Why is it reporting only 60GiB when I know I have 64GiB?!?") and means we have to allow a certain amount of wiggle room when determining whether or not the system meets the minimum production requirements.

**Solution:**
This commit makes the following changes:

1. We use `dmidecode -t 19` to get the exact amount of memory installed in the system.  No weird low values, no need for wiggle room.
2. In case the above fails and we have to fall back to `/proc/meminfo`, increase the wiggle room from 5% to 10%.
3. For really low RAM scenarios (e.g. tiny test VMs with <1GiB RAM), report available memory in MiB rather than KiB (nobody thinks in KiB anymore, right?)

**Related Issue:**
https://github.com/harvester/harvester/issues/5599

**Test plan:**
- Boot a system with <64GiB RAM. Verify that the warning message gives the exact amount of RAM you actually have (e.g. 16GiB, 32GiB), vs. undercounting like it used to do (e.g. 7GiB on an 8GiB system).

### DIFF
--- a/pkg/preflight/checks.go
+++ b/pkg/preflight/checks.go
@@ -9,6 +9,8 @@ import (
 	"os/exec"
 	"strconv"
 	"strings"
+
+	"github.com/sirupsen/logrus"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/preflight/checks.go
```diff
@@ -62,51 +62,139 @@ func (c CPUCheck) Run() (msg string, err error) {
...
+		}
+
+		for _, line := range strings.Split(string(out), "\n") {
+			var rangeSize uint
+			var unit string
+			if n, _ := fmt.Sscanf(strings.TrimSpace(line), "Range Size: %d %s", &rangeSize, &unit); n == 2 {
+				if unit == "TB" || unit == "PB" || unit == "EB" || unit == "ZB" {
+					// If we've somehow got a Memory Array Mapped Address
+					// with one of these enormous units, let's just pretend
+					// we've got a terabyte of RAM and be done with it ;-)
```
Vicente-Cheng, line 129: Could we add some logs to record this situation?
I know that might not happen very soon, but we could add a log at this time.

Vicente-Cheng (approved):
LGTM!  just a nit.

Thanks for the enhancement!

bk201 (approved):

pkg/preflight/checks.go
tserong: Done

tserong:
@Mergifyio backport v1.3

PR #734, opened 2024-05-17
[backport 1.3] enhancement: check default route during installation (#725)

mingshuoqiu:
RKE2 requires a default route to work which is required during installation. The gateway setting will be checked in Static IP mode, but it could be absent in DHCP mode.

Check the existence of the default route after applying network settings and stop the installation if no default route found. The check will be applied in both ISO and iPXE installation.

Link: https://github.com/harvester/harvester/issues/5675

**Problem:**
Per https://github.com/harvester/harvester/issues/5675

**Solution:**
add default route check during installation

**Related Issue:**
https://github.com/harvester/harvester/issues/5675

**Test plan:**
As steps in https://github.com/harvester/harvester-installer/pull/725


bk201 (approved):
LGTM

starbops (approved):
LGTM, thanks.

PR #737, opened 2024-05-22
Use v1.3 branch in the harvester repo

bk201:
Fix the wrong version of harvester images/charts used while packaging v1.3 ISO.


FrankYang0529 (approved):
LGTM. Thank you.

Vicente-Cheng (approved):
nice catch!

PR #739, opened 2024-05-22
Change package repos in Vagrantfile

votdev:
**Problem:**
This is necessary because the Leap 15.4 package repo, which is mentioned in the README to install `delve` for interactive debugging, is not available anymore.

**Solution:**
Change the package repos.

As the change only affects the development environment of harvester-installer, it has no effect on the installer itself.

**Related Issue:**
n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ To remote debug the `harvester-installer` in your IDE, you need to
 install the `delve` package in your Vagrant box first.
 
 ```sh
- $ zypper addrepo https://download.opensuse.org/repositories/devel:languages:go/15.4/devel:languages:go.repo
+ $ zypper addrepo https://download.opensuse.org/repositories/devel:/languages:/go/SLE_15_SP4/devel:languages:go.repo
  $ zypper refresh
  $ zypper install delve
 ```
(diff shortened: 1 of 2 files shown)


tserong (commented):
I'm going to be difficult (sorry!) and suggest an alternative:

- Leave the base OS on 15.4
- Change the devel:langauges:go repo to https://download.opensuse.org/repositories/devel:/languages:/go/SLE_15_SP4/devel:languages:go.repo
- Change https://download.opensuse.org/repositories/home:/vcheng:/Packages/15.4/home:vcheng:Packages.repo to https://download.opensuse.org/repositories/isv:/Rancher:/Harvester:/ExtraPackages:/Dev/15.4/isv:Rancher:Harvester:ExtraPackages:Dev.repo

The rationale for the above is just that I think it makes sense to switch from Vicente's home repo to isv:Rancher:Harvester:ExtraPackages, but that the latter doesn't build for 15.5...  Although, we could always enable that - I have no particular actual objection to updating the base OS :-)

WDYT?

tserong (approved):

Vicente-Cheng (approved):
LGTM!  Thanks @votdev @tserong

PR #745, opened 2024-05-30
feat: default security policy for mgmt/worker/witness nodes

mingshuoqiu:
**Problem:**
Need to add a basic security to Harvester to prevent unnecessary ports been used. Will need a UI in the future to allow users to add ports manually

**Solution:**
Put particular ports in whitelist per different role to allow particular traffic to pass through.

**Related Issue:**
https://github.com/harvester/harvester/issues/5681

**Test plan:**
1. Setup a basic environment with 3 roles of node. Management, Worker and Witness
2. All functions in the Harvester. need to work as usual.
3. ping should not work from outside network.


mingshuoqiu:
@rrajendran17 Could you help review this? Thanks

PR #757, opened 2024-06-18
build: remove mkinitrd and fix docker create

Vicente-Cheng:
**Problem:**
There were some build issues when we bumped the sle-micro 5.5.

**Solution:**
Use the correct command to build iso successfully.

**Related Issue:**
https://github.com/harvester/harvester/issues/6027

**Test plan:**
TBD

### DIFF
--- a/scripts/package-harvester-os
+++ b/scripts/package-harvester-os
@@ -78,7 +78,8 @@ then
 fi
 
 INITRD=$(docker run --rm ${HARVESTER_OS_IMAGE} readlink /boot/initrd)
-docker create --cidfile=os-img-container ${HARVESTER_OS_IMAGE}
+# we need to add entrypoint or the docker create failed in newer version
+docker create --cidfile=os-img-container ${HARVESTER_OS_IMAGE} -- tail -f /dev/null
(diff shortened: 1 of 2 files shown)


scripts/package-harvester-os
```diff
@@ -78,7 +78,7 @@ then
 fi
 
 INITRD=$(docker run --rm ${HARVESTER_OS_IMAGE} readlink /boot/initrd)
-docker create --cidfile=os-img-container ${HARVESTER_OS_IMAGE}
+docker create --cidfile=os-img-container ${HARVESTER_OS_IMAGE} -- sleep 30
```
tserong, line 81: Why do we need this? Is this a requirement of a newer version of Docker? (Also, you could maybe try something like `tail -f /dev/null` which should keep it running forever until it's explicitly removed - I never really liked fixed sleeps :-) )

tserong (approved):
LGTM, just a question.

scripts/package-harvester-os
Vicente-Cheng: Thanks @tserong.
Yeah, the newer version Docker requires the entry point (means the command to execute).
Thanks for the hint. The `tail -f /dev/null` is the better way. Updated!

tserong (approved):

bk201 (approved):

PR #759, opened 2024-06-26
fix installer not initialising on arm hosts in equinix metal

ibrokethecloud:
**Problem:**

Currently harvester installer does not initialise on arm instances on equinix metal.

The original arm implementation skipped the serial console /dev/ttyAMA0 in favour of non serial terminals exposed such as /dev/tty1 exposed on the ampere hosts.

Arm hosts in equinix metal only expose /dev/ttyAMA0 via serial-over-ssh. Since there is no alternate terminal exposed the installer does not start.

**Solution:**

The PR introduces a minor change to setup-installer script to favour non /dev/ttyAMA0 terminals if one is present such as on Ampere hosts, and failover to /dev/ttyAMA0 if that is the only option.

**Related Issue:**

**Test plan:**

### DIFF
--- a/package/harvester-os/files/usr/bin/setup-installer.sh
+++ b/package/harvester-os/files/usr/bin/setup-installer.sh
@@ -44,12 +44,17 @@ echo "Remove the serial-getty service..."
 rm -rf "/etc/systemd/system/serial-getty*"
 
 # reverse the ttys to start from the last one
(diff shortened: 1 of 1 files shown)


tserong (approved):

bk201 (approved):

bk201:
@mergifyio backport v1.3

PR #775, opened 2024-07-23
Log validateConfig, PrepareWebooks and doInstall errors

tserong:
**Problem:**
Prior to this commit, errors returned from `validateConfig()`, `PrepareWebooks()` and `doInstall()`  would be printed to the console, but not logged.  This meant that if you weren't able to access the console for whatever reason, but _could_ access the log file, you would have no idea what went wrong if one of those function calls failed.

**Solution:**
Call `logrus.Error()` in addition to `printToPanel()`.

Bonus: call `logrus.Info("Installation will proceed (harvester.install.skipchecks = true)")` to make it explicit when preflight checks fail but are intentionally skipped.

**Related Issue:**
https://github.com/harvester/harvester/issues/6214

**Test plan:**
- Do a PXE install, but with an invalid config file (see reproducer in https://github.com/harvester/harvester/issues/6214)
- Confirm that an appropriate error message appears on the console _and_ in `/var/log/console.log`, rather than only appearing on the console.

For example, with this configuration:

```
install:
  mode: This ain't right
```

...we should see something like the following on the console:

image

...and something like this in the log:

image

### DIFF
--- a/pkg/console/install_panels.go
+++ b/pkg/console/install_panels.go
@@ -2132,6 +2132,7 @@ func addInstallPanel(c *Console) error {
 						for _, warning := range preflightWarnings {
 							logrus.Warning(warning)
 						}
+						logrus.Info("Installation will proceed (harvester.install.skipchecks = true)")
 					} else {
 						// Checks were not explicitly skipped, fail the install
 						// (this will happen when PXE booted if checks fail and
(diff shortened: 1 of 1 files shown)


bk201 (approved):
Thanks!

Vicente-Cheng (approved):
good enhancement!

tserong:
@mergifyio backport v1.4 v1.3

PR #766, opened 2024-07-12
bump rancher-monitoring to 103.1.1+up45.1.1 rancher-logging to 103.1.0+up4.4.0

w13915984028:
**Problem:**

Bump the rancher-monitoring chart to 103.1.1+up45.1.1 and fix CVEs.

Bump rancher-logging to 103.1.0+up4.4.0

**Solution:**

Bump rancher-monitoring charts
Bump rancher-logging charts
Bump shell image (0.1.26) to reduce CVE
Bump eventrouter image (not available yet)

Upstream has fixed the rancher-monitoring-crd  RBAC issue, the local patch is removed.

The image `docker.io/rancher/shell:v0.1.23` is used by Rancher. The `rancher-monitoring` and `rancher-logging`  is verified not using the v0.1.23 but v0.1.26.

Depends on PR

- [x] https://github.com/harvester/addons/pull/5

**Related Issue:**
https://github.com/harvester/harvester/issues/6166

**Test plan:**

(1) Install a new cluster
(2) Enable rancher-monitoring addon, it should work
(3) Enable rancher-logging addon, it should work

The upgrade processing will be in another PR

local test:
```
$ cat /oem/harvester.config
...
runtimeversion: v1.27.13+rke2r1
rancherversion: v2.8.3
harvesterchartversion: 0.0.0-master-8c709bac
monitoringchartversion: 103.1.1+up45.31.1
..
loggingchartversion: 103.1.0+up4.4.0
```

image

pods:
```
harv21:/home/rancher # kubectl get pods -n cattle-monitoring-system
NAME                                                     READY   STATUS      RESTARTS   AGE
alertmanager-rancher-monitoring-alertmanager-0           2/2     Running     0          32m
helm-install-rancher-monitoring-xzwq9                    0/1     Completed   0          32m
prometheus-rancher-monitoring-prometheus-0               3/3     Running     0          32m
rancher-monitoring-grafana-d6f466988-vcc6x               4/4     Running     0          32m
rancher-monitoring-kube-state-metrics-7889dccd84-x9fpj   1/1     Running     0          32m
rancher-monitoring-operator-7545bff858-js5nh             1/1     Running     0          32m
rancher-monitoring-prometheus-adapter-55dc9ccd5d-mj7db   1/1     Running     0          32m
rancher-monitoring-prometheus-node-exporter-7h6pg        1/1     Running     0          32m
harv21:/home/rancher # kubectl get pods -n cattle-logging-system
[...]

PR #780, opened 2024-07-25
bump monitoring charts shell version

mallardduck:
**Problem:**
I'm on the O&B team that maintains shell and just looking to bump shell for CVE's we've addressed.
Unfamiliar with harvester stuff and this repo didn't have issues, so just made the PR myself; so something may be wrong.

Let me know if it needs changes.

If I follow correctly harvester uses Rancher 2.8.x currently, so the newest `v0.1.X` of shell should be the target; currently: https://github.com/rancher/shell/releases/tag/v0.1.26

**Solution:**
Added new tag to image list, then bumped version in patches to use newer version.

**Related Issue:**

**Test plan:**

### DIFF
--- a/pkg/config/templates/patch/rancher-monitoring-crd/103.0.3+up45.31.1/patch-values-shell-version.diff
+++ b/pkg/config/templates/patch/rancher-monitoring-crd/103.0.3+up45.31.1/patch-values-shell-version.diff
@@ -5,7 +5,7 @@
  image:
    repository: rancher/shell
 -  tag: v0.1.18
-+  tag: v0.1.22
++  tag: v0.1.26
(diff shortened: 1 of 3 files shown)


mallardduck:
Optionally - if the harvester side supports it, this customization to use newer shell could extend to the embedded rancher too. Since rancher controls shell via a setting named `shell-image` an ENV var of `CATTLE_SHELL_IMAGE` could overwrite the default one that would be used.

w13915984028:
@mallardduck Thanks very much.

The harvester PR https://github.com/harvester/harvester-installer/pull/766 has included your changes to solve CVE issues, and with more.

PR #783, opened 2024-07-29
leverage generateTemplates capability of harvester/addons

ibrokethecloud:
**Problem:**

PR introduces minor changes to installer to leverage the new `generateTemplate` capability added to `harvester/addons`

Prior to this change the addon updates were being managed by updates to version_info file, and also updating the version details in the rancher-22-addons.yaml template file.

**Solution:**

This has been simplified to use the chart versions in version_info file and allow ability to generate template file dynamically. This ensures that chart and image version change is only ever kept in one single location.

addons PR: https://github.com/harvester/addons/pull/4

**Related Issue:**
**Test plan:**

### DIFF
--- a/scripts/build-bundle
+++ b/scripts/build-bundle
@@ -15,6 +15,11 @@ if [ ! -d ${addons_path} ];then
     git clone --branch main --single-branch --depth 1 https://github.com/harvester/addons.git ../addons
 fi
 
+# will generate template file in ${addons_path}, and this is used for subsequent checks
+cd ${addons_path}
+go run . -generateTemplates 
+cd ${TOP_DIR}
+
 source ${SCRIPTS_DIR}/version
 source ${SCRIPTS_DIR}/version-rke2
 source ${SCRIPTS_DIR}/version-rancher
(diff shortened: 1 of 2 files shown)


ibrokethecloud:
@w13915984028 now that `version_info` is the sole driver of template and addon generation, do you think we still need the chart / addon version check in `build-bundle`?

tserong (approved):

scripts/build-bundle
```diff
@@ -140,9 +145,9 @@ check_addon_chart_version_matching() {
   ls -alht ${CHARTS_DIR}
 
   echo "addon template files"
-  ls -alht ${addons_path}/pkg/templates
+  ls -alht ${addons_path}
```
Yu-Jack, line 148: Should we change this to `ls -alht ${addons_path}/*.yaml` ? I think we just care the *.yaml, right?

Vicente-Cheng (approved):
lgtm, thanks for the improvement

scripts/build-bundle
ibrokethecloud: this is to only to help with debugging.. the subsequent loop limits itself to `*.yaml`

Yu-Jack (approved):

ibrokethecloud:
@mergifyio backport v1.4

PR #791, opened 2024-08-07
build: add installer commit hash to /etc/harvester-release.yaml

tserong:
**Problem:**
It's not always obvious exactly which version of the harvester-installer code is included in a given build.

**Solution:**
Add the harvester-installer commit hash to `/etc/harvester-release.yaml`. For release builds, this should give us something like the following:

```
os: Harvester v1.3 (ff64884)
```

For local dev builds, it's a bit redundant, but I don't think there's any harm in this:

```
os: Harvester 677447c-dirty (677447c)
```

**Related Issue:**
https://github.com/harvester/harvester/issues/6285

**Test plan:**
Run `cat /etc/harvester-release` and make sure the `os` line includes the harvester-installer commit hash in parentheses.

### DIFF
--- a/scripts/package-harvester-os
+++ b/scripts/package-harvester-os
@@ -33,6 +33,7 @@ PRETTY_NAME="Harvester ${VERSION}"
 cat > harvester-release.yaml <<EOF
 harvester: ${HARVESTER_VERSION}
 harvesterChart: ${HARVESTER_CHART_VERSION}
+installer: ${COMMIT}
 os: ${PRETTY_NAME}
 kubernetes: ${RKE2_VERSION}
 rancher: ${RANCHER_VERSION}


bk201:
I'm hitting an issue when testing the upgrade. The cluster repo has an invalid image name:

截圖 2024-08-07 晚上11 13 01

The reason is the upgrade code expects the image name to like `rancher/harvester-cluster-repo:$REPO_OS_VERSION` ([link](https://github.com/harvester/harvester/blob/16b587a190a78b7836ced9c975414f711a11b7c4/package/upgrade/upgrade_manifests.sh#L338))

One alternative is to add a new key like `installer_version: <commitish>` in the `harvester-release.yaml` file. This approach provides a way to determine the included installer commit without changing the existing upgrade codes. The drawback is the commit hash is not displayed in the pretty name.

tserong:
> The reason is the upgrade code expects the image name to like `rancher/harvester-cluster-repo:$REPO_OS_VERSION` ([link](https://github.com/harvester/harvester/blob/16b587a190a78b7836ced9c975414f711a11b7c4/package/upgrade/upgrade_manifests.sh#L338))

Good catch!

I've added `installer: ${COMMIT}` - this seems most straightforward to me, rather than also changing the upgrade code.

tserong:
Full example of a build from master:

```
harvester: 16b587a1
harvesterChart: 0.0.0-master-16b587a1
installer: b08c33c
os: Harvester master
kubernetes: v1.27.13+rke2r1
rancher: v2.8.3
monitoringChart: 103.1.1+up45.31.1
loggingChart: 103.1.0+up4.4.0
kubevirt: 1.2.2-150500.8.21.1
minUpgradableVersion: ''
rancherDependencies:
  fleet:
    chart: 103.1.2+up0.9.2
    app: 0.9.2
  fleet-crd:
    chart: 103.1.2+up0.9.2
    app: 0.9.2
  rancher-webhook:
    chart: 103.0.2+up0.4.3
    app: 0.4.3
```

bk201 (approved):
LGTM, thanks!

ibrokethecloud (approved):
lgtm. thanks.

tserong:
@mergifyio backport v1.4

PR #796, opened 2024-08-09
ci: Publish ISO images on branch push

tserong:
**Problem:**
ISOs are only published on pushes to the main harvester repo. Sometimes we need an ISO built when there's only been a change in harvester-installer, for example when doing testing of the head of one of the stable branches, or of master.

**Solution:**
Duplicate the relevant ISO publishing bits from https://github.com/harvester/harvester/blob/master/.github/workflows/build.yml

**Related Issue:**
https://github.com/harvester/harvester/issues/6296

**Test plan:**
Merge this PR and make sure an ISO is published to releases.rancher.com.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ env:
   # so wet just set it to master in this case.
   DRONE_BRANCH: ${{ endsWith(github.ref_name, '/merge') && 'master' || github.ref_name }}
 jobs:
-  build-iso-images:
+  build-iso:
     name: Build ISO Images
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


tserong:
Note to @bk201 and @FrankYang0529: I _think_ I've got this right, but I don't believe I have any way to test that this works properly other than merging the PR and hoping for success.

bk201 (approved):

tserong:
Apologies for the additional force push -- I'd missed dropping some 'refs/tags/' thingies.

FrankYang0529 (approved):
LGTM. Thank you.

PR #803, opened 2024-08-13
Some system settings are not allowed during the IPXE installation

votdev:
**Problem:**
We have an allow list in the installer: https://github.com/harvester/harvester-installer/blob/14913a95dc7cd28585814b52393285fb376a9b99/pkg/config/config.go#L41 to validate if provides a valid setting, but the list doesn't catch up the changes in the [harvester backend side](https://github.com/harvester/harvester/blob/0fa939f85ea57148160b9bd5fa48125906e97c5e/pkg/settings/settings.go#L22).

**Solution:**
Add the missing supported system settings.

**Related Issue:**
https://github.com/harvester/harvester/issues/6329

**Test plan:**
Do an IPXE Install and add a missing system setting like `auto-rotate-rke2-certs`. The installation should not fail.

This can be done easily by adding the setting to the file [vagrant-pxe-harvester/ansible/roles/harvester/templates/config-create.yaml.j2](https://github.com/harvester/ipxe-examples/blob/62e9d9086451dcdc9ae9b15b3f903cd2a7f2017b/vagrant-pxe-harvester/ansible/roles/harvester/templates/config-create.yaml.j2#L41)


bk201 (approved):

Vicente-Cheng (approved):
lgtm, thanks!

bk201:
@mergifyio backport v1.4 v1.3

PR #823, opened 2024-08-26
fix: fail to build raw images

bk201:
Skip the hardware check otherwise the installation will be stuck because the build VM doesn't meet the minimum hardware requirement.


bk201:
@mergifyio backport v1.3 v1.4

tserong (approved):

ibrokethecloud (approved):

PR #834, opened 2024-09-12
Remove management interface migration script

bk201:
**Problem:**

The boot message contains "grep: /oem/99_custom.yaml: No such file or directory".

**Solution:**

Remove the management interface migration script. It's used during the v1.0.3 -> v1.1.0 upgrade path.

**Related Issue:**

https://github.com/harvester/harvester/issues/5751

**Test plan:**

- Install Harvester with the change.
- After the installation is done. Log in to the node and run
    ```
    journalctl | grep migration
    ```
    You **should not** see something like this:
    ```
    May 07 10:38:07 localhost migration.sh[1167]: grep: /oem/99_custom.yaml: No such file or directory
    May 07 10:38:07 localhost migration.sh[1166]: harvester-mgmt not found. Skipping.
    ```


bk201:
Note I didn't remove the `mkinitrd` line in https://github.com/harvester/harvester-installer/pull/334/files#diff-da6d6162c492135baaa5d8aad8873ec00226c59cd63c8edbdbd8b44b13a37639R11 in case there is something added that requires a mkinitrd run.

tserong (approved):

Vicente-Cheng (approved):

Vicente-Cheng:
> Note I didn't remove the `mkinitrd` line in https://github.com/harvester/harvester-installer/pull/334/files#diff-da6d6162c492135baaa5d8aad8873ec00226c59cd63c8edbdbd8b44b13a37639R11 in case there is something added that requires a mkinitrd run.

Now, v1.4 and master use the dracut command to replace mkinitrd
(REF: https://github.com/harvester/harvester-installer/blob/master/package/harvester-os/Dockerfile#L12)

We can check if we do not need this anymore on the next version.

PR #838, opened 2024-09-18
bump nvidia-driver-runtime chart to v0.3.0

ibrokethecloud:
**Problem:**

For certain model of GPU's / driver combination the vGPU configuration fails since the driver looks for specific firmware available in driver packaging.

This is currently not shared with the underlying host which causes issues while attempting to manage the GPU.

**Solution:**

Update the nvidia driver runtime chart to include additional mounts to make this operation successful.
The nvidia-driver-runtime chart 0.3.0 already contains the fix, and this PR bumps the addon to use the new version.

**Related Issue:**
https://github.com/harvester/harvester/issues/6487
**Test plan:**


w13915984028 (approved):
LGTM, thanks.

Vicente-Cheng (approved):

PR #849, opened 2024-09-27
Use nginx user in harvester-cluster-repo image

brandboat:
**Problem:**
This is an enhancement, use nginx instead of root user in harvester-cluster-repo image.

**Solution:**
Use nginx user in harvester-cluster-repo

**Related Issue:**

**Test plan:**
1. `make ci` to build iso, and prepare a single node harvester cluster (easy for testing)
2. Enable addon (e.g. `pcidevices-controller`) and the addon could be successfully installed.

### DIFF
--- a/package/harvester-repo/Dockerfile
+++ b/package/harvester-repo/Dockerfile
@@ -8,6 +8,12 @@ RUN zypper -n rm container-suseconnect && \
 
 COPY charts /srv/www/htdocs/charts
 
+RUN touch /run/nginx.pid && \
+    chown nginx:nginx /run/nginx.pid && \
+    chown -R nginx:nginx /srv/www/htdocs/charts
+
+USER nginx
+
 EXPOSE 80
 
 STOPSIGNAL SIGQUIT


brandboat:
After reviewing the Dockerfiles, it appears that `Dockerfile.dapper` and those under `vendor` directory are false alarms, as we don't actually use the images produced by these Dockerfiles in production environment. That leaves us with two Dockerfiles to consider:
- [harvester/package/upgrade/Dockerfile](https://github.com/harvester/harvester/blob/master/package/upgrade/Dockerfile#L9): Currently, we use chroot to manage many services on the host during the upgrade, so root privilege is necessary here.
- [harvester-installer/package/harvester-repo/Dockerfile](https://github.com/harvester/harvester-installer/blob/master/package/harvester-repo/Dockerfile): we can use nginx instead of root user in `harvester-cluster-repo`, fixed in https://github.com/harvester/harvester-installer/pull/849

bk201 (approved):

brandboat:
@mergify backport v1.3 v1.4

tserong (approved):

PR #855, opened 2024-10-01
add /lib/firmware to list of RW filesystems

ibrokethecloud:
**Problem:**

Currently /lib is mounted directly from the image file, and is readonly filesystem.

For certain driver packages, such as nvidia, a firmware file is needed which needs to be written to /lib/firmware for the driver and GPU to perform correctly. This currently fails, and attempts to leverage https://www.kernel.org/doc/html/v4.18/driver-api/firmware/fw_search_path.html does not seem to result in the correct behaviour.

The PR makes `/lib/firmware` to be remounted as an overlay fs, allowing temporary changes to be stored for drivers to function correctly.

**Solution:**

**Related Issue:**
https://github.com/harvester/harvester/issues/6487

**Test plan:**


Vicente-Cheng (approved):

tserong (approved):

ibrokethecloud:
@mergifyio backport v1.4

PR #857, opened 2024-10-04
fine tune logic to filter disks

ibrokethecloud:
**Problem:**

Logic disks managed by the raid controller will have the same serial number. This results in disks being deduplicated and ignored.

**Solution:**

From the sample data available the wwn is unique despite the serial number being the same. 

The PR adds a minor tweak to continue comparing disks even if serial number is same and check for wwn uniqueness.

**Related Issue:**
https://github.com/harvester/harvester/issues/6694
**Test plan:**

### DIFF
--- a/pkg/console/util.go
+++ b/pkg/console/util.go
@@ -1039,9 +1039,11 @@ func identifyUniqueDisks(output []byte) ([]string, error) {
 				if !ok {
 					dedupMap[disk.Serial] = disk
 				}
-				continue
 			}
 
+			// disks may have same serial number but different wwn when used with a raid array
+			// as evident in test data from a host with a raid array
+			// in this case if serial number is same, we still check for unique wwn
 			if disk.WWN != "" {
 				_, ok := dedupMap[disk.WWN]
 				if !ok {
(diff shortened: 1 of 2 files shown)


tserong (approved):
This hurts my brain a little bit, but LGTM :-)

Vicente-Cheng (approved):
lgtm, but I am curious about the actual case.
Could we have the udev info for devices with the same serial but a different WWN?

ibrokethecloud:
seems to be easily reproducible when your disks are managed by a raid controller, and subsequent luns made available to the host. In this scenario as we have seen from our own testing the disk serial numbers are the same but wwn is different.

Vicente-Cheng:
> seems to be easily reproducible when your disks are managed by a raid controller, and subsequent luns made available to the host. In this scenario as we have seen from our own testing the disk serial numbers are the same but wwn is different.

Sure, I thought I could ask for this information with @irishgordo.
Just curious. I did not notice that I had used a similar raid card before.

Feel free to merge it  : ))

ibrokethecloud:
@mergifyio backport v1.4

PR #860, opened 2024-10-08
templates: add harvester-networkfs-manager

Vicente-Cheng:
**Problem:**
We need harvester-networkfs-manager pod enabled by default.

**Solution:**
Add corresponding config on harvester managedchart

**Related Issue:**

**Test plan:**
make sure the harvester-networkfs-manager is running by default


bk201 (approved):

Vicente-Cheng:
@mergifyio backport v1.4

PR #876, opened 2024-11-02
Fix Default Route check when not using DHCP

tayterz2:
**Problem:**
#725 introduces a bug wherein if an automated deployment (after a binaries-only installation) using user-data is initiated and does not use DHCP, the installation fails with the error "No default route found. Please check the router setting on the DHCP server" regardless of the "Gateway" settings.  This is because the check is performed before a static network is set up.

**Solution:**
Add a couple of "if" checks to determine if this is a DHCP installation.  If not, skip the default route check.

**Related Issue:**
#725 #734 

**Test plan:**
Using this user-data file on an attached CDROM or USB stick, successfully perform an automated install:

```
#cloud-config
token: token
os:
  ssh_authorized_keys:
    - ssh-rsa ...        # replace with your public key
  password: p@ssword   # replace with a your password
  ntp_servers:
    - 0.suse.pool.ntp.org
    - 1.suse.pool.ntp.org
install:
  mode: create
  automatic: true
  networks:
  management_interface:
    interfaces:
      - name: ens0
      - name: ens3
    method: static
    ip: 1.2.3.4
    gateway: 1.2.3.1
    subnet_mask: 255.255.255.0
  vipMode: static
  vip: 2.3.4.5
```

### DIFF
--- a/pkg/console/install_panels.go
+++ b/pkg/console/install_panels.go
@@ -1420,7 +1420,7 @@ func addNetworkPanel(c *Console) error {
 		if err != nil {
 			return fmt.Sprintf("Failed to check default route: %s.", err.Error()), nil
 		}
-		if !isDefaultRouteExist {
+		if !isDefaultRouteExist && mgmtNetwork.Method == config.NetworkMethodDHCP {
 			return ErrMsgNoDefaultRoute, nil
 		}
 
@@ -2404,7 +2404,7 @@ func addInstallPanel(c *Console) error {
 				printToPanel(c.Gui, "Failed to check default route.", installPanel)
 				return
 			}
-			if !installModeOnly && !isDefaultRouteExist {
+			if !installModeOnly && !isDefaultRouteExist && c.config.Install.ManagementInterface.Method == config.NetworkMethodDHCP {
 				logrus.Error(ErrMsgNoDefaultRoute)
 				printToPanel(c.Gui, ErrMsgNoDefaultRoute, installPanel)
 				return


mingshuo-qiu:
The code looks OK to me. Can you give different commit title for difference commits? Or maybe simply merge 2 commits into 1 commit.

mingshuoqiu:
The code looks good to me. Can give different commit title for 2 commits or simply squash 2 commits into a single one? Thanks

starbops (commented):
Thanks for sending us the patch. However, I wonder if this will fix the problem. Please describe the steps to reproduce the case. A new GH issue for it would be nice (it helps us better understand the scope of the issue regarding the context and additional details provided). I want to reproduce the problem and validate if the fix suggested by this PR works.

Many thanks.

tayterz2:
@starbops - GH issue [GH-7280](https://github.com/harvester/harvester/issues/7280) created.

tayterz2:
> The code looks good to me. Can give different commit title for 2 commits or simply squash 2 commits into a single one? Thanks

Done - merged into a single commit.  Temporarily broke everything, but fixed now!

Thanks!
Taylor

tayterz2:
Resolves harvester/harvester#7280

starbops (commented):
@tayterz2 I just followed the reproducing steps you provided in #7280 with the PR, and I got this:
image

May I ask what Harvester ISO image are you using? Is it a custom-built ISO?

tayterz2:
@starbops - Yes, the custom ISO is required in order to include this PR (since it modifies the harvester-installer binary), but once the PR is pulled I simply follow the instructions (here)[https://github.com/harvester/harvester-installer/blob/master/README.md#building].  Your error looks like something wrong with the interface enp1s0 - is it up and connected?

Thanks,
Taylor

zackbradys:
hey @starbops @mingshuoqiu... any update on this PR? It is pretty critical to a lot of our efforts

mingshuoqiu (approved):
LGTM

starbops:
@tayterz2 @zackbradys I'm still unable to get it right by following the reproduce steps and the test plan @tayterz2 provided. The error is the same as https://github.com/harvester/harvester-installer/pull/876#pullrequestreview-2531630783. It seems that the IP address is set correctly on the `mgmt-br` interface and the `enp1s0` interface is up:
image

I built the ISO image from this PR. Should I include any other code changes I might have missed?

tayterz2:
@starbops - Could you please share your exact cloud-init file with me, so that I can try and reproduce?  It looks like you're not getting a default route for some reason.

Thanks,
Taylor

tayterz2:
Appreciate you both @starbops @mingshuoqiu !!!

starbops:
@mergify backport v1.5

zackbradys:
Hey @starbops and @mingshuoqiu, since this PR is a relatively small change, are we able to backport this to `v1.4.x` as well? It will be awhile until `v1.5.x` is supported since it is over a month until `v1.5.0` is released and we don’t support `.0` releases so our customers have to wait quite a bit for `v1.5.1`

tayterz2:
@mergify backport v1.4

tayterz2:
Created #993 manually

PR #878, opened 2024-11-06
fix: DHCP fails with a VLAN ID

bk201:
**Problem:**

Management interface with a VLAN ID fails to do DHCP configuration in the interactive installation. The testing traffic is sent without a VLAN ID.

**Solution:**

We should send the traffic via the `mgmt-br.<vlan_id>` interface.

**Related Issue:**

- https://github.com/harvester/harvester/issues/6959
- https://github.com/harvester/harvester/issues/3428

**Test plan:**

- Boot ISO
- make sure not DHCPD in vlan1 or non-vlan network.
- When configuring the management interface, select a VLAN ID and use DHCP protocol.


w13915984028:
Thanks @bk201 for the quick validation, we merged this PR together with https://github.com/harvester/harvester-installer/pull/879.

PR #882, opened 2024-11-13
feat: make curl progress bar correctly shows on the gui

Yu-Jack:
**Problem:**
When downloading ISO, the progress bar doesn't show well. The reason is that curl uses `\r` to overwrite and print the progress bar. And the console.log in `/oem/install` is not aligned.

image

**Solution:**
We should consider the `\r` character as well when scanning. And add a tag to switch different printers to print the messages.

**Related Issue:**
https://github.com/harvester/harvester/issues/5990

**Test plan:**

Install it with PXE, and check progress bar should display correctly.

## Version 1

https://github.com/user-attachments/assets/a278cd26-d7fc-4f45-beb3-362013a74314

Console log in `/oem/install/console.log`

After

image

## Version 2

Same as Version 1, but I use `"Total   Spent    Left  Speed"` as flag to trigger different printers. Then, we don't have duplicated header in `/oem/install/console.log`. But ... we still need the `endcurl` to change back to original printer.

## Version 3 (Recommended Way)

Don't use `\r` to print the curl progress bar, just use `\n`. It will look like `console.log` in Version 1. But, if user's network bandwidth is not well, the log file and console would be large.

https://github.com/user-attachments/assets/4ebb34ed-0ad4-48d4-b23a-b7ddf7adf2f9

image

### DIFF
--- a/pkg/console/util.go
+++ b/pkg/console/util.go
@@ -2,6 +2,7 @@ package console
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"crypto/tls"
 	"encoding/binary"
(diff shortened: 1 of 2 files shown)


pkg/console/util.go
```diff
@@ -581,12 +629,10 @@ func doUpgrade(g *gocui.Gui) error {
 	}
 
 	scanner := bufio.NewScanner(stdout)
-	scanner.Split(bufio.ScanLines)
 	for scanner.Scan() {
 		printToPanel(g, scanner.Text(), upgradePanel)
 	}
 	scanner = bufio.NewScanner(stderr)
-	scanner.Split(bufio.ScanLines)
```
Yu-Jack, line 589: BTW, when using `bufio.NewScanner`, the `bufio.ScanLines` is already split function by default. We don't need to set it up again.

tserong (approved):
LGTM (untested). This should also help me with https://github.com/harvester/harvester/issues/5095 :-)

bk201 (approved):

PR #884, opened 2024-11-26
Persist the bridge-nf-call-iptables across rke2 restart

mingshuoqiu:
The bridge-nf-call-iptables should be disabled by network-controller. But rke2 will toggle it back to enable. Need to make sure it stays at disabled.

**Problem:**
https://github.com/harvester/harvester/issues/7041

**Solution:**
persist the bridge-nf-call-iptables across rke2 restart

**Related Issue:**
https://github.com/harvester/harvester/issues/3960

**Test plan:**

### DIFF
--- a/package/harvester-os/files/etc/systemd/system/rke2-agent.service.d/override.conf
+++ b/package/harvester-os/files/etc/systemd/system/rke2-agent.service.d/override.conf
@@ -4,3 +4,4 @@ Wants=time-sync.target
 
 [Service]
 ExecStartPre=/usr/sbin/harv-update-rke2-server-url agent
+ExecStartPost=/sbin/sysctl -w net.bridge.bridge-nf-call-iptables=0
--- a/package/harvester-os/files/etc/systemd/system/rke2-server.service.d/override.conf
+++ b/package/harvester-os/files/etc/systemd/system/rke2-server.service.d/override.conf
@@ -4,3 +4,4 @@ Wants=time-sync.target
 
 [Service]
 ExecStartPre=/usr/sbin/harv-update-rke2-server-url server
+ExecStartPost=/sbin/sysctl -w net.bridge.bridge-nf-call-iptables=0


w13915984028 (commented):
How about another 2 options? thanks.

```
sysctl -a | grep bridge
net.bridge.bridge-nf-call-arptables = 1
net.bridge.bridge-nf-call-ip6tables = 1
net.bridge.bridge-nf-call-iptables = 0
```

https://wiki.libvirt.org/Net.bridge.bridge-nf-call_and_sysctl.conf.html

starbops (approved):
Looks like this is the only way to cope with the issue right now. The kernel tunable is hardcoded to `1` in k3s: https://github.com/k3s-io/k3s/blob/55cda2200e0f3e670970b044871f9ea09134cff6/pkg/agent/syssetup/setup.go#L48

LGTM, thank you.

mingshuoqiu:
There will be a side effect that the `net.bridge.bridge-nf-call-iptables` needs to be 1 when RKE2 lauched at the very first time after boot. So I have to add more logic to make sure the `net.bridge.bridge-nf-call-iptables` is already touched by network-controller first.

bk201 (approved):

mingshuoqiu:
> How about another 2 options? thanks.
> 
> ```
> sysctl -a | grep bridge
> net.bridge.bridge-nf-call-arptables = 1
> net.bridge.bridge-nf-call-ip6tables = 1
> net.bridge.bridge-nf-call-iptables = 0
> ```
> 
> https://wiki.libvirt.org/Net.bridge.bridge-nf-call_and_sysctl.conf.html

The `ip6tables` doesn't take effect since we support IPv4 only now.
the `arptables` is for ARP packet which won't affect the HTTP(s) traffic in this case.
I'll leave them as-is.

w13915984028 (approved):
LGTM, thanks.

bk201:
@mingshuoqiu The PR needs to be backported to the v1.4/v1.3 branch before moving the [issue](https://github.com/harvester/harvester/issues/7044) to ready for testing.

mingshuoqiu:
[@Mergifyio](https://github.com/Mergifyio) backport to v1.4.x

mingshuoqiu:
[@Mergifyio](https://github.com/Mergifyio) backport v1.3

PR #885, opened 2024-11-29
Wipe Disks Improvements

ibrokethecloud:
**Problem:**

We introduced a field in Harvester config named `wipeDisks` which when set ran a script in the installer to iterate over all disk devices visible to the host and wipe them.

This ended up being too aggressive, not to mention not all disks may contain old Harvester installations.

**Solution:**

PR introduces two new fields in the harvester config wipeDisksList and wipeAllDisks and drops the old wipeDisks introduced in v1.3

There is an extra panel in disk configuration page. The extra panel contains a drop down menu which gets rendered and populated if installer finds additional disks with existing elemental based installations which are not already used in install device or additional data device.

The extra `Wipe Disks` panels will only be available when there are additional disks visible to the host which have a partition with `COS_OEM` label and such disk is not already being used in `Install Device` or `Install Data Device`.

image

User can also pass these fields via ipxe using the following two new fields

```
install:
  wipeAllDisks: true
```

or 

```
install:
  wipeDisksList: 
     - device1
     - device2
```

`wipeAllDisks` uses the same logic as the installers `Wipe Disks Panel` where it will identify additional disks with `COS_OEM` partition which are not already being used as `Install Device` or `Install Data Device`.

**Related Issue:**
https://github.com/harvester/harvester/issues/6304
https://github.com/harvester/harvester/issues/6533

**Test plan:**

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -154,19 +154,20 @@ type Install struct {
 	ClusterPodCIDR     string `json:"clusterPodCidr,omitempty"`
 	ClusterServiceCIDR string `json:"clusterServiceCidr,omitempty"`
 
-	ForceEFI      bool   `json:"forceEfi,omitempty"`
-	Device        string `json:"device,omitempty"`
-	ConfigURL     string `json:"configUrl,omitempty"`
-	Silent        bool   `json:"silent,omitempty"`
-	ISOURL        string `json:"isoUrl,omitempty"`
-	PowerOff      bool   `json:"powerOff,omitempty"`
-	NoFormat      bool   `json:"noFormat,omitempty"`
-	Debug         bool   `json:"debug,omitempty"`
-	TTY           string `json:"tty,omitempty"`
-	ForceGPT      bool   `json:"forceGpt,omitempty"`
-	Role          string `json:"role,omitempty"`
-	WithNetImages bool   `json:"withNetImages,omitempty"`
-	WipeDisks     bool   `json:"wipeDisks,omitempty"`
+	ForceEFI      bool     `json:"forceEfi,omitempty"`
+	Device        string   `json:"device,omitempty"`
+	ConfigURL     string   `json:"configUrl,omitempty"`
+	Silent        bool     `json:"silent,omitempty"`
+	ISOURL        string   `json:"isoUrl,omitempty"`
+	PowerOff      bool     `json:"powerOff,omitempty"`
+	NoFormat      bool     `json:"noFormat,omitempty"`
+	Debug         bool     `json:"debug,omitempty"`
+	TTY           string   `json:"tty,omitempty"`
+	ForceGPT      bool     `json:"forceGpt,omitempty"`
+	Role          string   `json:"role,omitempty"`
+	WithNetImages bool     `json:"withNetImages,omitempty"`
+	WipeAllDisks  bool     `json:"wipeAllDisks,omitempty"`
+	WipeDisksList []string `json:"wipeDisksList,omitempty"`
 
 	// Following options are not cOS installer flag
 	ForceMBR bool   `json:"forceMbr,omitempty"`
--- a/pkg/console/install_panels.go
+++ b/pkg/console/install_panels.go
@@ -5,7 +5,7 @@ import (
 	"net"
 	"net/netip"
 	"os"
-	"os/exec"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
@@ -290,28 +290,6 @@ func getDataDiskOptions(hvstConfig *config.HarvesterConfig) ([]widgets.Option, e
 	return nil, nil
 }
 
-func getDiskOptions() ([]widgets.Option, error) {
-	output, err := exec.Command("/bin/sh", "-c", `lsblk -J -o NAME,SIZE,TYPE,WWN,SERIAL`).CombinedOutput()
-	if err != nil {
-		return nil, err
-	}
-
-	lines, err := identifyUniqueDisks(output)
-
-	var options []widgets.Option
-	for _, line := range lines {
-		splits := strings.SplitN(line, " ", 2)
-		if len(splits) == 2 {
-			options = append(options, widgets.Option{
-				Value: "/dev/" + splits[0],
-				Text:  line,
-			})
-		}
-	}
-
-	return options, nil
-}
-
 func addDiskPanel(c *Console) error {
 	diskConfirmed = false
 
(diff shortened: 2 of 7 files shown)


pkg/console/util.go
```diff
@@ -512,6 +515,25 @@ func doInstall(g *gocui.Gui, hvstConfig *config.HarvesterConfig, webhooks Render
...
+		if err != nil {
+			return err
+		}
+		for _, opt := range diskOpts {
+			hvstConfig.Install.WipeDisksList = append(hvstConfig.Install.WipeDisksList, opt.Value)
+		}
+	}
+
+	// prepare to wipe disks
+	for _, disk := range hvstConfig.Install.WipeDisksList {
```
ihcsim, line 531: Will it be faster to perform each iteration with a goroutine?

pkg/console/util_test.go
```diff
@@ -738,3 +1115,10 @@ func Test_identifyUniqueDisks(t *testing.T) {
 	assert.NoError(err, "expected no error while parsing disk data")
 	t.Log(out)
 }
+
+func Test_identifyUniqueDisksWithHarvesterInstall(t *testing.T) {
+	assert := require.New(t)
+	results, err := filterHarvesterInstallDisks([]byte(existingHarvesterInstalls))
+	assert.NoError(err, "expected no error while parsing disk data")
+	assert.Len(results, 1, "expected to find 1 disk from sample data")
```
ihcsim, line 1123: To confirm my understanding, in this case, the identified disk should be `sdd`, right?

pkg/console/util.go
```diff
@@ -1059,11 +1097,107 @@ func identifyUniqueDisks(output []byte) ([]string, error) {
...
+}
+
+// identifyUniqueDisksWithHarvesterInstall will identify disks which may already be in use with old Harvester
+// installs. This is done by check if a label with prefix COS exists on any of the partitions
+// and only those disks are returned for getWipeDiskOptions
+
+func identifyUniqueDisksWithHarvesterInstall() ([]string, error) {
+	output, err := exec.Command("/bin/sh", "-c", `lsblk -J -o NAME,SIZE,TYPE,WWN,SERIAL,LABEL`).CombinedOutput()
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
ibrokethecloud: sounds good.. we can log the output when the command errors

pkg/console/util.go
ibrokethecloud: the actual wipe barely takes a second which is a not much in the overall time taken during the installation process

pkg/console/util_test.go
ibrokethecloud: it will find `sdc` since both `sdc` and `sdd` are paths to the same device and sdc is found first when iterating through the list of disks

pkg/config/config.go
```diff
@@ -150,19 +150,20 @@ type Install struct {
...
-	Silent        bool   `json:"silent,omitempty"`
-	ISOURL        string `json:"isoUrl,omitempty"`
-	PowerOff      bool   `json:"powerOff,omitempty"`
-	NoFormat      bool   `json:"noFormat,omitempty"`
-	Debug         bool   `json:"debug,omitempty"`
-	TTY           string `json:"tty,omitempty"`
-	ForceGPT      bool   `json:"forceGpt,omitempty"`
-	Role          string `json:"role,omitempty"`
-	WithNetImages bool   `json:"withNetImages,omitempty"`
-	WipeDisks     bool   `json:"wipeDisks,omitempty"`
```
ihcsim, line 165: Do we need any doc to indicate that `harvester.install.wipe_disks` is now deprecated? Presumably, with this change, all existing references to `harvester.install.wipe_disks` won't work anymore, right?

pkg/console/util.go
```diff
@@ -512,6 +515,25 @@ func doInstall(g *gocui.Gui, hvstConfig *config.HarvesterConfig, webhooks Render
 		env = append(env, fmt.Sprintf("HARVESTER_ADDITIONAL_KERNEL_ARGUMENTS=%s", hvstConfig.OS.AdditionalKernelArguments))
 	}
 
+	// when WipeAllDisks is enabled then find all non installation disks with COS_ prefixed labels
+	// and add them to a list for wiping
```
ihcsim, line 519: From a UX perspective, we might need to document that this setting wipe all disks except for the installation and data disks.

pkg/console/install_panels.go
```diff
@@ -547,6 +550,34 @@ func addDiskPanel(c *Console) error {
...
+	// isWipeDisksNeeded is a helper function to render the wipeDisksPanel if needed
+	// if there are no additional disks to wipe then it checks if MBR needs to be enabled
+	// else will move on to the next apge
+	isWipeDisksNeeded := func(g *gocui.Gui, v *gocui.View) error {
+		options, err := getWipeDisksOptions(c.config)
+		if err != nil {
+			return err
+		}
+		if len(options) != 0 {
+			if slices.Contains(c.config.WipeDisksList, c.config.Device) || slices.Contains(c.config.WipeDisksList, c.config.DataDisk) {
```
ihcsim, line 563: Doesn't `getWipeDisksOptions()` already perform this filtering?

pkg/console/install_panels.go
```diff
@@ -547,6 +550,34 @@ func addDiskPanel(c *Console) error {
 		}
 		return showNext(c, persistentSizePanel)
 	}
+
+	// isWipeDisksNeeded is a helper function to render the wipeDisksPanel if needed
+	// if there are no additional disks to wipe then it checks if MBR needs to be enabled
+	// else will move on to the next apge
+	isWipeDisksNeeded := func(g *gocui.Gui, v *gocui.View) error {
+		options, err := getWipeDisksOptions(c.config)
```
ihcsim, line 558: Is there an option for the user to not select a disk, and leave this blank? Or is it required to always provide at least one disk to wipe?

simonfogliato:
I noticed this is using "blkid" from "util-linux" however that didn't seem to be included in the base install ISO v1.4.0 so I'm assuming that will also be included or is included elsewhere?

pkg/config/config.go
ibrokethecloud: there will be a docs update to cover this.

pkg/console/install_panels.go
ibrokethecloud: the panel will be hidden if there are no disks which need to be wiped. If there are multiple disks, with older installations, then assuming those disks are not used for re-install or data disk, then the panel will be rendered with list of such additional disks with old installs.

pkg/console/install_panels.go
ibrokethecloud: the extra filter is needed in case you selected an old installation disk for data or install disk, in which case we do not need to wipe the disk anymore.

example, sda, sdb, sdc are 3 disks on your system, and sda and sdb contain old installs.

in this case `getWipeDiskOptions` options will return sda and sbd.

If you choose options to have sda as install disk and sdb as datadisk then there are no disks left to wipe so the panel is hidden. 

This check allows to dynamically filter disks if you change your selection for install and data disks in their panels

ihcsim:
LGTM - can approve once conflicts are resolved.

ihcsim (approved):

Vicente-Cheng (approved):
lgtm, thanks for the enhancement!

ibrokethecloud:
@mergifyio backport v1.5

PR #897, opened 2024-12-16
Enhancement: Add IP mask field to Network Panel

mingshuoqiu:
Add IP mask field to provide better network configuration options:
- Display calculated netmask when CIDR notation is used
- Allow manual netmask input for non-CIDR IP addresses

**Problem:**
Allow users to specify the address as either `<address>/<netmask>` or CIDR. 

**Solution:**
Display calculated netmask when CIDR notation is used. And validate the netmask input for non-CIDR IP addresses

**Related Issue:**
https://github.com/harvester/harvester/issues/6295

**Test plan:**
1. Install Harvester
2. Proceed to Network Configuration Panel
3. Check the existence of `IPv4 Mask` field
4. Input a CIDR format in `IPv4 Address` field and check if the netmask generated correctly in the `IPv4 Mask` field
5. Input a non-CIDR format IP and manually input the netmask with values to validate if the netmask applies or not.


pkg/console/install_panels.go
```diff
@@ -2,7 +2,7 @@ package console
 
 import (
 	"fmt"
-	"net"
+	netutil "k8s.io/utils/net"
```
bk201, line 5: 3rd party import should be listed on the second import group

mingshuoqiu: Thanks. Fix it by force-push the PR>

bk201:
@mingshuoqiu the commits need a clean-up. Please rebase on the target branch when seeing a conflict.

bk201:
I tested the installer and saw this before the installation was just about to kick off
截圖 2025-01-13 下午4 26 07

Here is the log:
[console.log](https://github.com/user-attachments/files/18394198/console.log)

bk201:
Create [a new issue ](https://github.com/harvester/harvester/issues/7362)to track, the key is to use a static VIP. @ihcsim could be this caused by recent changes of the VIP panel?

bk201 (approved):
the PR iteself lgtm

mingshuoqiu:
@starbops @rrajendran17 Please help review. Thanks

rrajendran17:
LGTM, Thanks

rrajendran17 (dismissed):

starbops (approved):
LGTM, thank you.

pkg/console/install_panels.go
```diff
@@ -1285,6 +1289,69 @@ func addHostnamePanel(c *Console) error {
 	return nil
 }
 
+// ParseMask parses s as an IP mask, returning the result.
+// The string s must be in IPv4 dotted decimal form ("255.255.255.0").
+// Each octet must be a valid number between 0 and 255, and the bits must be
+// continuous (all 1s followed by all 0s when viewed as a 32-bit number).
+// If s is not a valid textual representation of an IP mask,
+// ParseMask returns an error describing why the parsing failed.
+func ParseMask(mask string) (IPMask, error) {
```
votdev, line 1298: I'm missing unit tests for this important function.

pkg/console/install_panels.go
```diff
@@ -1285,6 +1289,69 @@ func addHostnamePanel(c *Console) error {
...
+    // Validate mask format (must be continuous 1s followed by continuous 0s)
+    if !isValidMaskFormat(result) {
+           return nil, &ParseError{Type: "invalid subnet mask: not continuous", Text: mask}
+    }
+
+	return result, nil
+}
+
+// Helper function to validate mask format
+func isValidMaskFormat(mask IPMask) bool {
```
votdev, line 1333: I'm missing unit tests for this function. 

Why not using functions from `net` to validate the mask. You can use e.g.:
```
ones, bits := net.IPv4Mask(ip[0], ip[1], ip[2], ip[3]).Size()
valid = bits == 32 && ones >= 0 && ones <= 32
```

mingshuoqiu: force-pushed based on the suggestion to use net API for mask validation

pkg/console/install_panels.go
```diff
@@ -1285,6 +1289,61 @@ func addHostnamePanel(c *Console) error {
...
+    var valid bool
```
votdev, line 1334:
```suggestion
    valid := false
```

pkg/console/install_panels.go
```diff
@@ -1285,6 +1289,61 @@ func addHostnamePanel(c *Console) error {
...
+    outMask := net.IPv4Mask(mask[0], mask[1], mask[2], mask[3])
+    if outMask != nil {
+        ones, _ := outMask.Size()
+        cidrMask := net.CIDRMask(int(ones), 32)
+        valid = outMask.String() == cidrMask.String()
+    } else {
```
votdev, line 1340: If `valid := false` is set above, then the whole “else” branch is superfluous.
```suggestion
    }
```

mingshuoqiu: Thanks @votdev . I've fixed it based on you suggestion.

votdev:
The error message looks strange: `invalid invalid mask format ...`

grafik

votdev (approved):
LGTM

PR #925, opened 2024-12-31
fix: Skip wipe of live usb device.

simonfogliato:
**Problem:**
The option harvester.install.wipe_disks=true should skip the live usb device.

**Solution:**
Skip the live usb device mounted as the root rfs.

**Related Issue:**
https://github.com/harvester/harvester/issues/6533

### DIFF
--- a/package/harvester-os/files/usr/sbin/harv-install
+++ b/package/harvester-os/files/usr/sbin/harv-install
@@ -30,6 +30,9 @@ clear_disk_label()
         echo "wiping all disks"
         for disk in $(lsblk -d -n -J -o NAME,TYPE | jq -r '.blockdevices[] | select(.type == "disk") | .name')
         do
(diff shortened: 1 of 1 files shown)


ihcsim:
@simonfogliato Does https://github.com/harvester/harvester-installer/pull/885 do what you want? This seems like a subset of what that one is doing.

simonfogliato:
Interesting. I noticed https://github.com/harvester/harvester-installer/pull/885 is using "blkid" from "util-linux" however that didn't seem to be included in the base install ISO v1.4.0 so I'm assuming that will also be included? I think "blkid" will naturally not list USB devices since they are not the same as a disk with an associated UUID.

simonfogliato:
I was wrong, it is included in the ISO but I was not able to call the executable. I was playing around on the live ISO looking in /run/blkid/blkid.tab and I do see the UUID with a "COS_LIVE" label listing the USB device. I'm not sure but as far as I can tell it may list and attempt to touch the USB.

simonfogliato:
Agreed. I will close this.

PR #927, opened 2025-01-10
Bump golang to v1.23 plus some additional cleanup

tserong:
**Problem:**
Need to bump golang to v1.23. While doing that I found a couple of other little things to clean up which I have also included in this PR as separate commits.

**Solution:**
This PR

**Related Issue:**
https://github.com/harvester/harvester/issues/7336

**Test plan:**
N/A


bk201 (approved):

Vicente-Cheng (approved):
thanks!

PR #930, opened 2025-01-15
Set VipMode value in static mode

w13915984028:
**Problem:**

When using static mode VIP, the installer is blocked.

**Solution:**

Set proper VipMode value.

**Related Issue:**
https://github.com/harvester/harvester/issues/7362

**Test plan:**

Install Harvester in any mode, it should success.

### DIFF
--- a/pkg/console/install_panels.go
+++ b/pkg/console/install_panels.go
@@ -2581,6 +2581,10 @@ func addVIPPanel(c *Console) error {
 
 		c.config.Vip = vip
 		c.config.VipHwAddr = ""
+		// gotoVipPanel is only called in DHCP mode, it is still empty in static mode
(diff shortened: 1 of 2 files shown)


ihcsim (approved):
Thanks for the fix :+1:.

tserong (approved):
LGTM, but the original issue is targeting v1.5.0, and says "The v1.4.0 ISO has no such issue", so I assume we don't need to backport this to v1.4.0? Or did you subsequently reproduce it there too?

w13915984028:
@tserong You are right, the issue is existing on v1.5 and master, v1.4 does not require the fix. The backport PR 
 https://github.com/harvester/harvester-installer/pull/935 has been closed.

PR #931, opened 2025-01-15
Allow for a Fully-Qualified Hostname

tayterz2:
**Problem:**
When using domain names in hostnames, inform RancherD and the Dashboard that this is the proper hostname.

**Solution:**
Add hostname to rancherd node adds, and perform "hostnamectl" instead of "hostname" to pull the FQDN from the hosts.

**Related Issue:**
harvester/harvester#7312

**Test plan:**
Install Harvester with an FQDN as hostname (taylor.rancher.io) and observe that same hostname in the "hosts" and "node" fields throughout.

### DIFF
--- a/pkg/console/dashboard_panels.go
+++ b/pkg/console/dashboard_panels.go
@@ -351,9 +351,7 @@ func getNodeInfo() string {
 	)
 
 	// find node hostname
-	cmd = `hostname | tr -d '\r\n'`
-	out, err = exec.Command("/bin/sh", "-c", cmd).Output()
-	hostname = string(out)
+	hostname, err = GetFullHostname()
 	if err != nil || hostname == "" {
 		logrus.Warnf("node didn't have a hostname")
 		hostname = ""
@@ -497,7 +495,7 @@ func isPodReady(namespace string, labelSelectors ...string) bool {
 }
 
 func nodeIsPresent() bool {
-	hostname, err := os.Hostname()
+	hostname, err := GetFullHostname()
 	if err != nil {
 		logrus.Errorf("failed to get hostname: %v", err)
 		return false
--- a/pkg/console/install_panels.go
+++ b/pkg/console/install_panels.go
@@ -2985,7 +2985,7 @@ func configureInstallModeDHCP(c *Console) {
 
 func checkDHCPHostname(c *config.HarvesterConfig, generate bool) {
 	if c.Hostname == "" {
-		hostname, err := os.Hostname()
+		hostname, err := GetFullHostname()
 		if err != nil {
 			logrus.Errorf("error fetching hostname from underlying OS: %v", err)
 		}
--- a/pkg/console/util.go
+++ b/pkg/console/util.go
@@ -1235,3 +1235,24 @@ func runCommand(cmd *exec.Cmd) error {
 	}
 	return err
 }
+
+// Get fully-qualified Hostname
+func GetFullHostname() (string, error) {
+	var (
+		hcmd      string
+		hostname string
+		out      []byte
+		err      error
+	)
+
+	// find node hostname
+	hcmd = `hostnamectl --static | tr -d '\r\n'`
+	out, err = exec.Command("/bin/sh", "-c", hcmd).Output()
+	if err != nil {
+		logrus.Errorf("failed to get hostname: %v", err)
+		return "", err
+	}
+
+	hostname = string(out)
+	return hostname, nil
+}
--- a/pkg/config/templates/rancherd-config.yaml
+++ b/pkg/config/templates/rancherd-config.yaml
@@ -4,6 +4,7 @@ role: agent
 {{- else -}}
 role: cluster-init
 {{- end }}
+nodeName: {{ .Hostname }}
 token: {{ printf "%q" .Token }}
 kubernetesVersion: {{ .RuntimeVersion }}
 rancherVersion: {{ .RancherVersion }}


pkg/console/dashboard_panels.go
```diff
@@ -497,7 +497,17 @@ func isPodReady(namespace string, labelSelectors ...string) bool {
...
+		hcmd      string
+		hostname string
+		out      []byte
+		err      error
+	)
+
+        // find node hostname
+        hcmd = `hostnamectl --static | tr -d '\r\n'`
+        out, err = exec.Command("/bin/sh", "-c", hcmd).Output()
+        hostname = string(out)
```
Yu-Jack, line 510: There are few places where calls `os.Hostname()` function.  You could create a new function to replace them.

pkg/console/dashboard_panels.go
```diff
@@ -351,7 +351,7 @@ func getNodeInfo() string {
 	)
 
 	// find node hostname
-	cmd = `hostname | tr -d '\r\n'`
+	cmd = `hostnamectl --static | tr -d '\r\n'`
```
Yu-Jack, line 354: Include this one.

Yu-Jack (changes requested):

pkg/console/dashboard_panels.go
tayterz2: Agreed, done!

tayterz2:
Added the requested function and tested on the test cases listed by @Yu-Jack [here](https://github.com/harvester/harvester/issues/7312#issuecomment-2591560667).

Yu-Jack:
@tayterz2 BTW, please use rebase master, thank you!

tayterz2:
@Yu-Jack - I rebased 2 days ago, and currently github reports this branch as caught up.  What am I missing?

Thanks!
Taylor

Yu-Jack:
Hi @tayterz2 I think you can rebase again. Or just open another branch, cherry-pick your commit and push it again.

Yu-Jack:
@tayterz2  If you use rebase, the commits wouldn't be 19. Could try it one more time? BTW, you could squash your commits into one commit? Thanks.

pkg/console/util.go
```diff
@@ -1235,3 +1235,24 @@ func runCommand(cmd *exec.Cmd) error {
...
+func GetFullHostname() (string, error) {
+	var (
+		hcmd      string
+		hostname string
+		out      []byte
+		err      error
+	)
+
+	// find node hostname
+	hcmd = `hostnamectl --static | tr -d '\r\n'`
```
ihcsim, line 1249: @Yu-Jack with the changes in https://github.com/rancher/rancherd/pull/50, do we still need to use `hostnamectl` or can we switch back to Go's `os.Hostname()` (which uses `uname` or `/proc/sys/kernel/hostname`, if Linux).

pkg/console/util.go
```diff
@@ -1235,3 +1235,24 @@ func runCommand(cmd *exec.Cmd) error {
...
+	out, err = exec.Command("/bin/sh", "-c", hcmd).Output()
+	if err != nil {
+		logrus.Errorf("failed to get hostname: %v", err)
+		return "", err
+	}
```
ihcsim, line 1254: We should use `CombinedOutput()` to capture both stdout and stderr. Also, `if err !=nil`, `out` should be logged too to capture the actual Linux outputs. In most cases, `err` only capture Go-specific errors. It will be great if we can revert back to using `os.Hostname()`, for portability.

pkg/console/util.go
tayterz2: @ihcsim not exactly - os.Hostname() can't return the Fully-Qualified hostname of a host, so rancher/harvester hosts end up with shortened names.  This is a huge problem for customers with larger Harvester clusters spanning multiple domain names.  @Yu-Jack 's fix only addresses the rancherd end of the problem (primarily shows up in node joins), not the Harvester console or install (first node) scenarios.

tayterz2:
> @tayterz2 If you use rebase, the commits wouldn't be 19. Could try it one more time? BTW, you could squash your commits into one commit? Thanks.

@Yu-Jack 
Yes - my git ignorance is showing.  Please take a look now, I think I've gotten it fixed up.

Thanks,
Taylor

pkg/console/util.go
ihcsim: i was hoping that once we appease rancher everywhere, the fqdn becomes _the_ (static and transient) hostname where we [no longer need to use the short name](https://github.com/rancher/rancherd/blob/6f7255eb0eafd3294e3ec5140d617ff1831b0b9b/pkg/resources/resources.go#L55-L62).

Yu-Jack: There are four sub-issues in this topic

1. For first node, we don't bring customized hostname. (The fix is https://github.com/harvester/harvester-installer/pull/931/files#diff-fb345406e175053d95f23645fa17f1ba365649c8baafd1df4313cf662aaa64c5)
2. For joiner node, we don't bring customized hostname. (The fix is https://github.com/rancher/rancherd/pull/50)
3. Which name should Harvester Install show? Which command should we use to fetch proper hostname?
4. `hostname -f` only shows the short one. We don't have this setting `127.0.0.1 harvester1.dap.sys harvester1` in `/etc/hosts`

For the 1 and 2, there is no doubt. We should pass customized hostname to rancher. But, for 3 and 4, we need a discussion about it.

Yu-Jack:
> > @tayterz2 If you use rebase, the commits wouldn't be 19. Could try it one more time? BTW, you could squash your commits into one commit? Thanks.
> 
> @Yu-Jack Yes - my git ignorance is showing. Please take a look now, I think I've gotten it fixed up.
> 
> Thanks, Taylor

@tayterz2  Nice, it's good now. But, do you think we need to split this PR to different PRs to focus on different issues?

pkg/console/util.go
ihcsim: why do we need the short name? i thought the only place where the short name comes from is https://github.com/rancher/rancherd/blob/6f7255eb0eafd3294e3ec5140d617ff1831b0b9b/pkg/resources/resources.go#L55-L62.

Yu-Jack: Harvester Installer Dashboard uses `hostname` or `os.Hostname` to fetch it. Even, you pass FQDN to rancherd in this PR, it stills shows short one. But, I think we should display FQDN in Harvester Installer Dashboard, so we need to change the way how it fetches.

tayterz2:
@

> > > @tayterz2 If you use rebase, the commits wouldn't be 19. Could try it one more time? BTW, you could squash your commits into one commit? Thanks.
> > 
> > 
> > @Yu-Jack Yes - my git ignorance is showing. Please take a look now, I think I've gotten it fixed up.
> > Thanks, Taylor
> 
> @tayterz2 Nice, it's good now. But, do you think we need to split this PR to different PRs to focus on different issues?

Completely understand why it might be 2 issues and 2 PRs from you devs' perspective - but from the customer's viewpoint it is only one.  The FQDN doesn't work, or it does.  I'll leave to you guys to let me know which way to go - happy to accommodate either way so long as we can get both fixes into v1.5!

pkg/console/dashboard_panels.go
```diff
@@ -497,7 +495,7 @@ func isPodReady(namespace string, labelSelectors ...string) bool {
 }
 
 func nodeIsPresent() bool {
-	hostname, err := os.Hostname()
```
Yu-Jack, line 500: About `os.Hostname`, I'm thinking we maybe just need to change behavior here. We could keep host name as original one. 

But, we could fetch host name from `os.Hostname()` and `/etc/hostname` to check if node name matches one of them.

It seems we need to separate these two names for now.

> `os.Hostname()` is from `/proc/sys/kernel/hostname`.

pkg/console/dashboard_panels.go
```diff
@@ -351,9 +351,7 @@ func getNodeInfo() string {
 	)
 
 	// find node hostname
-	cmd = `hostname | tr -d '\r\n'`
-	out, err = exec.Command("/bin/sh", "-c", cmd).Output()
-	hostname = string(out)
```
Yu-Jack, line 356: `os.Hostname()`

pkg/console/install_panels.go
```diff
@@ -2985,7 +2985,7 @@ func configureInstallModeDHCP(c *Console) {
 
 func checkDHCPHostname(c *config.HarvesterConfig, generate bool) {
 	if c.Hostname == "" {
-		hostname, err := os.Hostname()
```
Yu-Jack, line 2988: `os.Hostname()`

Yu-Jack (changes requested):

pkg/console/install_panels.go
Yu-Jack, line 2988: including this one

pkg/console/dashboard_panels.go
Yu-Jack, line 356: including this one

pkg/console/dashboard_panels.go
Yu-Jack: I think we could keep it short in host name. We should display node name instead of host name in console.

before
```
Hostname: harvester1
IP Address: x.x.x.x
```

after
```
Node Name: harvester1.dap.sys
IP Address: x.x.x.x
```

Besides, we should figure it out how to add FQDN in /etc/hosts to make `hostname -f` produce the FQDN.

cc @bk201 @ibrokethecloud

ihcsim: I think what's the important here is what the host's kernel says its hostname is. In this case, it's the output of `/proc/sys/kernel/hostname` (which is used by `os.Hostname()` as @Yu-Jack and I pointed out [before](https://github.com/harvester/harvester-installer/pull/931#discussion_r1972455827)). 

Normally, the hostname is the leftmost part of the name before the first period. Anything after the first period is the domain name. In some heterogeneous environments, DHCP servers might automatically supply the domain name to the hosts, which might lead to something like `harvester1.dap.sys.dap.sys`.

I think fixing the original issue of the duplicated k8s node is more important than deciding whether to display the FQDN or the short name. (I still can't quite figure out where the short name comes from - probably an OS thing? Debian variants don't behave like that.)

@Yu-Jack I agree with the earlier proposal of displaying both formats, whether in this PR or a future one. I don't know outside of the context of k8s, if `node name` means anything. How about something like this:

```
Hostname: harvester2
FQDN: harvester2.dap.sys
```

tayterz2: I'm just the peanut gallery here, but I love this proposal @ihcsim !

ihcsim: Technically, FQDN isn't even correct because FQDN should come from a DNS lookup. But for our purpose here, it should suffice. I'm fine with displaying both fields even if they are the same, just for consistency, instead of having logic to decide whether to display one or both fields.

Yu-Jack: So, it would be like this:

```
Hostname: harvester1
Node Name: harvester1.dap.sys
```

```
Hostname: harvester1
Node Name: harvester1
```

Yu-Jack: If it's okay, we could go with this one. Then, I'll keep digging the `/etc/hosts` part.

Yu-Jack:
Duplicated, fix has been merged in https://github.com/harvester/harvester-installer/pull/1011

bk201:
@Yu-Jack I do want to recognize @tayterz2 's contribution to the fix. It will be nice to make him a co-author in the PR https://github.com/harvester/harvester-installer/pull/1011. But since the PR has already merged, we can add him to [the contribution list](https://github.com/harvester/release-notes/issues/43) in 1.6.0

Yu-Jack:
@bk201 Thanks for reminding! 

@tayterz2 Sorry for that. I'll add you in [contribution list](https://github.com/harvester/release-notes/issues/43). I'm really appreciate that you try to fix it, thanks for your contribution!

tayterz2:
Thank you all very much, no recognition needed - team effort!

PR #932, opened 2025-01-17
arm upgrade fix: patch harvester-release.yaml for arm iso builds

ibrokethecloud:
**Problem:**

Currently the harvester iso build includes the kubevirt version from the values.yaml file in the harvester/harvester chart.
During the ARM packaging we patch the kubevirt version to upstream version, to allow correct images to be included in the packaging.

However the changes are reverted before the OS packaging, which results in the `harvester-release.yaml` containing the values from the amd64 build. 

This causes the upgrade to fail as the upgrade helper waits for the kubevirt version to be the one defined in the `harvester-release.yaml`. This never reconciles as we end up using the amd64 kubevirt patch version which results in an upgrade being stuck.

**Solution:**

minor changes to allow patching of harvester-release.yaml for ARM builds. This change is needed to ensure correct kubevirt version is included in the harvester-release.yaml. This is essential to ensure the upgrade manifests script can successfully reconcile the kubevirt version with expected version as part of the upgrade process

**Related Issue:**
https://github.com/harvester/harvester/issues/6257

**Test plan:**

To test:
* Install Harvester v1.4.0 on ARM nodes
* Create a custom version object using the iso built from this change or harvester-master-arm64.iso once the PR is merged
* Trigger upgrade of Harvester
* Harvester Upgrade should complete successfully

### DIFF
--- a/scripts/package-harvester-os
+++ b/scripts/package-harvester-os
@@ -15,6 +15,7 @@ fi
 
 mkdir -p ${ARTIFACTS_DIR}
 
+
 source ${SCRIPTS_DIR}/version
 source ${SCRIPTS_DIR}/version-rke2
 source ${SCRIPTS_DIR}/version-rancher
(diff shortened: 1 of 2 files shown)


ibrokethecloud:
Not needed anymore since with new sles based kubevirt images the patching of harvester chart is not required.

PR #933, opened 2025-01-17
chore: update to v1.5 branch for harvester and addons repos

starbops:
Update harvester and addons to use v1.5 repos for the v1.5 stable branch.


tserong (approved):
LGTM, but we should probably also make the same change in `scripts/test` as well, in case someone tries to run `make test` by itself

starbops:
> LGTM, but we should probably also make the same change in `scripts/test` as well, in case someone tries to run `make test` by itself

@tserong Good catch! I've updated the script, PTAL. I will also update the wiki page. Thanks!

tserong (approved):

PR #938, opened 2025-01-23
change multipath default behaviour

ibrokethecloud:
**Problem:**

Currently when externalDisk support is not enabled, the installer adds a default kernel argument `multipath=off` via `third_party_kernel_arguments`.

This works fine, however it breaks CSI's needing multipath since it cannot be enabled due to the kernel argument.

Disabling `multipath=off` from `third_party_kernel_args` breaks the boot, when OS has been installed to nvme, as multipath attempts to create multiple paths for nvme disks.
**Solution:**

To fix this a minor change has been made to default kernel argument being passed when multipath is not needed. This allows dracut to still boot with nvme disks as it ignores multipath module packaged in dracut, while allowing multipathd to be enable subsequently

In addition a default multipath config is added to ignore longhorn based volumes from multipath

**Related Issue:**
https://github.com/harvester/harvester/issues/7438

**Test plan:**

Failure case:
* Install v1.4.0 to a host with nvme disks, with the nvme disk being used as the install disk
* After installation is completed, ssh into the node and attempt to enable multipathd
```
systemctl enable multipathd
systemctl start multipathd
```
This will fail as can be checked via `journalctl -fu multipathd`
* remove the `multipath=off` kernel argument by removing it from /oem/grubenv
`grub2-editenv /oem/grubenv set third_party_kernel_args=""`
* reboot node

Expect node boot to fail

Success case:
* Install build with this change to a host with nvme disks, with the nvme disk being used as the install disk
* After installation is completed, ssh into the node and attempt to enable multipathd
```
systemctl enable multipathd
systemctl start multipathd
```
* create a VM using any image
* once VM is running shell into the node and check `multipath -ll`
* No longhorn volumes should be reported

This will be successful

### DIFF
--- a/pkg/config/cos.go
+++ b/pkg/config/cos.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"io/ioutil"
@@ -179,6 +180,9 @@ func ConvertToCOS(config *HarvesterConfig) (*yipSchema.YipConfig, error) {
 	if err := setupExternalStorage(config, &initramfs); err != nil {
 		return nil, err
 	}
+
+	// disable multipath for longhorn
+	disableLonghornMultipathing(&initramfs)
 
 	// TOP
 	if cfg.Mode != ModeInstall {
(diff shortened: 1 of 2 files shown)


tserong (approved):

tserong (approved):
Neat!

ihcsim (approved):

Vicente-Cheng (approved):
thanks for the enhancement!

ibrokethecloud:
@mergifyio backport v1.5 v1.4

PR #942, opened 2025-01-31
fix kernel command line for management interface

m-ildefons:
Fix parsing management interface configuration from the kernel command line.
The parser now accepts more different specifications for the management interface, including just the systemd name for the device. Additionally, should the configuration not be parseable (e.g. due to a typo), the installer doesn't crash anymore, but rather produces a human readable error.

related-to: harvester/harvester#7470

**Problem:**

**Solution:**

**Related Issue:**

**Test plan:**

### DIFF
--- a/pkg/util/cmdline.go
+++ b/pkg/util/cmdline.go
@@ -83,17 +83,65 @@ func toNetworkInterfaces(data map[string]interface{}) error {
 
 	outDetails := make([]interface{}, 0, len(ifDetails))
 	for _, v := range ifDetails {
-		tmpStrings := strings.SplitN(v, ":", 2)
-		n := make(map[string]interface{})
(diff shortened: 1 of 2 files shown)


ibrokethecloud (approved):
lgtm. thanks.

bk201 (approved):

brandboat:
gentle ping @m-ildefons, do we need to update the doc as well? https://docs.harvesterhci.io/v1.5/install/harvester-configuration#installmanagement_interface

m-ildefons:
I don't think a docs update is necessary.
This fix does not affect the config syntax, only the kernel command line.
The documented way to specify parameters on the kernel command line is still valid. This fix just makes the installer more robust when the parameters are specified in an incorrect way.

PR #957, opened 2025-02-10
drop patching of harvester chart for arm builds

ibrokethecloud:
**Problem:**

PR drops patching of the harvester chart during arm builds.

This was needed to use upstream images and also remove emulated machine types.

With Harvester 1.5.x, we will be using sles based kubevirt v1.4.x images where multi-arch images are available.

As a result the patching of the chart is no longer needed.

**Solution:**

**Related Issue:**
https://github.com/harvester/harvester/issues/7098

**Test plan:**


Vicente-Cheng (approved):
lgtm, thanks!

we could find more info from the harvester PR: https://github.com/harvester/harvester/pull/7565

tserong (approved):

ibrokethecloud:
@mergifyio backport v1.5

PR #991, opened 2025-02-25
config: add rke2 config to improve the device permission

Vicente-Cheng:
**Problem:**
We encounter the permission issue when using Longhorn v2 volume with CDI

**Solution:**
Set `device_ownership_from_security_context` to true for better permission handling.

**Related Issue:**
https://github.com/harvester/harvester/issues/1199

**Test plan:**
Please see the test plan on https://github.com/harvester/harvester/pull/7640


WebberHuang1118 (approved):
LGTM, thanks.

bk201 (approved):

PR #1005, opened 2025-03-06
Add only vlan specified by user to mgmt-br and mgmt-bo during installation

rrajendran17:
**Problem:**
During Harvester installation, the system attempts to add all VLAN IDs to the management bond/bridge. With Mellanox ConnectX-6 NICs (which support 512 hardware-offloaded VLANs and have hardware VLAN offloading enabled by default), this causes boot log warnings and prevents management VLAN 2301 from being added to the hardware offloaded VLAN list.

**Solution:**
Add only configured vlans to the mgmt-br and mgmt-bo

**Related Issue:**
https://github.com/harvester/harvester/issues/7650

**Test plan:**
1.Select vlan id during harvester-installation
2.configure and proceed with installation
3.Once installed, verify "bridge vlan show"
mgmt-br and mgmt-bo should list the below for example 

bridge vlan show
port              vlan-id  
mgmt-br           1 PVID Egress Untagged
                  2021
mgmt-bo           1 PVID Egress Untagged
                  2021

where 2021 is the user configured vlan on the mgmt interface instead of all 2-4094 vlans

This PR should be merged along with https://github.com/harvester/network-controller-harvester/pull/156

### DIFF
--- a/pkg/config/templates/wicked-setup-bridge.sh
+++ b/pkg/config/templates/wicked-setup-bridge.sh
@@ -11,7 +11,7 @@ case $ACTION in
 
 	post-up)
 		# accept all vlan, PVID=1 by default
-		bridge vlan add vid 2-4094 dev $INTERFACE self
-		bridge vlan add vid 2-4094 dev {{ . }}
+		bridge vlan add vid {{ .VlanID }} dev $INTERFACE self
+		bridge vlan add vid {{ .VlanID }} dev {{ .IntfName }}
 		;;
 esac
--- a/pkg/config/cos.go
+++ b/pkg/config/cos.go
@@ -522,6 +522,7 @@ func UpdateManagementInterfaceConfig(stage *yipSchema.Stage, mgmtInterface Netwo
 			Method:      NetworkMethodNone,
 			BondOptions: mgmtInterface.BondOptions,
 			MTU:         mgmtInterface.MTU,
+			VlanID:      mgmtInterface.VlanID,
 		}
 		if err := updateBond(stage, MgmtBondInterfaceName, &bondMgmt); err != nil {
 			return "", err
@@ -580,7 +581,12 @@ func updateBond(stage *yipSchema.Stage, name string, network *Network) error {
 		Group:       0,
 	})
 
-	postUpScript, err := render("wicked-setup-bond.sh", MgmtInterfaceName)
+	mgmtData := map[string]interface{}{
+		"VlanID":   network.VlanID,
+		"IntfName": MgmtBondInterfaceName,
+	}
+
+	postUpScript, err := render("wicked-setup-bond.sh", mgmtData)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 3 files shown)


pkg/config/templates/wicked-setup-bridge.sh
```diff
@@ -11,7 +11,7 @@ case $ACTION in
 
 	post-up)
 		# accept all vlan, PVID=1 by default
-		bridge vlan add vid 2-4094 dev $INTERFACE self
-		bridge vlan add vid 2-4094 dev {{ . }}
+		bridge vlan add vid {{ .VlanID }} dev $INTERFACE self
+		bridge vlan add vid {{ .VlanID }} dev {{ .IntfName }}
```
w13915984028, line 15: When user does not input VID, what's the default value is used, 0 or 1?
Should below two lines be fully removed on above case?
```
		bridge vlan add vid {{ .VlanID }} dev $INTERFACE self
		bridge vlan add vid {{ .VlanID }} dev {{ .IntfName }}
```
L13 comment alose needs to be updated.

thanks.

rrajendran17: When no vlan input from user, the default value is 0.
As vid 0 is not a valid value, its automatically skipped in kernel.so an explicit check is not needed to skip vid 0.

But if we need to be more safer, then I could add the following check before "bridge vlan add...."

#skip bridge vlan setting when no custom vlan specified by user
if [ {{ .VlanID }} -eq 0 ]; then
   exit 0
fi

w13915984028: such a line
```
bridge vlan add vid 0 dev $INTERFACE self
```
will be a bit misleading, even though kernel can handle it correctly.

if user does not set vid, or clears vid, could we convert this line to:

```
bridge vlan add vid 1 dev $INTERFACE self
```

w13915984028 (approved):
LGTM, thanks.

w13915984028:
Just recall another scenario:

How to handle this in the upgrade path? Old cluster has the legacy `.sh` file, when and how to replace them, did you check the upgrade code path. 

https://github.com/harvester/harvester/blob/master/package/upgrade/upgrade_node.sh
https://github.com/harvester/harvester/blob/master/package/upgrade/upgrade_manifests.sh

pkg/config/templates/wicked-setup-bridge.sh
rrajendran17: > such a line
> 
> ```
> bridge vlan add vid 0 dev $INTERFACE self
> ```
> 
> will be a bit misleading, even though kernel can handle it correctly.
> 
> if user does not set vid, or clears vid, could we convert this line to:
> 
> ```
> bridge vlan add vid 1 dev $INTERFACE self
> ```

If we skip "bridge vlan ..." when the vlan id with 0, then we will not have any entry for "bridge vlan add vid 0 dev $INTERFACE self", and since default PVID set is 1, I believe we do not have to explicitly make an entry as below for vid 0.

bridge vlan add vid 1 dev $INTERFACE self

rrajendran17:
> Just recall another scenario:
> 
> How to handle this in the upgrade path? Old cluster has the legacy `.sh` file, when and how to replace them, did you check the upgrade code path.
> 
> https://github.com/harvester/harvester/blob/master/package/upgrade/upgrade_node.sh https://github.com/harvester/harvester/blob/master/package/upgrade/upgrade_manifests.sh

Sure, I did not think about the upgrade path, I will check how this can affect upgrade scenario and also the reference you provided.Thanks for the suggestion.

PR #1014, opened 2025-03-26
enhancement: use rancher/harvester-os:sl-micro-6 as base OS

tserong:
A few things seem to have changed with the newer version of elemental in SL Micro 6, notably:

- rootfs.squashfs is no longer in the root directory of the ISO, rather it's in an architecture specific subdirectory under /boot.
- The ARM initrd and kernel on the ISO moved from /boot/x86_64/loader/ to /boot/arm64/loader/ (although that begs the question - why were they under /boot/x86_64/loader previously?)
- The initrd and kernel symlinks under /boot in the baseos image may be to either other files in /boot (without extra path information), or to paths elsewhere under /usr/lib/modules/, hence the change to `readlink -f` when extracting these files.

Related issue: https://github.com/harvester/harvester/issues/7025

_NOTE: I DO NOT INTEND TO MERGE THIS, I JUST WANTED TO SEE IF THE ISO BUILD SUCCEEDS FOR BOTH ARCHITECTURES_

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -87,15 +87,19 @@ jobs:
         credentials_json: '${{ env.GOOGLE_AUTH }}'
 
     - name: upload-iso
-      uses: 'google-github-actions/upload-cloud-storage@v2'
-      if: ${{ startsWith(github.ref, 'refs/heads/') }}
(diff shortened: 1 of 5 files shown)


tserong:
Old experiment, closing

PR #1015, opened 2025-03-27
feat: bump RKE2 to v1.31.7+rke2r1

FrankYang0529:
Bump RKE2 to v1.31.7+rke2r1

**Related Issue:**
https://github.com/harvester/harvester/issues/7948

**Test plan:**
* Create an airgapped 3-node cluster without error.
* Upgrade from v1.4.2 without error.


tserong (approved):

FrankYang0529:
@mergify backport v1.4

PR #1020, opened 2025-04-01
Disable AuthorizeNodeWithSelectors on v1.5.0

w13915984028:
**Problem:**

After https://github.com/harvester/harvester-installer/pull/1016, k8s is v1.3.2, Harvester hits limitation/issue:

https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/4601-authorize-with-selectors/README.md
https://github.com/canonical/microk8s/issues/4802

**Solution:**

Per https://github.com/rancher/rke2/discussions/3194 and test on Harvester, the current solution is to add the related info to Harvester config.

**Related Issue:**

https://github.com/harvester/harvester/issues/7963

**Test plan:**

1. Install multi-node cluster
2. Each node should show "Harvester Cluster` Status `Ready`

https://github.com/harvester/harvester/issues/7963#issuecomment-2769821218

**Note**
Upgrade PR will be submitted to Harvester repository.

### DIFF
--- a/.github/workflows/scan.yml
+++ b/.github/workflows/scan.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   build:
     name: Build
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
--- a/pkg/config/templates/rke2-90-harvester-server.yaml
+++ b/pkg/config/templates/rke2-90-harvester-server.yaml
@@ -11,3 +11,5 @@ kubelet-arg:
 {{- end }}
 {{- end }}
 audit-policy-file: /etc/rancher/rke2/config.yaml.d/92-harvester-kube-audit-policy.yaml
+kube-apiserver-arg:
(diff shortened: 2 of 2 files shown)


ihcsim:
Good find! When this feature goes GA, disabling the feature gate won't work though. Can we add the field selector to the `kubectl get po` command in the `isPodReady()` method?

E.g., this works on my worker node:
```
kubectl get po -n harvester-system \
    --kubeconfig=/var/lib/rancher/rke2/agent/kubelet.kubeconfig \
    --field-selector spec.nodeName=$workerNodename \
    -l "app.kubernetes.io/name=harvester" \
    -l "app.kubernetes.io/component=apiserver"
```

Vicente-Cheng:
Thanks @w13915984028 for finding this issue!

I have the same question as @ihcsim. Since the `AuthorizeNodeWithSelectors` goes GA, should we use the formal way to get the pod?

cc @ibrokethecloud, @bk201

w13915984028:
This solution requres more efforts on the upgrade path.

Replaced by https://github.com/harvester/harvester-installer/pull/1022

PR #1024, opened 2025-04-07
Remove versions not support ARM in image-lists-archive

brandboat:
**Problem:**
In Harvester v1.4.0-rc2 ARM iso, under directory bundle/harvester/images-lists-archive we can see image lists for versions v1.1.0 ~ v1.2.2, which do not support the ARM architecture. Keeping these files does no harm, as the images-lists-archive is mainly used to compare current and previous versions, helping identify unused images for image cleanup after upgrade. Since ARM support started with v1.3.0, we never reference versions prior to that.

**Solution:**
Remove all versions prior to v1.3.0 under bundle/harvester/images-lists-archive in ARM iso.

**Related Issue:**
https://github.com/harvester/harvester/issues/6794

**Test plan:**


Vicente-Cheng (approved):
lgtm, thanks!

innobead (approved):

PR #1025, opened 2025-04-08
Set default mode to Active-Backup for bond

rrajendran17:
Problem:
The default bond mode set during harvester installation using seeder or ipxe is balance-tlb

Expected Behavior:
The recommended or default bond mode is Active backup

Solution:
Set the default mode to Active backup in harvester-installer when no bond options are specified by user.

Note:This is not applicable to harvester installed in manual mode as install panel defaults to "Active-Backup" and the value cannot be empty

This is related to https://github.com/harvester/harvester/issues/8042

### DIFF
--- a/pkg/config/cos.go
+++ b/pkg/config/cos.go
@@ -583,7 +583,7 @@ func updateBond(stage *yipSchema.Stage, name string, network *Network) error {
 	if network.BondOptions == nil {
 		logrus.Infof("Adding default NIC bonding options for \"%s\"", name)
 		network.BondOptions = map[string]string{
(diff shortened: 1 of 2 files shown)


ibrokethecloud (approved):
lgtm. thanks.

starbops (approved):
LGTM, thank you.

starbops:
@rrajendran17 It seems the change breaks the CI: https://github.com/harvester/harvester-installer/actions/runs/14787945073/job/41526517564. Might need to check if we need to adjust anything in the ipxe-examples repo.

PR #1033, opened 2025-04-17
feat: add rancher integration pipeline

FrankYang0529:
#### Related Issue(s):

https://github.com/harvester/harvester/issues/7889


.github/workflows/rancher-integration.yaml
```diff
@@ -0,0 +1,194 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Patch RKE2 Version
+        run: |
+          sed -i 's/^RKE2_VERSION=".*"/RKE2_VERSION="${{ github.event.inputs.rke2Version }}"/' ./scripts/version-rke2
+
+      - name: Patch Rancher Version
+        run: |
+          sed -i 's/^RANCHER_VERSION=".*"/RANCHER_VERSION="${{ github.event.inputs.rke2Version }}"/' ./scripts/version-rancher
```
innobead, line 34: Should be rancherVersion instead of rke2Version.

.github/workflows/rancher-integration.yaml
```diff
@@ -0,0 +1,194 @@
...
+          sed -i 's/^  rancherImageTag: .*/  rancherImageTag: ${{ github.event.inputs.rancherVersion }}/' ./package/harvester-os/files/usr/share/rancher/rancherd/config.yaml.d/50-defaults.yaml
+          sed -i 's|^rancherInstallerImage: .*|rancherInstallerImage: rancher/system-agent-installer-rancher:${{ github.event.inputs.rancherVersion }}|' ./package/harvester-os/files/usr/share/rancher/rancherd/config.yaml.d/50-defaults.yaml
+
+      - name: Make a commit
```
innobead, line 38: Do we need this step? Or if not commit the changed file, the following make will encounter issues? for example, getting version

.github/workflows/rancher-integration.yaml
```diff
@@ -0,0 +1,194 @@
...
+          ./setup_harvester.sh
+      - name: Enable soft emulation
+        working-directory: ./ci/terraform
+        run: |
+          ./enable_soft_emulation.sh ${{ steps.ipxe.outputs.VAGRANT_HOME }}/settings.yml
+      - name: Clean the previous temp files
+        working-directory: ./ci/terraform
+        run: |
+          ./cleanup_test_files.sh
+      - name: Testing existing files
```
innobead, line 95: `Test existing files`

The name is quite general, can we make them specific? What files should exist. `check_files.sh` is also too common.

innobead, line 95: `Check files exist on provisioned hosts`

nit: The name is quite general, can we make them specific? What files should exist. `check_files.sh` is also too common.

.github/workflows/rancher-integration.yaml
```diff
@@ -0,0 +1,194 @@
...
+        working-directory: ./ci/terraform
+        run: |
+          ./check_files.sh ${{ steps.ipxe.outputs.VAGRANT_HOME }}/settings.yml
+      - name: Testing services status
+        working-directory: ./ci/terraform
```
innobead, line 100: `Check system services status on provisioned hosts`

.github/workflows/rancher-integration.yaml
```diff
@@ -0,0 +1,194 @@
...
+        run: |
+          ./check_services_status.sh ${{ steps.ipxe.outputs.VAGRANT_HOME }}/settings.yml
+      - name: Testing basic operations with terraform
+        working-directory: ./ci/terraform
```
innobead, line 104: `Create Harvester resources for testing`

.github/workflows/rancher-integration.yaml
```diff
@@ -0,0 +1,194 @@
...
+        run: |
+          curl https://releases.hashicorp.com/terraform/1.3.7/terraform_1.3.7_linux_amd64.zip -o terraform_bin.zip
+          unzip -o terraform_bin.zip
+          ./get_kubeconfig.sh ${{ steps.ipxe.outputs.VAGRANT_HOME }}/settings.yml
+          ./terraform init -no-color
+          ./terraform apply -auto-approve -no-color
+      - name: Test network on the VMs
+        working-directory: ./ci/terraform
```
innobead, line 112: `Test VM networking`

.github/workflows/rancher-integration.yaml
```diff
@@ -0,0 +1,194 @@
...
+        run: |
+          ./test_terraform_vm.sh ${{ steps.ipxe.outputs.VAGRANT_HOME }}/settings.yml
+
+      - name: Check Rancher System Agent
+        run: |
```
innobead, line 117: Should we move this just after `Setup cluster`?

.github/workflows/rancher-integration.yaml
```diff
@@ -0,0 +1,194 @@
...
+          token=$(echo ${login_output} | jq -r '.token')
+          changepassword_output=curl -kX POST -H "Content-Type: application/json" -H "Authorization: Bearer ${token}" -d '{"currentPassword":"admin","newPassword":"testtesttest"}' \
+          https://${node0_ip}/v3/users?action=changepassword
+          if [ -n "${changepassword_output}" ]; then
+            echo "Change password result: ${changepassword_output}"
+            exit 1
+          fi
+
+      - name: Check rotate RKE2 certificates
+        run: |
```
innobead, line 145: `Test RKE2 cert rotation`

.github/workflows/rancher-upgrade-integration.yaml
```diff
@@ -0,0 +1,184 @@
...
+      - uses: actions/checkout@v4
+
+      - name: Patch RKE2 Version
+        run: |
+          sed -i 's/^RKE2_VERSION=".*"/RKE2_VERSION="${{ github.event.inputs.rke2Version }}"/' ./scripts/version-rke2
+
+      - name: Patch Rancher Version
+        run: |
+          sed -i 's/^RANCHER_VERSION=".*"/RANCHER_VERSION="${{ github.event.inputs.rke2Version }}"/' ./scripts/version-rancher
+          sed -i 's/^  rancherImageTag: .*/  rancherImageTag: ${{ github.event.inputs.rancherVersion }}/' ./package/harvester-os/files/usr/share/rancher/rancherd/config.yaml.d/50-defaults.yaml
```
innobead, line 38: `rancherVersion` instead of `rke2Version`?

.github/workflows/rancher-upgrade-integration.yaml
```diff
@@ -0,0 +1,184 @@
...
+        working-directory: ./ci/terraform
+        run: |
+          ./check_files.sh ${{ steps.ipxe.outputs.VAGRANT_HOME }}/settings.yml
+      - name: Testing services status
+        working-directory: ./ci/terraform
+        run: |
+          ./check_services_status.sh ${{ steps.ipxe.outputs.VAGRANT_HOME }}/settings.yml
+
+      - name: Create new-version.yaml
+        run: |
```
innobead, line 115: Suggest creating a script for this upgrade to simplify the steps here.

.github/workflows/rancher-integration.yaml
FrankYang0529: I would like to make sure there is no "-dirty" in the version name.

https://github.com/harvester/harvester-installer/blob/6973fcdfca330bb1920710f2620b6e71a8496e76/scripts/version#L3-L14

innobead (changes requested):
Can we make a reusable workflow which can be used in both workflow to reduce duplicated implementation here? Basically, have a job running the reusbale workflow, then start the different steps in another job.

.github/workflows/rancher-integration.yaml
```diff
@@ -0,0 +1,194 @@
...
+          cd ipxe-examples
+          git reset && git checkout .
+          git clean -fd
+          git pull
+          echo "VAGRANT_HOME=$HOME/ipxe-examples/vagrant-pxe-harvester" >> $GITHUB_OUTPUT
+      - name: Clean up previous vagrant deployment
+        working-directory: ${{ steps.ipxe.outputs.VAGRANT_HOME }}
+        run: |
+          vagrant destroy -f
+      - name: Remove OVMF.fd line if needed
```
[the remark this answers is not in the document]
FrankYang0529: I copy this from https://github.com/harvester/harvester-installer/blob/6973fcdfca330bb1920710f2620b6e71a8496e76/.github/workflows/vagrant-install.yaml#L47-L53. @bk201 do you know why we need this in CI? Does it avoid UEFI firmware not found error?

innobead:
The ultimate goal is to automatically detect the versions of Rancher and RKE2, and then run these workflows.

There are a few things we need to improve later.
- Create a matrix strategy outlining which versions to run in the workflow.  For instance: Harvester 1.5/Rancher 2.11/RKE2 1.32, we only need to execute pipelines for supported branches and the head. Currently, these are head/1.6, 1.5, 1.4, and 1.3. To conserve resources, we will only run head, n, and n-1, which means head, 1.5, and 1.4.
- Schedule a daily cron job to check for any new releases (just released on the day) of Rancher or RKE2, and run the testing pipeline as needed. You may require an additional step/job initially to determine them. If no new releases are available, there's no need to execute anything. 
- Ensure the capability to run manually by entering upstream versions.

.github/workflows/rancher-integration.yaml
innobead: This is what I assumed due to version.

.github/workflows/rancher-integration.yaml
FrankYang0529: Since check_files.sh is used in other workflow, I will update the naming in the next PR.

.github/workflows/rancher-integration.yaml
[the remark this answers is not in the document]
innobead: If some snippets are resuable, suggest creating util scripts or workflows, then share them with each other. In addition, please also add comments to clarify the purpose, or prefer making the step name self-explantory. It will be good for understanding.

.github/workflows/rancher-integration.yaml
FrankYang0529: I prefer to keep it after terraform check. If it checks rancher-system-agent just after setup cluster, the cluster may not be ready.

innobead: I see, then it can be part of tests.

ci/upgrade/create_upgrade_file.sh
starbops, line 1: The name of the script does not match what it does. Same for `create_version_file.sh`.

FrankYang0529: Thanks. Updated it.

.github/workflows/rancher-upgrade-integration.yaml
```diff
@@ -0,0 +1,170 @@
...
+          ./check_files.sh ${{ steps.ipxe.outputs.VAGRANT_HOME }}/settings.yml
+      - name: Check system services status on provisioned hosts
+        working-directory: ./ci/terraform
+        run: |
+          ./check_services_status.sh ${{ steps.ipxe.outputs.VAGRANT_HOME }}/settings.yml
+
+      - name: Create new-version.yaml
+        working-directory: ./ci/upgrade
+        run: |
+          ./create_upgrade_file.sh
```
innobead, line 117: This should be create_version_file.sh instead? 

BTW, can we simplify the upgrade operation as follows? When creating a new version or upgrade, apply it directly.

```yaml
- name: Start file server
  run: |
    python3 -m http.server 8000 --bind 0.0.0.0 &

- name: Create the upgrade version
  working-directory: ./ci/upgrade
  run: |
    ./create_version.sh

- name: Create the upgrade
  working-directory: ./ci/upgrade
  run: |
    ./create_upgrade.sh

```

.github/workflows/rancher-integration.yaml
```diff
@@ -0,0 +1,194 @@
...
+          ./terraform init -no-color
+          ./terraform apply -auto-approve -no-color
+      - name: Test VM networking
+        working-directory: ./ci/terraform
+        run: |
+          ./test_terraform_vm.sh ${{ steps.ipxe.outputs.VAGRANT_HOME }}/settings.yml
+
+      - name: Check Rancher System Agent
+        run: |
+          for i in 0 1 2; do
```
innobead, line 118: Can we move this to check_rancher_system_agent.sh? I would like to move all checks and tests under the terraform folder to make them consistent as the existing test scripts like test_terraform_vm.sh.

.github/workflows/rancher-integration.yaml
```diff
@@ -0,0 +1,194 @@
...
+              echo "Log check passed for $node_name"
+            else
+              echo "Log check failed for $node_name"
+              exit 1
+            fi
+          done
+
+      - name: Test Login
+        run: |
+          node0_ip=$(yq e ".harvester_network_config.cluster[0].ip" ${{ steps.ipxe.outputs.VAGRANT_HOME }}/settings.yml)
```
innobead, line 132: ditto, move this to check_login.sh

.github/workflows/rancher-integration.yaml
```diff
@@ -0,0 +1,194 @@
...
+          echo "login result: ${login_output}"
+          token=$(echo ${login_output} | jq -r '.token')
+          changepassword_output=curl -kX POST -H "Content-Type: application/json" -H "Authorization: Bearer ${token}" -d '{"currentPassword":"admin","newPassword":"testtesttest"}' \
+          https://${node0_ip}/v3/users?action=changepassword
+          if [ -n "${changepassword_output}" ]; then
+            echo "Change password result: ${changepassword_output}"
+            exit 1
+          fi
+
+      - name: Test RKE2 cert rotation
```
innobead, line 144: check_rke2_cert_rotation.sh

innobead (changes requested):
Just a few minor feedback. After resolving them, we can merge this first, so you can verify and fine-tune this later.

PR #1035, opened 2025-04-28
Fix cni pod network cache reset

w13915984028:
#### Problem:

`/var/lib/cni/networks/k8s-pod-network/` is not fully cleandup when node/cluster reboots.

#### Solution:

Fix stage file definition.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/8150

#### Test plan:

1. Restart node/cluser
2. Check node/nodes, the `/var/lib/cni/networks/k8s-pod-network` path is not existing or only has files with timestamp after node/cluster rebooting.

It is validated on both v1.5.0 and v1.4.2 cluster.

### DIFF
--- a/package/harvester-os/files/system/oem/99_cni_reset.yaml
+++ b/package/harvester-os/files/system/oem/99_cni_reset.yaml
@@ -1,5 +1,6 @@
 name: "reset container dhcp leases"
 stages:
-   initramfs:
-     - command:
-       - rm -rf /var/lib/cni/networks/k8s-pod-network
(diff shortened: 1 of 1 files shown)


w13915984028:
@mergify backport v1.5 v1.4

ibrokethecloud (approved):

Vicente-Cheng (approved):
lgtm!

Vicente-Cheng:
I noticed the CI failures. Triggered it to re-run.

PR #1056, opened 2025-06-18
Update grafana nginx proxy to sync with Rancher

w13915984028:
Rancher names the imported cluster from `c-m-*` to `c-*`

#### Problem:

After bumping Rancher to v2.11.2, the Rancher->Harvester can't load grafana

#### Solution:

Update the grafana nginx proxy setting to filter the imported cluster name.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/8498

Upgrade path processing will be added to `harvester` repo from another PR.


w13915984028:
@mergify backport to v1.5

innobead (approved):
LGTM. 

Have you verified the older Rancher versions still work well with this change?

PR #1058, opened 2025-06-23
Skip bridge vlan add when vid less than 2

rrajendran17:
Problem:
Adding vid 1 to bridge vlan will add vid=1 as tagged vlan to mgmt interface.
This will replace the default PVID=1 behaviour on mgmt interface.

#### Solution:
When user does not configure any vlan id or configured vlan id as 1,skip configuring vid=0 or vid=1 on wicked scripts, so that the default PVID=1 is retained on the mgmt interface.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8515

#### Test plan:
1.Install Harvester manually without any vlan id or vlan id=1
2.Check if harvester nodes are setup correctly and ready.


w13915984028 (approved):
LGTM, thanks.

starbops (approved):

PR #1072, opened 2025-07-01
feat: reinstate rsync progress in harv-install

tserong:
#### Problem:
The rsync `--progress` flag was removed in https://github.com/harvester/harvester-installer/pull/641 in order to avoid locking up when installing via extremely slow virtual media.  The original problem was that we were waiting for a '\n' character, and rsync's stdout buffer eventually got full of progress output before the '\n' was printed and the whole thing locked up.
    
Since https://github.com/harvester/harvester-installer/pull/882 was added in Harvester v1.5.0 to handle '\r' characters in command output, it's now safe to re-instate the --progress flag when running rsync, because we now split on both '\r' and '\n'.

#### Solution:
- Reinstate the `--progress` flag in rsync calls
- Search for '\r' before '\n' when splitting lines to avoid odd extra '\r' characters in the last line of rsync output

#### Related Issue(s):

https://github.com/harvester/harvester/issues/5095

#### Test plan:
- Install harvester from the ISO
- Watch the console during installation and ensure that you see multiple lines of rsync progress output printed in real time (one per second or so) _while_ the rsync is occurring:

https://github.com/user-attachments/assets/2a378d8b-9678-4db6-8d52-865d3471a8b7

### DIFF
--- a/package/harvester-os/files/usr/sbin/harv-install
+++ b/package/harvester-os/files/usr/sbin/harv-install
@@ -188,10 +188,10 @@ preload_rke2_images()
     # Otherwise (PXE), use bind-mount directly to speed up the process
     if [ "$INSTALL_MODE" = "ISO" ]; then
         echo "Copying RKE2 images to the target location..."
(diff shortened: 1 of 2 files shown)


starbops (approved):
LGTM, thanks for the improvement!

PR #1075, opened 2025-07-04
feat: don't require a data disk for witness nodes

tserong:
#### Problem:

 The witness node only runs a minimal workload, so we don't need a data disk/partition.

#### Solution:

When adding a witness node:
- The data disk and persistent size fields are removed from the interactive installer.
- No data partition will be created on the installation disk.
- Minimum installation disk size becomes 180GiB (the same as is used for other nodes when there's a separate data disk).

#### Related Issue(s):

https://github.com/harvester/harvester/issues/5457

#### Test plan:
TBD

This PR includes a tiny bit of unrelated code reformatting and a lint error fix. If you want to ignore those when reviewing, view commit by commit.

I still need to build an ISO and do a test install to make sure this all works properly. Stay tuned...

### DIFF
--- a/pkg/console/install_panels.go
+++ b/pkg/console/install_panels.go
@@ -1352,50 +1352,50 @@ func addHostnamePanel(c *Console) error {
 // If s is not a valid textual representation of an IP mask,
 // ParseMask returns an error describing why the parsing failed.
 func ParseMask(mask string) (IPMask, error) {
-    // Split the mask string by dots
-    parts := strings.Split(mask, ".")
+	// Split the mask string by dots
+	parts := strings.Split(mask, ".")
 
-    // Validate number of parts
-    if len(parts) != 4 {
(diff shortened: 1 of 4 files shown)


pkg/console/install_panels.go
```diff
@@ -521,7 +532,34 @@ func addDiskPanel(c *Console) error {
 		return showNetworkPage(c)
 	}
 
-	diskConfirm := func(_ *gocui.Gui, _ *gocui.View) error {
+	// isWipeDisksNeeded is a helper function to render the wipeDisksPanel if needed
+	// if there are no additional disks to wipe then it checks if MBR needs to be enabled
+	// else will move on to the next apge
+	isWipeDisksNeeded := func(g *gocui.Gui, v *gocui.View) error {
```
Vicente-Cheng, line 538: could we rename "isWipeDisksNeeded" to "isWipeDisksPanelNeeded"?

Vicente-Cheng (approved):
just a nit, lgtm!

PR #1077, opened 2025-07-06
Trivial: Fix code style: cfg.Mode -> cfg.Install.Mode

jjqq2013:
#### Problem:

Not a problem, just a trivial code style confusion.

There is a place using
```
cfg.Mode
```
which actually means
```
cfg.Install.Mode
```
All other places are using `cfg.Install.Mode`.

I know that Golang correctly treat `cfg.Mode` as `cfg.Install.Mode` (struct embedding).

But this caused confusion for Golang newbies, especially when using other pure text tools to analyse the usage of `Mode`.

It would be better if unify it to the formal expression `cfg.Install.Mode`, so other tools without golang support can easily grep and analyse the usage of the Install.Mode.

#### Solution:

Unify `cfg.Mode` to `cfg.Install.Mode`.

### DIFF
--- a/pkg/config/cos.go
+++ b/pkg/config/cos.go
@@ -186,7 +186,7 @@ func ConvertToCOS(config *HarvesterConfig) (*yipSchema.YipConfig, error) {
 	disableLonghornMultipathing(&initramfs)
 
 	// TOP
-	if cfg.Mode != ModeInstall {
+	if cfg.Install.Mode != ModeInstall {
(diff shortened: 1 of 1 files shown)


Vicente-Cheng (commented):
I am OK with that, which would make things more friendly for the one who first checks the installer code.

IIUC, we have other similar cases. Could you help with them all?
Thanks!

ibrokethecloud (approved):

PR #1081, opened 2025-07-09
Remove `harvesterInstalled` function

dharmit:
#### Problem:

https://github.com/harvester/harvester/issues/8620

#### Solution:

Remove the function returning hardcoded values.
#### Related Issue(s):

https://github.com/harvester/harvester/issues/8620
#### Test plan:

* CI should pass, I guess.
* Running `make` on my dev setup is taking obnoxiously long. I'll update here if/when it succeeds.

### DIFF
--- a/pkg/console/install_panels.go
+++ b/pkg/console/install_panels.go
@@ -864,15 +864,6 @@ func addAskCreatePanel(c *Console) error {
 				Text:  "Join an existing Harvester cluster",
 			},
 		}
-		installed, err := harvesterInstalled()
-		if err != nil {
-			logrus.Error(err)
-		} else if installed {
-			options = append(options, widgets.Option{
-				Value: config.ModeUpgrade,
-				Text:  "Upgrade Harvester",
-			})
-		}
 
 		// layoutInstall is now called from layoutDashboard due to the addition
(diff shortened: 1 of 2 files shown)


pkg/console/install_panels.go
```diff
@@ -863,15 +863,10 @@ func addAskCreatePanel(c *Console) error {
...
 			},
-		}
-		installed, err := harvesterInstalled()
-		if err != nil {
-			logrus.Error(err)
-		} else if installed {
-			options = append(options, widgets.Option{
+			{
 				Value: config.ModeUpgrade,
 				Text:  "Upgrade Harvester",
```
tserong, line 868: This value and text should be deleted too, as we don't want there to be an "Upgrade Harvester" item in this menu - the `harvesterInstalled()` function previously always returned `false`, so the `if installed` condition above would always be `false`, meaning this "Upgrade Harvester" item was never added before.

tserong (changes requested):
Thanks! Just one little tweak... :-)

innobead (approved):

tserong (approved):

dharmit:
> so I'm wondering if I should "squash and merge" (I do see an option to do that) this now that there are two approvals.

Taking cue from some old PRs on this repo, going ahead with merging the PR.

PR #1092, opened 2025-07-22
Introduce toolchain directive to specify go version

votdev:
... to prevent errors like
```
$ go: downloading go1.23 (linux/amd64)
go: download go1.23 for linux/amd64: toolchain not available
```
Latest version number can be found here: https://go.dev/dl/


votdev:
Closing in favor of https://github.com/harvester/harvester-installer/pull/1097

PR #1093, opened 2025-07-22
Optimize the managedchart Harvester installed status check

w13915984028:
#### Problem:

https://github.com/harvester/harvester/issues/8655

#### Related Issue(s):

https://github.com/harvester/harvester/issues/8655

#### Test plan:

1. Setup a single-node Harvester
2. When console is ready,  change harvester deployment to a new image or new memory limit, the managedchart/bundle will have complain soon
3.1 Reboot the cluster, console does not stuck on https://github.com/harvester/harvester/issues/8655#issue-3228172883
3.2 For a running cluster, if it runs into the state, run `kubectl annotate managedchart -n fleet-local harvester 'installer.harvesterhci.io/chartInstalled=true'`, the console turns into ready again

### DIFF
--- a/pkg/console/dashboard_panels.go
+++ b/pkg/console/dashboard_panels.go
@@ -453,7 +453,33 @@ func k8sIsReady() bool {
 	return true
 }
 
+func chartIsAnnotated() bool {
+	cmd := exec.Command("/bin/sh", "-c", `kubectl get managedchart -n fleet-local harvester -ojsonpath="{.metadata.annotations['installer\.harvesterhci\.io\/chartInstalled']}"`)
+	cmd.Env = os.Environ()
+	output, err := cmd.Output()
+
+	if err != nil {
+		// any return error means false
+		return false
+	}
+	return string(output) != ""
+}
+
+func annotateChart() {
+	cmd := exec.Command("/bin/sh", "-c", `kubectl annotate managedchart -n fleet-local harvester 'installer.harvesterhci.io/chartInstalled=true'`)
+	cmd.Env = os.Environ()
+	// skip output
+	_, _ = cmd.Output()
+}
+
 func chartIsInstalled() bool {
+	// after the managedchart installation step in done, it turns into `Ready`
+	// later if the user changes the downstream object like deployment, the managedchart might become `non-Ready`
+	// however, the `non-Ready` does not mean the chart is not `installed`
+	// use an annotation to mark the `installed` of managedchart
+	if chartIsAnnotated() {
+		return true
+	}
 	cmd := exec.Command("/bin/sh", "-c", `kubectl -n fleet-local get ManagedChart harvester -o jsonpath='{.status.conditions}' | jq 'map(select(.type == "Ready" and .status == "True")) | length'`)
 	cmd.Env = os.Environ()
 	output, err := cmd.Output()
@@ -471,7 +497,12 @@ func chartIsInstalled() bool {
 		return false
 	}
 
-	return processed >= 1
+	if processed >= 1 {
+		// installer annotates the managedchart when the installation step in done
+		annotateChart()
+		return true
+	}
+	return false
 }
 
 func isAPIReady(managementURL, path string) bool {


ihcsim:
Is it necessary to check the managed chart? I thought we had a similar previous discussion on `isAPIReady()`, where it can be used to confirm readiness or kube-vip, nginx-ingress, rancher, harvester etc.

w13915984028:
> If the chart installation failed on first boot, won't the API check fail? If yes, checking the managed chart sounds redundant.

Agree.

Check the managedchart status on first node is mainly for showing `Setting up Harvester` on day-0 installation, it takes time to be ready. A new user might feel a bit comfort than just seeing `Not Ready`. Or some other corner cases not documented.

Meanwhile, keeping it also does not harm too much.

ihcsim:
If it's just cosmetic, then just replace this hard-coded line from `loading` with `Setting up Harvester` :smile:: https://github.com/harvester/harvester-installer/blob/5a09f65d161fbe7598baa98c9ecb68654de66939/pkg/console/dashboard_panels.go#L111

And update it only after API is ready. Then we won't need these additional code to assess the managed charts states.

w13915984028:
@ihcsim BTW, my original plan is to optimize this via https://github.com/harvester/harvester/issues/6719, the Harvester controller makes decision and the installer is only responsible for displaying the information.

w13915984028:
Before we have a better solution, I would rather keep the `Setting up Harvester` wording in this RP, users have been familiar with it, change to another words might cause unnecessary concerns. thanks.

ihcsim:
I think https://github.com/harvester/harvester/issues/6719 is a good one. Meanwhile, I don't this PR is necessary and we end up with stopgap code to maintain. If user changes resources spec directly, instead of changing the managed chart via fleet, then they put the system in a not `Ready` state.

Vicente-Cheng:
Does it mean that once we successfully install the managedchart, it will always show ready?

The upgrade webhook also checks the managedchart. If we show ready, but the upgrade gets stuck. Does it make sense?
To me, the console represents the current status, rather than the installation status.

Do we have any document that mentions what is called `Ready`?

w13915984028:
> Does it mean that once we successfully install the managedchart, it will always show ready?
> 
> The upgrade webhook also checks the managedchart. If we show ready, but the upgrade gets stuck. Does it make sense? To me, the console represents the current status, rather than the installation status.
> 
> Do we have any document that mentions what is called `Ready`?


Some additional details:

(1) Harvester installer on the `initial node` runs below code to get a flag `current.installed`, if it is false, then it shows `Setting up Harvester`, otherwise, go to next check.

Note: The `installed` is a single-way flag, it won't change back to none-ready unless the `installer` process is restarted or node is rebooted.

```
func getHarvesterStatus() string {
	if current.firstHost && !current.installed {
		if !k8sIsReady() || !chartIsInstalled() {
			return statusSettingUpHarv
		}
		current.installed = true
	}
```

https://github.com/harvester/harvester-installer/blob/80e328e00d5f6beea6d032de34e007d5b6257356/pkg/console/dashboard_panels.go#L524

(2) In which case it stucks on `Setting up Harvester`?

a. The node is rebooted/the `harvester-installer` process is restarted
b. The managedchart has those complains like:  network controller deployment is directly updated with a new memory limit;  the second control-planed node is condored so that `harvester` and other are marked as `has no minimal...`;   before those complains are gone, it always show such

(3) In which case it does not show `Setting up Harvester`?

When `current.installed` is ever set to `true`, it is always true. Even when later there are complains, it is still `Ready`.

Regarding the question `The upgrade webhook also checks the managedchart`:  upgrade script check managedchart separately, it does not influence/is not influced by `harvester-installer` directly.

It is easy to produce: Harvester upgrade denies the upgrade due to `managedchart` is not ready; but the console just shows info `Ready`.


(4) Short summary:

(a) Console `setting up Harvester` is a specific logic running on `first node`
(b) The `current.installed` is a `single-way` bool flag, the only change is `false` to `true`
(c) It does not reflect the real running status of the cluster
(d) It does not affect upgrade webhook, does not reflect upgrade change (e.g. the managedchart is stucking upgrading)
(e) After day-0, the `Setting up Harvester` in most cases is kind of false alarm

PR #1100, opened 2025-07-25
fix: don't show persistentSizeNote panel for witness nodes

tserong:
#### Problem:
Showing the persistent size note doesn't make sense for witness nodes, because we don't let you set the persistent size in this case.  Also it causes the installer to lock up and not take keypresses (I assume that note field somehow ends up with the focus which messes things up).

#### Solution:
Don't show the persistent size note when installing witness nodes.

#### Related Issue(s):

Related issue: https://github.com/harvester/harvester/issues/5457

#### Test plan:
Install a witness node with only one disk. The disk selection screen should not lock up like it did in Test 3 mentioned in https://github.com/harvester/harvester/issues/5457#issuecomment-3112973243

### DIFF
--- a/pkg/console/install_panels.go
+++ b/pkg/console/install_panels.go
@@ -588,14 +588,24 @@ func addDiskPanel(c *Console) error {
 			return showNext(c, dataDiskPanel)
 		}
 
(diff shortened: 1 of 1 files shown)


ihcsim (approved):

votdev (approved):
LGTM

PR #1102, opened 2025-07-25
fix: build: don't rm -rf addons source

tserong:
#### Problem:
If you run make from the harvester-installer repo, and have `LOCAL_ADDONS_SRC` set to a local checkout of the addons repo (to save cloning it at build time), your local checkout of the addons repo will be deleted.

#### Solution:
Don't rm -rf the addons source

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8763

#### Test plan:
Build an ISO

### DIFF
--- a/scripts/package-harvester-os
+++ b/scripts/package-harvester-os
@@ -22,7 +22,6 @@ LOGGING_VERSION=${RANCHER_LOGGING_CHART_VERSION}
 echo "Rancher monitoring version: ${MONITORING_VERSION}"
 echo "Rancher logging version: ${LOGGING_VERSION}"
 echo "Harvester eventrouter image tag: ${HARVESTER_EVENTROUTER_FULL_TAG}"
-rm -rf ${addons_path}
 
 mkdir -p ${ARTIFACTS_DIR}
 


scripts/package-harvester-os
```diff
@@ -22,7 +22,6 @@ LOGGING_VERSION=${RANCHER_LOGGING_CHART_VERSION}
 echo "Rancher monitoring version: ${MONITORING_VERSION}"
 echo "Rancher logging version: ${LOGGING_VERSION}"
 echo "Harvester eventrouter image tag: ${HARVESTER_EVENTROUTER_FULL_TAG}"
-rm -rf ${addons_path}
```
w13915984028, line 25: BTW: use env `LOCAL_ADDONS_SRC` to determin if rm or not ? thanks.

tserong: I don't think `LOCAL_ADDONS_SRC` is actually passed through to the build environment right now. Why do this? Is it important that the addons source be removed from the build environment at this point?

w13915984028: no speicial purpose, just avoid remaining file affecting others;  skip `rm` is ok

tserong: cool, thanks!

PR #1133, opened 2025-09-05
Enable longhorn storage reserved percentage conf

martindekov:
Enabling storage-reserved-percentage-for-default-disk configuration
as part of the install.harvester.longhorn.defaultSettings

The following behaviour is enforced:
* values can range between 0% and 30%
* default value if no config passed is 0%
* max value is 30%

#### Problem:
Longhorn setting: `storage-reserved-percentage-for-default-disk` is not exposed in the harvester installation and can only be changed manually after installation.

#### Solution:
Expose the setting through configuration file

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8200

#### Test plan:
Create local configuration file containing the setting (the file can be only that):
```yaml
install:
  harvester:
    longhorn:
      defaultSettings:
        storageReservedPercentageForDefaultDisk: 50
```
Served it locally through python to the harvester installation:
```
python3 -m http.server --bind 0.0.0.0 8000
```

Once ready ssh into the node and validate configuration was inside (as per description 30 not 50 since 30 is max):
```
harvester:/home/rancher # kubectl get settings.longhorn.io -n longhorn-system | grep storage-reserved-percentage-for-default-disk
storage-reserved-percentage-for-default-disk                      30                                                              true      24m
```

Default value with no configuration passed needs to be the expected one:
```
harvester:/home/rancher # kubectl get settings.longhorn.io -n longhorn-system | grep storage-reserved-percentage-for-default-disk
storage-reserved-percentage-for-default-disk                      0                                                               true      3m2s
```

So testing scenarios:
* set value above max limit of 30 - like 100, expect in env to set to 30
* install without configuration - expect value 0 - the default
* set value of a negative number - would something panic out?
* set value to custom number between 0 and 30, like 15 - expect the number to be reflected in the env

N/A

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -123,6 +123,8 @@ type LHDefaultSettings struct {
 	// from Longhorn v1.5.0, LH merges the above two into one
 	// the above two are not used afterwards, but Harvester keeps them for compatibility
 	GuaranteedInstanceManagerCPU *uint32 `json:"guaranteedInstanceManagerCPU,omitempty"`
+
+	StorageReservedPercentageForDefaultDisk *uint32 `json:"storageReservedPercentageForDefaultDisk,omitempty"`
 }
 
 type LonghornChartValues struct {
--- a/pkg/config/cos.go
+++ b/pkg/config/cos.go
@@ -28,12 +28,15 @@ const (
 	ifcfgGlobPattern       = networkConfigDirectory + "ifcfg-*"
 	ifrouteGlobPattern     = networkConfigDirectory + "ifroute-*"
 
-	bootstrapConfigCount                = 6
-	defaultReplicaCount                 = 3
-	defaultGuaranteedEngineManagerCPU   = 12   // means percentage 12%
-	defaultGuaranteedReplicaManagerCPU  = 12   // means percentage 12%
-	defaultGuaranteedInstanceManagerCPU = 12   // means percentage 12%
-	defaultSystemImageSize              = 3072 // size of /run/initramfs/cos-state/cOS/active.img in MB
+	bootstrapConfigCount                           = 6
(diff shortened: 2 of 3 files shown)


w13915984028 (approved):
LGTM, thanks.

w13915984028:
@Vicente-Cheng 
For the independent panel, need to also consider:
(1) There are some installer configuration options which are only available on configuration file, not all are exposed to panel; those  mandatory &/ frequently used parameters are exposed in general.
(2) The default value 0 of `storage-reserved-percentage-for-default-disk` is used in most cases unless user plans to set a specific value.
(3) From past experiences, adding new pannel has higher chance to import/trigger bugs.

Back to LH `storage-reserved-percentage-for-default-disk`, it seems not necessary to add a dedicated panel. And it is also possible that in the future, Harvester supports other CSI driver as the default one,

martindekov:
This was the reason for my suggestion of opening a separate issue where this can be discussed I was not aware of a lot of details - e.g. Longhorn configuration panel or Storage configuration panel? Generic config to storage or specific to longhorn?

If we intend to have other storage drivers then the panel would need to expose only configuration which intersects with configuration of other drivers as criteria e.g. all drivers usually have replica manager so put the generic configuration up in the panel. I jump started implementing panel for longhorn, but as per @w13915984028's input it made me think.

In that case would going back to separate issue, where all the details can be discussed, be a good idea?

Vicente-Cheng:
If we change the default value to 0%, and yes, I thought you were correct, that might be the rarely changed parameters.

But creating an extra setting panel for Longhorn would not be a bad idea. I will create another issue to track it if anyone would like to have it.
For this PR, the current implementation would be enough.
Thanks

PR #1132, opened 2025-09-04
fix: check for number of disks during preflight

wheatdog:
#### Problem:

When running the Harvester installer on a host without an accessible disk, the installer will terminate with panic.

#### Solution:

Check number of disks during preflight

#### Related Issue(s):

https://github.com/harvester/harvester/issues/7863

#### Test plan:

Install on a machine without an accessible disk,
- Path 1: Shall report "No disk detected" in the first page during interactive installation
```mermaid
flowchart TD
    A[Begin] --> B1{"len(preflightFatals) > 0"};
    B1 -- "true" --> C["preflightFatalPanel"];
    B1 -- "false" --> B2{"len(preflightWarnings) > 0"};
    B2 -- "true" --> D1["preflightCheckPanel"];
    B2 -- "false" --> E["askCreatePanel: Continue installation"]; 
    
    C -- "Acknowledged" --> F[Reboot];
    D1 --> D2{"Do you wish to proceed?"};
    
    D2 -- "Yes" --> E;
    D2 -- "No" --> F;
```

preflightFatalPanel:

image

- ~~Path 2: Shall abort with "No disk detected" during PXE installation~~

#### Additional documentation or context

Move disk related functions from `pkg/console/util.go` to `pkg/util/disk.go` such that they can be reused in `pkg/preflight/check.go`.

Currently, following error messages can be observed when trying PXE installation without disk. I think we can keep this behavior and this PR would not affect it.

<details>
<summary> /run/initramfs/rdsosreport.txt </summary>

```
+ cat /lib/dracut/dracut-055+suse.396.g701c6212-150500.3.29.2
dracut-055+suse.396.g701c6212-150500.3.29.2
+ echo /proc/cmdline
/proc/cmdline
+ sed -e 's/\(ftp:\/\/.*\):.*@/\1:*******@/g;s/\(cifs:\/\/.*\):.*@/\1:*******@/g;s/cifspass=[^ ]*/cifspass=*******/g;s/iscsi:.*@/iscsi:******@/g;s/rd.iscsi.password=[^ ]*/rd.iscsi.password=******/g;s/rd.iscsi.in.password=[^ ]*/rd.iscsi.in.password=******/g' /proc/cmdline
[...]

PR #1138, opened 2025-09-10
fix: check number of disks on disk page and report error gracefully when there is none

wheatdog:
#### Problem:

When running the Harvester installer on a host without an accessible disk, the installer will terminate with panic.

#### Solution:

Check number of disks on disk page

#### Related Issue(s):

https://github.com/harvester/harvester/issues/7863

#### Test plan:

Path 1: Install on a machine without a disk

https://github.com/user-attachments/assets/9c0256dc-fd3e-49bc-a623-d76fbf3723c0

Path 2: Hotplug: Install on a machine without a disk and attach one later

https://github.com/user-attachments/assets/433e94ae-4d4a-4bab-b9c2-f47d46a70153

Path 3: Hotplug: Install on a machine with a disk and detach it later

https://github.com/user-attachments/assets/1a349ef9-e2d0-4a56-a100-fdb7071e5cdd

https://github.com/harvester/harvester-installer/pull/1132 is another way to tackle the same problem.

### DIFF
--- a/pkg/console/constant.go
+++ b/pkg/console/constant.go
@@ -4,6 +4,7 @@ const (
 	titlePanel                  = "title"
 	debugPanel                  = "debug"
 	diskPanel                   = "disk"
+	diskFatalPanel              = "diskFatal"
 	persistentSizePanel         = "persistentSize"
 	dataPersistentSizePanel     = "dataPersistentSize"
 	dataPersistentSizeNotePanel = "dataPersistentSizeNote"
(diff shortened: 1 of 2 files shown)


w13915984028 (commented):
some questions, thanks.

w13915984028 (dismissed):
LGTM, thx.

Vicente-Cheng (dismissed):
lgtm, thanks for the enhancement!

Vicente-Cheng:
Hi @wheatdog, 
Please rebase and squash.

Thanks!

PR #1139, opened 2025-09-11
Local config support

clanktron:
#### Problem:
Currently only remote http urls are allowed to be used for sourcing a harvester config file in the installation process. Sourcing a config file from the local filesystem instead can be useful in many cases.

#### Solution:
By simply adding support for the `file://` url prefix in the `getURL()` function, one can source from the local system without any invasive changes to the codebase.

#### Related Issue(s):
Issue harvester/harvester#9124

#### Testing:
A relevant unit test is bundled with this PR. Proper e2e functionality was also manually checked on multiple systems.

### DIFF
--- a/pkg/console/install_panels.go
+++ b/pkg/console/install_panels.go
@@ -2344,14 +2344,14 @@ func addProxyPanel(c *Console) error {
 }
 
 func addCloudInitPanel(c *Console) error {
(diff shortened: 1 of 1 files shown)


Vicente-Cheng:
Let's clarify the user scenario first, then we can do the PR review

PR #1141, opened 2025-09-15
[WIP, DNM] Use NetworkManager instead of Wicked

tserong:
#### Problem:
Need to migrate from Wicked to NetworkManager

#### Solution:
This PR and others (TBD: describe in more detail)

#### Related Issue(s):

#### Test plan:
TBD

Still WIP, opening PR now because I want to see what our automated testing does with it.


.github/workflows/build.yml
```diff
@@ -22,7 +22,6 @@ jobs:
       matrix:
         arch:
         - amd64
-        - arm64
```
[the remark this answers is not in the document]
tserong: It's just temporary so I could get a build for testing, I won't commit that.

pkg/console/network.go
```diff
@@ -107,6 +126,19 @@ func applyNetworks(network config.Network, hostname string) ([]byte, error) {
...
 	}
+
+	if hostname == "" && previousHostname != "" {
+		currentHostname, err := getHostname()
+		if err == nil && currentHostname == "localhost" {
+			// Restore the system hostname to what it was before, provided
+			// we haven't got a new hostname from the DHCP server.
+			err = setHostname(previousHostname)
+		}
+		if err != nil {
```
ibrokethecloud, line 137: if there was an error in getting `currentHostname` the message returned would get confusing

not sure if we should handle the `setHostname` call error in the inner code block?

tserong: Good point, I'll shuffle that around.

tserong: ...and done.

PR #1150, opened 2025-09-24
feat: add generate-network-yaml subcommand

tserong:
#### Problem:
We need to generate a YAML file in /oem containing NetworkManager connection profiles during upgrade.

#### Solution:
Add a subcommand to the harvester-installer binary which can generate the relevant YAML, then publish that binary in a container so it  can be pulled into and used by Harvester's upgrade container.

As implemented here, the subcommand to generate `/oem/91_networkmanager.yaml` from `/oem/harvester.config` looks like this:

```
harvester-installer generate-network-yaml \
    --config /oem/harvester.config \
    --cloud-init /oem/91_networkmanager.yaml
```

#### Related Issue(s):

#### Test plan:
TBD

The harvester-installer container publishing relies on https://github.com/rancherlabs/eio/issues/3619

This PR doesn't handle old-style (Harvester <= v1.1) harvester.confg format. I'd like to get this in as-is sooner rather than later, and address the old format separately if everyone's cool with that.

We need to get this in before https://github.com/harvester/harvester/pull/9200 can be completed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -80,6 +80,19 @@ jobs:
         sbom: true
         provenance: mode=max
 
+    # rancher/harvester-installer image
+    - name: docker-publish-harvester-installer
+      if: ${{ startsWith(github.ref, 'refs/heads/') }}
+      uses: docker/build-push-action@v6
+      with:
+        context: dist/harvester-installer
+        push: true
+        platforms: linux/${{ matrix.arch }}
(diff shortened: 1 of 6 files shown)


ibrokethecloud (approved):
lgtm. thanks for the enhancement.

Vicente-Cheng (approved):
lgtm, verified the command works well!

PR #1159, opened 2025-10-03
fix: Don't overwrite /oem/91_networkmanager.yaml if it already exists

tserong:
#### Problem:
It's too easy to run `harvester-installer generate-network-yaml` and overwrite possible manual changes to `/oem/91_networkmanager.yaml`

#### Solution:
By default, don't overwrite the output file if it already exists. Add a `--force` option to override if necessary.

#### Related Issue(s):

#### Test plan:
- Upgrade from v1.6.0 to v1.7.x, so `/oem/91_networkmanager.yaml` is already generated
- Make some manual change to that file (or, skip the upgrade and just create a file with that name manually)
- Run `harvester-installer generate-network-yaml`. It should refuse to overwrite `/oem/91_networkmanager.yaml `
- Run `harvester-installer generate-network-yaml --force`. It should overwrite `/oem/91_networkmanager.yaml `

### DIFF
--- a/main.go
+++ b/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"log"
 	"os"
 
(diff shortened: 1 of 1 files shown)


tserong:
Dropped the second commit that was setting dhcp-client-id=duid as it didn't work in my testing. I'll look at that separately.

ibrokethecloud (approved):
lgtm. thanks.

PR #1160, opened 2025-10-07
feat: support two types of multipath configuration

Yu-Jack:
#### Problem:

Reading `[]` will break the code.

```diff
--- /oem/harvester.config.bak	2025-10-01 08:36:28.887734155 +0000
+++ /oem/harvester.config	2025-10-01 08:39:41.915771323 +0000
@@ -24,7 +24,7 @@
     persistentstatepaths: []
     externalstorage:
         enabled: false
-        multipathconfig: []
+        multipathconfig: {}
     additionalkernelarguments: ""
 install:
     automatic: false
```

#### Solution:

Bring https://github.com/harvester/harvester-installer/pull/1117/commits/fb1da59e318dc841514e2f0d7538b32e5cfdb4aa commit back to support two types of multuipath configuration. We could open another new issue to migrate the setting. Otherwise, it's too risky to do in this version.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9240

#### Test plan:

1. Upgrade from v1.6.0 to master version.
2. Ensure unit test cases in this PR run well.

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -867,7 +867,7 @@ func Test_MultipathConfigOption1_DirectFromPureYAML(t *testing.T) {
 
 	yamlContent := `
 os:
-  externalstorage:
+  externalstorageconfig:
     enabled: true
     multipathconfig:
       - vendor: "HP"
@@ -878,8 +878,8 @@ os:
         product: "STORAGE3"
 `
 
-	config := NewHarvesterConfig()
-	err := yaml.Unmarshal([]byte(yamlContent), config)
+	// Be careful the yaml key name, please check this https://github.com/harvester/harvester/issues/9290
+	config, err := LoadHarvesterConfig([]byte(yamlContent))
 	assert.NoError(err, "expected no error while unmarshaling YAML")
 
 	assert.True(config.OS.ExternalStorage.Enabled, "expected external storage to be enabled")
@@ -927,7 +927,7 @@ func Test_MultipathConfigOption2_DirectFromPureYAML(t *testing.T) {
 
 	yamlContent := `
 os:
-  externalstorage:
+  externalstorageconfig:
     enabled: true
     multipathconfig:
       blacklist:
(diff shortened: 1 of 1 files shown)


tserong:
Do we really need to actually _support_ the old format? 

The problem with parsing occurs in getHarvesterConfig(): https://github.com/harvester/harvester-installer/blob/1262a8a0a188ce028ac8aedf852bc18a433ed99f/pkg/console/dashboard_panels.go#L557 which uses `yaml.Unmarshal()`, which fails when the data structure is unexpected.

If we replace that `yaml.Unmarshal()` call with a call to `config.LoadHarvesterConfig()` https://github.com/harvester/harvester-installer/blob/1262a8a0a188ce028ac8aedf852bc18a433ed99f/pkg/config/schemas.go#L26 it won't/shouldn't care if the old format is used, it'll just skip over it and leave us with an empty MultiPathConfig if there's something unexpected there.

I know that means the MultiPathConfig read after installation will be invalid/empty, but it's never actually used after installation, so theoretically this doesn't matter.

OTOH, if there ever were a case where we _did_ need the old config from some previous install, then my suggestion here would break that.

tserong:
My point was that if we replace it with LoadHarvesterConfig, it doesn't matter if the result is unexpected, because we never actually use the multipath config data _after_ the first install. But you're probably right, actually supporting the old format would be safest :-)

Yu-Jack:
Yeah. I know we don't use it after first installation. But, we still need to care of parsing function. And, the reason why It generates two different result in my test cases is probably json/yaml tag issue. I prefer to stay safe and open a new issue to fix this problem. It should be fixed together.

```
KubeovnOperatorChartVersion string            `json:"kubeovnChartVersion,omitempty"`
ExternalStorage           ExternalStorageConfig `json:"externalStorageConfig,omitempty"`
```

tserong (changes requested):
LGTM in general, but I'd suggest that in addition to what you've got here, we also change `console.getHarvesterConfig()` to make it call `config.LoadHarvesterConfig()` instead of `yaml.Unmarshal()`, and also change the tests so that they're implemented to use `console.LoadHarvesterConfig()`. If we do this, it means we use the exact same code path for loading the config during install, and after install.

PR #1213, opened 2026-01-02
feat: add initial iptables rules via NetworkManager dispatcher

mingshuoqiu:
#### Problem:

There should be a way to blacklist external network or whitelist the Harvester node network only to protection from malicious scan and access.

#### Solution:

This commit introduces a NetworkManager dispatcher script to apply initial iptables rules when network interfaces are brought up.

This commit introduces a NetworkManager dispatcher script to apply initial iptables rules when network interfaces are brought up. Creating two custom chains:
  - `HARVESTER_ESSENTIAL`: Contains core rules required for Harvester and RKE2 functionality (SSH, K8s API, VXLAN, etc.).
  - `HARVESTER_DYNAMIC`: A placeholder chain for users or external components

#### Related Issue(s):

https://github.com/harvester/harvester/issues/5681

#### Test plan:

Should include all feature as follows which should be broken into details
- Core Cluster Functionalities
- Storage Network Tests
- VM Operation Tests
- External Integration Tests
  - Rancher management and other Cloud Providers
- Security Tests
  - Use checkmk or nmap scan against nodes

### DIFF
--- a/package/harvester-os/Dockerfile
+++ b/package/harvester-os/Dockerfile
@@ -5,7 +5,9 @@ ARG ARCH=amd64
 RUN curl -sfL https://github.com/rancher/wharfie/releases/download/v0.6.8/wharfie-${ARCH}  -o /usr/bin/wharfie && chmod +x /usr/bin/wharfie
 
 COPY files/ /
-RUN chmod 0600 /system/oem/*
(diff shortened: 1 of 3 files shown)


Vicente-Cheng:
no need this one

PR #1274, opened 2026-05-05
Do not assume harvester repo a git repo

bk201:
#### Problem:

We remove the need to pass the `.git` directory into building container in https://github.com/harvester/harvester/pull/10477

#### Solution:

Don't assume we can restore changes by checking out modified files. Copy the unmodified charts to a backup location and restore once we are done.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/10478


wheatdog (approved):

Vicente-Cheng (approved):
lgtm, thanks!

PR #1286, opened 2026-05-28
fix: trivy vex issue

pohanhuang:
#### Problem:

- current trivy scan action will break, due to the download vex file broken

#### Solution:

- old endpoint will not provide file any more, use lfs instead


wheatdog (dismissed):

pohanhuang:
@mergify backport v1.8 v1.7

Vicente-Cheng (approved):

PR #1298, opened 2026-06-15
fix: allow ipv6 on kernel/host level

martindekov:
Allowing ipv6 on kernel level persisted over reboots so on subsequent PR we can create NetworkManager
configuration which can configure dual-stack which allows assignment of ipv6 addresses.

The current change layers dynamic config on top of the baseline ipv6.conf in the repo along with sysctl config allowing ipv6.

Change is safe to merge with no impact because Network Manager still disabled ipv6 on network interface level.

Added unit tests with check as part of the table for cos file. Check is part of the table as the assertion steps were different based on directory/file/content.

#### Problem:
Currently we disable ipv6 on a kernel level which also prevents networkmanager from configuring network interface which supports dual stack.

#### Solution:
Allow ipv6 on kernel level, while still disabling it on network interface level. That way we can then add change which configures the network interface with dual stack along with installer UX.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/10795
https://github.com/harvester/harvester/issues/10755

#### Test plan:
Added unit tests.

End to end testing:
<details>
<summary>After first boot IPv6 is allowed on kernel level</summary>

```
mdekov@localhost:~> ssh rancher@192.168.122.66
The authenticity of host '192.168.122.66 (192.168.122.66)' can't be established.
ED25519 key fingerprint is: SHA256:23J9vMktalJp39LH1KuNflQ8sLGudOLkxkxzTPENvp8
This key is not known by any other names.
Are you sure you want to continue connecting (yes/no/[fingerprint])? yes
Warning: Permanently added '192.168.122.66' (ED25519) to the list of known hosts.
Have a lot of fun...
rancher@node-1:~> sudo su
node-1:/home/rancher # sysctl net.ipv6.conf.all.disable_ipv6
net.ipv6.conf.all.disable_ipv6 = 0
node-1:/home/rancher # cat /etc/sysctl.d/zz-harvester-enable-ipv6.conf
# Written by harvester-installer: overrides /etc/sysctl.d/ipv6.conf
net.ipv6.conf.all.disable_ipv6 = 0
net.ipv6.conf.default.disable_ipv6 = 0
net.ipv6.conf.lo.disable_ipv6 = 0
node-1:/home/rancher # sysctl net.ipv6.conf.all.disable_ipv6
net.ipv6.conf.all.disable_ipv6 = 0
node-1:/home/rancher # sysctl net.ipv6.conf.all.disable_ipv6
net.ipv6.conf.all.disable_ipv6 = 0
node-1:/home/rancher # sysctl net.ipv6.conf.default.disable_ipv6
net.ipv6.conf.default.disable_ipv6 = 0
node-1:/home/rancher # sysctl net.ipv6.conf.lo.disable_ipv6
net.ipv6.conf.lo.disable_ipv6 = 0
node-1:/home/rancher # reboot
```
</details>

<details>
<summary>NetworkManager has IPv6 disabled</summary>

```
node-1:/home/rancher # grep -r "method" /etc/NetworkManager/system-connections/
/etc/NetworkManager/system-connections/bridge-mgmt.nmconnection:method=auto
/etc/NetworkManager/system-connections/bridge-mgmt.nmconnection:method=disabled
node-1:/home/rancher # nmcli connection show
NAME               UUID                                  TYPE      DEVICE       
bridge-mgmt        713373f6-bf9d-3ee3-abb2-5f295bd18145  bridge    mgmt-br      
bond-mgmt          48589773-cb7d-38c5-b579-7dd119af73b6  bond      mgmt-bo      
bond-slave-enp1s0  97ed9ced-a5cc-306a-b37c-8a6a1e6f0ad8  ethernet  enp1s0       
lo                 1fef8cb1-e3d9-48d5-820f-0e1e01b1b437  loopback  lo           
vip-f4baf473       6ec7587c-ba1e-483b-9ecb-6631bb8b334b  macvlan   vip-f4baf473 
node-1:/home/rancher # nmcli connection show bridge-mgmt | grep IP6
IP6.GATEWAY:                            --
node-1:/home/rancher # sysctl net.ipv6.conf.mgmt-br.disable_ipv6
net.ipv6.conf.mgmt-br.disable_ipv6 = 1
node-1:/home/rancher # 
```
</details>

<details>
<summary>After reboot kernel settings persisted and NetworkManager still has IPv6 disabled</summary>

```
node-1:/home/rancher # reboot
node-1:/home/rancher # Read from remote host 192.168.122.66: Connection reset by peer
Connection to 192.168.122.66 closed.
client_loop: send disconnect: Broken pipe
mdekov@localhost:~> ssh rancher@192.168.122.66
Have a lot of fun...
rancher@node-1:~> sudo su
node-1:/home/rancher # sysctl net.ipv6.conf.all.disable_ipv6
net.ipv6.conf.all.disable_ipv6 = 0
node-1:/home/rancher # cat /etc/sysctl.d/zz-harvester-enable-ipv6.conf
# Written by harvester-installer: overrides /etc/sysctl.d/ipv6.conf
net.ipv6.conf.all.disable_ipv6 = 0
net.ipv6.conf.default.disable_ipv6 = 0
net.ipv6.conf.lo.disable_ipv6 = 0
node-1:/home/rancher # sysctl net.ipv6.conf.all.disable_ipv6
net.ipv6.conf.all.disable_ipv6 = 0
node-1:/home/rancher # 
node-1:/home/rancher # sysctl net.ipv6.conf.all.disable_ipv6
net.ipv6.conf.all.disable_ipv6 = 0
node-1:/home/rancher # sysctl net.ipv6.conf.default.disable_ipv6
net.ipv6.conf.default.disable_ipv6 = 0
node-1:/home/rancher # sysctl net.ipv6.conf.lo.disable_ipv6
net.ipv6.conf.lo.disable_ipv6 = 0
node-1:/home/rancher # grep -r "method" /etc/NetworkManager/system-connections/
/etc/NetworkManager/system-connections/bridge-mgmt.nmconnection:method=auto
/etc/NetworkManager/system-connections/bridge-mgmt.nmconnection:method=disabled
node-1:/home/rancher # nmcli connection show bridge-mgmt | grep IP6
IP6.GATEWAY:                            --
node-1:/home/rancher # sysctl net.ipv6.conf.mgmt-br.disable_ipv6
net.ipv6.conf.mgmt-br.disable_ipv6 = 1
node-1:/home/rancher # 
```
</details>

[HEP](https://github.com/martindekov/harvester/blob/c7eab08d811393988c07d2ec632f1f3079c565d6/enhancements/20260528-dual-stack-ipv4-first.md)

The change is added so it won't break the existing functionality or features e.g. - log warning in case we can't enable the system ipv6 settings as of today when we only support ipv4.

### DIFF
--- a/pkg/config/cos.go
+++ b/pkg/config/cos.go
@@ -204,10 +204,8 @@ func ConvertToCOS(config *HarvesterConfig) (*yipSchema.YipConfig, error) {
 	})
 	initramfs.Files = append(initramfs.Files, yipSchema.File{
 		Path: "/etc/sysctl.d/zz-harvester-enable-ipv6.conf",
-		Content: "# Written by harvester-installer: overrides /etc/sysctl.d/ipv6.conf\n" +
-			"net.ipv6.conf.all.disable_ipv6 = 0\n" +
-			"net.ipv6.conf.default.disable_ipv6 = 0\n" +
-			"net.ipv6.conf.lo.disable_ipv6 = 0\n",
+		Content: fmt.Sprintf("# Written by harvester-installer: overrides /etc/sysctl.d/ipv6.conf\n%s = 0\n%s = 0\n%s = 0\n",
+			SysctlDisableIPv6All, SysctlDisableIPv6Default, SysctlDisableIPv6Lo),
 		Permissions: 0644,
 		Owner:       0,
 		Group:       0,
@@ -215,9 +213,9 @@ func ConvertToCOS(config *HarvesterConfig) (*yipSchema.YipConfig, error) {
 	if initramfs.Sysctl == nil {
 		initramfs.Sysctl = make(map[string]string)
 	}
-	initramfs.Sysctl["net.ipv6.conf.all.disable_ipv6"] = "0"
-	initramfs.Sysctl["net.ipv6.conf.default.disable_ipv6"] = "0"
-	initramfs.Sysctl["net.ipv6.conf.lo.disable_ipv6"] = "0"
+	initramfs.Sysctl[SysctlDisableIPv6All] = "0"
+	initramfs.Sysctl[SysctlDisableIPv6Default] = "0"
+	initramfs.Sysctl[SysctlDisableIPv6Lo] = "0"
 
 	// TOP
 	if cfg.Install.Mode != ModeInstall {
--- a/pkg/config/cos.go
+++ b/pkg/config/cos.go
@@ -195,6 +195,30 @@ func ConvertToCOS(config *HarvesterConfig) (*yipSchema.YipConfig, error) {
 	// disable multipath for longhorn
 	disableLonghornMultipathing(&initramfs)
 
+	// write a persistent sysctl drop-in and apply at runtime; persists after reboot
+	initramfs.Directories = append(initramfs.Directories, yipSchema.Directory{
+		Path:        "/etc/sysctl.d",
+		Permissions: 0755,
+		Owner:       0,
+		Group:       0,
+	})
+	initramfs.Files = append(initramfs.Files, yipSchema.File{
+		Path: "/etc/sysctl.d/zz-harvester-enable-ipv6.conf",
+		Content: "# Written by harvester-installer: overrides /etc/sysctl.d/ipv6.conf\n" +
+			"net.ipv6.conf.all.disable_ipv6 = 0\n" +
+			"net.ipv6.conf.default.disable_ipv6 = 0\n" +
+			"net.ipv6.conf.lo.disable_ipv6 = 0\n",
+		Permissions: 0644,
+		Owner:       0,
+		Group:       0,
+	})
+	if initramfs.Sysctl == nil {
+		initramfs.Sysctl = make(map[string]string)
+	}
+	initramfs.Sysctl["net.ipv6.conf.all.disable_ipv6"] = "0"
+	initramfs.Sysctl["net.ipv6.conf.default.disable_ipv6"] = "0"
+	initramfs.Sysctl["net.ipv6.conf.lo.disable_ipv6"] = "0"
+
 	// TOP
 	if cfg.Install.Mode != ModeInstall {
 		if err := initRancherdStage(config, &initramfs); err != nil {
--- a/pkg/console/network.go
+++ b/pkg/console/network.go
@@ -89,6 +89,17 @@ func applyNetworks(network config.Network, hostname string) error {
 		}
 	}
 
+	// enable IPv6 before applying NetworkManager profiles
+	for _, param := range []string{
+		"net.ipv6.conf.all.disable_ipv6=0",
+		"net.ipv6.conf.default.disable_ipv6=0",
+		"net.ipv6.conf.lo.disable_ipv6=0",
+	} {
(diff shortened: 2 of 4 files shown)


pkg/config/cos.go
```diff
@@ -195,6 +195,30 @@ func ConvertToCOS(config *HarvesterConfig) (*yipSchema.YipConfig, error) {
 	// disable multipath for longhorn
 	disableLonghornMultipathing(&initramfs)
 
+	// write a persistent sysctl drop-in and apply at runtime; persists after reboot
```
w13915984028, line 198: Is there an installer option to indicate if ipv6 is required (dual mode or ipv6 only mode) ? thanks.

martindekov: This is the groundwork PR, on subsequent PR (the sub task 1.2 of the installer) we can add that feature along with conditional dual stack cidr and conditional network interface configuration based on installer option / panel

w13915984028 (approved):
LGTM, thanks.

pkg/console/network.go
```diff
@@ -89,6 +89,17 @@ func applyNetworks(network config.Network, hostname string) error {
...
 	}
 
+	// enable IPv6 before applying NetworkManager profiles
+	for _, param := range []string{
+		config.SysctlDisableIPv6All,
+		config.SysctlDisableIPv6Default,
+		config.SysctlDisableIPv6Lo,
+	} {
+		if out, execErr := exec.Command("sysctl", "-w", fmt.Sprintf("%s=0", param)).CombinedOutput(); execErr != nil {
+			logrus.Warnf("Failed to enable IPv6 sysctl %s: %v (%s)", param, execErr, string(out))
```
votdev, line 99: Why is this only a warning and why is an error silently ignored?

martindekov: Thanks for the review Volker, good question!

As of today we can't say whether we are in dual stack or ipv4 only mode. This is warning as in case it fails it's not critical for harvester - we don't expect ipv6 to work in  IPv4 scenario and we don't expect our installations to fail with ipv6 related errors as of today.

Once we have way to say - ok we are in dual stack mode, then the error here becomes critical for the work of the current configuration and it becomes distinguishable. We can either assign err = execErr or skip evaluation altogether. This is close to what Jian said above.

But merging the change as is today won't break anything in case we decide to release tomorrow with this change in.

This is good catch also it is not clear that's why I added as first point in the feature itself to make this a toggle:

https://github.com/harvester/harvester/issues/10796

>Make sure IPv6 is enabled at Kernel level: When we are in dual stack scenario make sure we error out in case we are in dual stack scenario. Currently in [applyNetworks](https://github.com/harvester/harvester-installer/blob/0d8dd3b9a8ddc8629fbf55e0a05c770ea3e4bcb7/pkg/console/network.go#L37) we only log warning as we support IPv4 before this feature. Concerns were raised https://github.com/harvester/harvester-installer/pull/1298#discussion_r3419446637 and https://github.com/harvester/harvester-installer/pull/1298#discussion_r3412893947. So stating it clearly as first point.

martindekov: Again thanks for the question, will leave the discussion open for others to take a look in case they wonder as well

tserong (approved):

PR #1313, opened 2026-07-10
fix: do not fatally abort install on nm-online connectivity check

rustydb:
#### Problem:

The management-network readiness check ran `nm-online -x`, which exits immediately instead of waiting, and treated any non-zero result as fatal. On the already-installed first-boot path this check runs before the RancherD config is written, so a management link that is briefly offline while the bond re-converges aborts the whole install and leaves the node with no role until manual repair.

#### Solution:

Two changes:

- Wait for the network instead of sampling once: use `nm-online -s -t 45` instead of `-x`. `-s` waits for NetworkManager startup to complete (interfaces finished activating) rather than for global connectivity, which an isolated management network may never reach. The 100s timeout gives a slow-rate LACP bond headroom: `lacp_rate=slow` emits an LACPDU only every 30s, so a single missed interval can cost ~30s before the bond syncs. This gives slow-LACP three attempts with grace. Because `-s` returns as soon as startup completes, this budget only costs time on a genuinely slow or degraded link.
- Make the check non-fatal: log a warning and continue instead of returning an error, so a slow or briefly-degraded management link no longer aborts the install (and, on the installed-node path, no longer strands the node without a RancherD config).

This is observed on both LACP/802.3ad and single-NIC active-backup bonds, and across install methods, so it is not specific to any one bond mode.

#### Related Issue(s):

Refs: harvester/harvester#10885

### DIFF
--- a/pkg/config/cos.go
+++ b/pkg/config/cos.go
@@ -588,14 +588,28 @@ func UpdateManagementInterfaceConfig(mgmtInterface Network, dnsNameServers []str
 			logrus.Error(err, string(output))
 			return err
 		}
-		// This next command waits up to 30 seconds to ensure there's
-		// a connection.  Without this, it's possible that a slow DHCP
-		// server won't return in time, and the installer will subsequently
-		// fail the check for a default route.
-		output, err = exec.Command("nm-online", "-x").CombinedOutput()
+		// Wait (up to the timeout) for NetworkManager to finish (re)starting the
+		// management network before continuing.
+		//
+		// Use "-s" (wait for startup to complete) instead of "-x": "-x" exits
+		// immediately and defeats the intended wait, and startup (interfaces
+		// finished activating) is the right signal rather than *global*
+		// connectivity, which an isolated management network may never reach.
+		//
+		// The 45s timeout gives a slow-rate LACP bond headroom to converge:
+		// lacp_rate=slow emits an LACPDU only every 30s, so a single missed
+		// interval can cost ~30s before the bond syncs.  Because "-s" returns as
+		// soon as startup completes, this budget only costs time on a genuinely
+		// slow or degraded link.
+		//
+		// The check is best-effort and must NOT be fatal.  On the
+		// already-installed first-boot path it runs before the RancherD config is
+		// written, so aborting the install here on a briefly-degraded link (e.g. a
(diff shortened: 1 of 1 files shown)


pkg/config/cos.go
```diff
@@ -588,14 +588,28 @@ func UpdateManagementInterfaceConfig(mgmtInterface Network, dnsNameServers []str
...
+		// interval can cost ~30s before the bond syncs.  Because "-s" returns as
+		// soon as startup completes, this budget only costs time on a genuinely
+		// slow or degraded link.
+		//
+		// The check is best-effort and must NOT be fatal.  On the
+		// already-installed first-boot path it runs before the RancherD config is
+		// written, so aborting the install here on a briefly-degraded link (e.g. a
+		// bond still re-converging) leaves the node with no role until manual
+		// repair.  See harvester/harvester#10885.
+		output, err = exec.Command("nm-online", "-s", "-t", "45").CombinedOutput()
```
w13915984028: It looks we could wait `100s`, it includes three times `30s` retry room for slow lacp, and use `-s` could be a bit robust than `-x`

as on next stage, there are DNS/NTP/remote SSH ID fetch proces, if the link is still not working, it will fail there, and the remaining link gives more chance to run command like `ip link` for debugging.

w13915984028 (commented):
@rustydb  Thanks for your PR.

Could you rebase the code to run PR upon repo `harvester/harvester` and `/pkg/installer`  (the `harvester/harvester-installer` repo is merged to `harvester/harvester`? Then we could backport the PR to v1.9 branch.

pkg/config/cos.go
rustydb: > and use `-s` could be a bit robust than `-x`

It is using `-s`, just pointing that out incase you meant something else.

I'll extend it to 100s, I take no issue with that.

rustydb:
> @rustydb Thanks for your PR.
> 
> Could you rebase the code to run PR upon repo `harvester/harvester` and `/pkg/installer` (the `harvester/harvester-installer` repo is merged to `harvester/harvester`? Then we could backport the PR to v1.9 branch.

ok, I will make a PR to harvester/harvester:master.

rustydb:
Closing. 
Re-raised in harvester/harvester
https://github.com/harvester/harvester/pull/11690

